PR #3, opened 2019-08-18
[Jeremy Tan] Duke Increments

Parcly-Taxel:
My set of increments for the individual project up to week 3.

Pull requests are called merge requests over on GitLab, where [my four personal repositories](https://gitlab.com/parclytaxel) lie.


dingyuchen:
great job!

shuiyao-sg (commented):
Your Command class has a static method parse(String cmd). Would it be better if you introduce a Parser class to pass the user commands so as to achieve higher level of abstraction and make it more similar to real-life situation?

You used lower-case letters for naming the final variables. The [CS2103T Coding Standard](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#naming) specifies that all constants should be in uppercase. However, I also checked against the Oracle Official Website on [Java naming conventions](https://www.oracle.com/technetwork/java/codeconventions-135099.html) and found that only class constants and of ANSI constants need to be all uppercase. So I'm a bit unsure of the naming for non-static final variables. Would you mind helping me clarify the doubts?

Everything else looks great!

Parcly-Taxel:
My code used to have a `Parser` class, following the specifications. However, I decided that since only one object of that class is used in the entire code, and since the parsed format is intricately tied to the application-specific `Command`s, it would be better to make `parse()` a factory method of `Command`. For the same reason I also removed the `Ui` class.

As for the final variables, I added them on IntelliJ's advice – initialised but never modified afterwards. I recognise that some of them would be better off non-final for simplicity's sake, and have changed them accordingly.

eejian97 (commented):
Hi there, Jeremy . Your code is very well structured and clean . I have learnt alot from reviewing the code . Can i ask why you decided against implementing a Ui class so that the logic and the Ui is better separated? Thanks!

Parcly-Taxel:
Simple. When I implemented the GUI, the text `Ui` was made redundant.

PR #10, opened 2019-08-23
[Kheng Yau Dong] Duke Increments

ChrisKheng:


src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
+import duke.command.Command;
```
TSAI-HSIAO-HAN: Can place a line between different package import.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,146 @@
+package duke.command;
+
+import duke.component.DukeDatabase;
```
TSAI-HSIAO-HAN, line 3: Can place a line between different package import.

TSAI-HSIAO-HAN (commented):
The overall Java style seems correct.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            throw new DukeException("Internal logic bug occurred!");
+        }
+    }
+
+    /**
+     * Adds to do entry to the taskList.
+     */
+    private void addToDo() throws DukeException {
+        String topic = input.substring(4).trim();
```
muhammadaljunied: An alternative to String manipulation would be to use Regular Expressions. You can refer to my fork in Duke.java to see how it could be used. As the complexity of the commands increase, writing equivalent String Manipulation code would become increasingly tedious.

In our use case, the commands are trivial, and would likely be faster than using Regex, however the tradeoff would be a little more refactoring in the future when commands become more complex.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,146 @@
...
+
+        if (details.length == 1 || "".equals(details[0].trim()) || "".equals(details[1].trim())) {
+            throw new DukeException("The description and date of an event cannot be empty.");
+        }
+
+        try {
+            String topic = details[0].stripTrailing();
+            String dateUserInput = details[1].stripLeading();
+            String date = formatDateAndTime(dateUserInput);
+
```
muhammadaljunied: An alternative to String manipulation would be to use Regular Expressions. You can refer to my fork in Duke.java to see how it could be used. As the complexity of the commands increase, writing equivalent String Manipulation code would become increasingly tedious.

In our use case, the commands are trivial, and would likely be faster than using Regex, however the tradeoff would be a little more refactoring in the future when commands become more complex.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Adds event entry to the taskList.
+     */
+    private void addEvent() throws DukeException {
+        String[] details = input.substring(5).trim().split("/at");
+
+        if (details.length == 1 || "".equals(details[0].trim()) || "".equals(details[1].trim())) {
+            throw new DukeException("The description and date of an event cannot be empty.");
+        }
```
muhammadaljunied: I think I understand where you are coming from from an abstraction point of view for your commands. However, in terms of separation of concerns, I feel that the Parser object is more suited to handle the parsing of the command arguments rather than for the command itself to handle it.

I too debated with this, and settled for the Parser to handle it. Let me know what you think!

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,25 @@
...
+public class DukeException extends Exception {
+    /**
+     * The prefix of the error message of all DukeException objects.
+     */
+    public static final String PREFIX =  "\u2639  OOPS!!!"; // The unicode is a sad face symbol.
+
+    public DukeException() {
+        super(String.format("%s DukeException occurs!", PREFIX));
+    }
+
```
muhammadaljunied, line 21: I would add the following Constructor here:

```
public DukeException(String message, Throwable cause) {
    super(message, cause);
}
```

What this does is it allows you to catch the underlying exception in the `catch` block, then re-throw a DukeException together with the real cause. This is reflected in the stack trace which provides the developer with a better inkling as to what exactly caused the error in the `try` block.

Again, our app is trivial and may not require that much finesse but hey, it'll help when we do AddressBook!

src/main/java/duke/component/DukeDatabase.java
```diff
@@ -0,0 +1,144 @@
...
+        return task;
+    }
+
+    /**
+     * Updates the database file (used at the end of the program).
+     *
+     * @param tasks a list of tasks as updated at the point when the program ends.
+     */
+    public void update(TaskList tasks) {
+        // Extract all the data summaries of the tasks and append them into a single string.
```
muhammadaljunied, line 142: A few things I noticed here:

1. If program terminates (assume an unhandled exception) without calling `exit()` in ExitCommand.java, my changes would not be saved. An easy fix would be to add the following to your Duke.java:run() try-catch block:

```
...
finally {
    database.update(taskList);
}
```

2. Should my list size grow to the thousands (though unlikely), the list would take longer to write to file.

For 2. you could write an `append(int index, String data)` method that appends the new task into the file. Since each line denotes one task, and they are in running order, it would be easy for us to "insert" a new deadline. This reduces our write operation to a single line rather than n lines.

Then calling the new `append(int index, String data)` after each task is added to the list, would then eliminate the issues in 1 and 2! There is a tradeoff however, there will be more IO operations.

src/main/java/duke/component/DukeDatabase.java
```diff
@@ -0,0 +1,144 @@
...
+        StringBuilder bldr = new StringBuilder(150);
+        tasks.forEach(t -> {
+            bldr.append(String.format("%s\n", t.getSummaryForDatabase()));
+        });
+
+        // Write all the data summaries into the database file.
+        try {
+            FileWriter fw = new FileWriter(databaseDirectory);
+            fw.write(bldr.toString());
+            fw.close();
```
muhammadaljunied, line 152: Here it would be advisable to move `fw.close()` into a `finally` block. 

```
try {
    FileWriter fw = new FileWriter(databaseDirectory);
    fw.write(bldr.toString());
} catch (IOException e) {
    System.out.println("Database error occurs when updating data!");
} finally {
    fw.close();
}
```

The reason for this is, suppose the `try-catch` catches an exception after `fw` has gained file access. When the exception is handled, `fw` still has access to the file in `databaseDirectory`. The next time you try to access this file, you will encounter an IOException again.

The `finally` block always executes regardless of Exceptions caught in the `try` block.

src/main/java/duke/component/DukeDatabase.java
```diff
@@ -0,0 +1,144 @@
...
+    private DukeDatabase() {
+    }
+
+    /**
+     * Returns a DukeDatabase object.
+     *
+     * @return a DukeDatabase object.
+     */
+    public static DukeDatabase getDukeDatabaseInstance() {
+        return new DukeDatabase().initialise();
```
muhammadaljunied, line 37: Nice Singleton usage (Y)

src/main/java/duke/component/DukeDatabase.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @return the database object itself.
+     */
+    private DukeDatabase initialise() {
+        // Generates the directory for the database
+        String projectRoot = new File(System.getProperty("user.dir"))
+                .getParentFile()
+                .getPath();
+        databaseDirectory = String.format("%s/data/duke.txt", projectRoot);
+
```
muhammadaljunied, line 51: I would add the following here:

```
if (tasksFile.exists()) {
   return;
}
```

It appears that if the file exists, IOException will always occur on line 51. After checking Java Documentation, it actually does not throw an exception. The exist check happens atomically with all other filesystem operations. From a readability standpoint, it would be clearer to read with the above snippet.

src/main/java/duke/command/AddCommand.java
ChrisKheng: Personally, I think it would be good to separate the abstraction instead of putting all the parsing jobs in the Parser class cuz this may make Parser class become too big, but maybe you're right too!

src/main/java/duke/component/DukeDatabase.java
ChrisKheng: Thanks for the suggestion! Regarding point 2, I understand where you're coming from, however, it can become an issue when we want to update and delete a certain task in the list. If we want to only delete or update a task in the list and then we will need to go through the entire data file in order to do so as we could not just operate on a certain line in the data file without going through the rest of the lines in it, as pointed in some StackOverflow posts. Thus, I choose to only write the file at the end of the program to reduce the overall number of write operations needed.

Anyway, thanks for the suggestion on point 1 and I'll look into it if I have the time!

src/main/java/duke/exception/DukeException.java
ChrisKheng: Alright, thanks a lot for the suggestion!

PR #11, opened 2019-08-23
[Ahmed Bahajjaj] Duke Increments

madanalogy:
I did the thing


eugenood:
Good job on your contribution. I think the abstraction you introduced is good, and the quality of your code is consistent. I only have two nitpicks:

- I think your [`TaskList`](https://github.com/madanalogy/duke/blob/master/src/main/java/duke/task/TaskList.java) could just inherits `ArrayList`.
- Be careful when IntelliJ automatically combine import statements in your [`Parser`](https://github.com/madanalogy/duke/blob/7d1e3ecbf4dcdf7596fda715568b504424190961/src/main/java/duke/io/Parser.java#L3) class. See [this](https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports) SO post for fix.

danielwys:
Good job! I see you've implemented Duke exactly as suggested by the specifications, and even accomplished all the optional stretch goals. Your formatting and code style are also consistent. Keep up the good work :)

madanalogy:
> Good job on your contribution. I think the abstraction you introduced is good, and the quality of your code is consistent. I only have two nitpicks:
> 
> * I think your [`TaskList`](https://github.com/madanalogy/duke/blob/master/src/main/java/duke/task/TaskList.java) could just inherits `ArrayList`.
> * Be careful when IntelliJ automatically combine import statements in your [`Parser`](https://github.com/madanalogy/duke/blob/7d1e3ecbf4dcdf7596fda715568b504424190961/src/main/java/duke/io/Parser.java#L3) class. See [this](https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports) SO post for fix.

I didn't know I could just inherit ArrayList, thanks for the suggestion! And thanks for the fix also haha, appreciate the help :)

> Good job! I see you've implemented Duke exactly as suggested by the specifications, and even accomplished all the optional stretch goals. Your formatting and code style are also consistent. Keep up the good work :)

Thank you! :D

PR #13, opened 2019-08-23
[Ramachandran Sandhya] Duke Increments

sandydays:
Add Week 2 iPs


src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+                        printAddedTask(event, tasks.size());
+                        break;
+                    default:
+                        hasChanged = false;
+                        throw new DukeException("\u2639 OOPS!!! I'm sorry, but I don't know what that means :-(");
+                }
+            } catch (DukeException ex) {
+                System.out.println(ex.getMessage());
+            } finally {
+                if (hasChanged) {
```
shiyao821: The current implementation of saving your tasks after every command is fine for now. However, this process may take a long time when the to-do list gets long. Perhaps save only when the 'bye' command is entered?

src/main/java/DateAndTime.java
```diff
@@ -0,0 +1,34 @@
...
+    protected int date;
+    protected int month;
+    protected int year;
+    protected int timeNumber;
+    protected String timeDay;
+
+    protected String[] monthMappings = new String[]{"", "January", "February", "March", "April", "May", "June",
+                                                    "July", "August", "September", "October", "November", "December"};
+
+    public DateAndTime(String dateAndTime) {
```
shiyao821: Documentation on parse-able dateAndTime strings will help improve code readability

src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+            }
+        }
+
+        boolean hasChanged = true;
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?");
+        String userInput = sc.nextLine();
+
+        while (!userInput.equals("bye")) {
+            try {
+                String[] command = userInput.split(" ", 2);
```
shiyao821: This is smart.

src/main/java/DateAndTime.java
```diff
@@ -0,0 +1,34 @@
+public class DateAndTime {
```
shiyao821: Very intuitive way of processing date and time. I like it! Moving on, checks should be made to prevent invalid date/time inputs from being processed eg. 13/16/19 2460

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
+public class Event extends Task {
+    protected DateAndTime at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = new DateAndTime(at);
```
shiyao821: What if the input given by the user is not an intended date and time? eg. (at: Block 903/904)?

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    @Override
+    public String toSave() {
+        return "E | " + (isDone ? 1 : 0) + " | " + description + " | " + at;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + at + ")";
```
shiyao821: Calling the super class toString() here is pretty smart. I've never thought of it.

shiyao821 (commented):
Very intuitive and generally easy to understand code. 

To be done: 
- [ ] Documentation
- [ ] Tests
- [ ] Adding Ui, Parser, Storage, TaskList classes (as per Wk 3 requirements)
- [ ] Level-9 find command

src/main/java/DateAndTime.java
larrylawl: I like the effort that you have put in! 
Just sharing another POV - you can use the built-in type `Date` (find out more [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Date.html)), instead of creating your own class.

larrylawl: I just realised the parsing of the date is also handled here. You can use the built in class `SimpleDateFormat` to parse input strings, and also output it in the format you want. 

Eg on parsing
```
String format = "dd/MM/yyyy HHmm";
SimpleDateFormat readFormat = new SimpleDateFormat(format);
readFormat.parse(//input string)
```

src/main/java/Duke.java
larrylawl: Hmm I think Sandy is right though. Specification from the [website](The specification on the [website](https://nus-cs2103-ay1920s1.github.io/website/schedule/week3/project.html) says ) - "Save the tasks in the hard disk automatically whenever the task list changes."

src/main/java/DukeException.java
```diff
@@ -0,0 +1,14 @@
+public class DukeException extends Exception {
+
```
larrylawl: Good effort on creating your own Duke Exception!

src/main/java/Event.java
larrylawl: Agreed :)

src/main/java/Event.java
sandydays: I have tried to, in my more recent commits, ensure the date and time is of the correct format. I throw an exception if it is not!

src/main/java/Duke.java
sandydays: Thanks :)

src/main/java/DateAndTime.java
sandydays: I now have included Javadoc comments in later commits!

src/main/java/DateAndTime.java
sandydays: Thanks for the inputs!

src/main/java/Event.java
sandydays: Thanks :)

src/main/java/DukeException.java
sandydays: Thanks :)

src/main/java/Duke.java
sandydays: I think @shiyao821, your suggestion is definitely the more efficient one in terms of the number of times you write to your file in memory, but I guess as @larrylawl pointed out, the website seems to indicate otherwise. Right now, with all the GUI also set up, user can just hit the close button on the top right of the interface to exit from the application, and this can be problematic if we only save all changes to the task list when user enters command `bye`. So, I am going with writing the tasks in the file again each time the user changes the task list in any way.

PR #15, opened 2019-08-23
[Ong Yan Chun] Duke Increments

yaaanch:
Added Levels 1 - 6.


yaaanch:
Added Levels 7-9.

yyuanxin:
Hi Yan Chun! I like how your code is very well organised and easy to read.

What I like about your code:
- Well structured with the use of packages 
- Good Javadoc comments that adheres to the coding standards (which is also something I should learn and improve on)
- Good OOP design - I like how you have different classes to handle more specific exceptions

It’s great that you have handled your test cases (for junit testing) in different classes as well! - definitely something that I should consider

Minor suggestions that you can work on:
- Include more Javadoc comments for other public methods and at class level

Overall, I think you have done a great job!

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Date;
+
+public class Parser {
+    public static Command parse(String input) throws DukeException {
+        String[] currentInputArray = input.split("\\s");
+        switch (currentInputArray[0]) {
+        case "list":
+            return new ListCommand();
+        case "done":
+            if (currentInputArray.length < 2) {
```
yyuanxin: You may want to consider shifting the if cases to a separate method that handles checking of valid commands - reduce repeated codes in the switch block for eg. line 35

yyuanxin:
You may want to consider shifting the if cases to a separate method that handles checking of valid commands - reduce repeated codes in the switch block for eg in line 35

https://github.com/nus-cs2103-AY1920S1/duke/blob/b9d96c72905db542c5742d170fd3170deddc0180/src/main/java/duke/main/Parser.java#L26-L28

PR #16, opened 2019-08-23
[Atharv Joshi] Duke increments

atharvjoshi:
Duke at iP Week 2 deliverable stage


dragontho (commented):
Nothing to say, if anything I need to ask you for help to make my code look nicer HAHA. All the principles of software engineering is there (at least those i am aware of).

src/main/java/commands/DoneCommand.java
```diff
@@ -0,0 +1,102 @@
...
+
+                    // update storage
+                    storage.update(tasks);
+
+                    // return message
+                    return ui.showTaskDoneMessage(taskToMarkDone);
+                } else {
+                    return ui.showTaskAlreadyDoneMessage();
+                }
+            } catch (IndexOutOfBoundsException exceptionOne) {
```
shawnkoh, line 72: avoid catching runtime exceptions - they should be handled using if/else instead. a good example of this would be user input errors (e.g. delete a instead of delete 2)
exceptions should be for truly exceptional cases that are out of your control; such as when you try to save the data to the file and for some reason it does not work i.e. IOException. For exceptions that you can control like FileNotFoundException, they should be handled.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ *  @author atharvjoshi
+ *  @version CS2103 AY19/20 Sem 1 iP Week 4
+ */
+public class Parser {
+    /**
+     * Enumeration of all command types supported by Duke.
+     */
+    private enum Imperative {
+        BYE, DEADLINE, DELETE, DONE, EVENT, FIND, LIST, TODO, UNDO
```
shawnkoh, line 35: consider declaring a value for your enums i.e. BYE ("bye"), DEADLINE ("deadline")
This prevents the scenario where a user can provide 0, 1, 2 - currently your enums are defaulted to 0, 1 and 2.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     *  Instantiates the user interface handler.
+     */
+    Ui() {
+        this.logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        this.welcomeMessage = "Hello! I'm Duke!!!\n";
```
shawnkoh, line 35: consider declaring welcomeMessage as a constant i.e. private static final

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,225 @@
...
+        return exitMessage;
+    }
+
+    /**
+     * Returns startup error message.
+     *
+     * @return startup error message.
+     */
+    String showStartupError() {
+        return "Unable to load tasks from hard disk\n";
```
shawnkoh, line 63: inconsistent return - one is a variable and the other a string.
Consider using System.lineSeperator() instead to support all platforms instead of \n

PR #17, opened 2019-08-23
[Liu Zechu] Duke Increments

LiuZechu:
Add the 6 levels i.e. Levels 1-6


SQwQ:
- Good use of enums for different tasks

- Good use of shellscript for testing

- Very well-formatted and readable code

- Good job!

teika97:
Hi Zechu,

It was very easy to read your code because of the formatting and the descriptive names for the variables and methods used. You've also put a lot of thought in packaging the different classes into larger groupings like Command, etc.

A possible improvement I would suggest is to add several classes like Todo, Deadline and Event class that inherit from the Task class. I feel that it could help make your code even more concise. For instance, commands that involve printing information about tasks could be simplified further by adding a toString method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+    }
+
+    /**
+     * Kick-starts the Duke application.
+     */
+    private void run() {
+        ui.greet();
+
+        boolean isExit = false;
+        while (!isExit) {
```
wilfredbtan, line 58: Could it be possible to use an exitHandler in MainWindow instead of using a loop?

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+
+        exitDuke();
+    }
+
+    /**
+     * Saves tasks into hard disk and exits the Duke programme.
+     */
+    public static void exitDuke() {
+        String toSave = "";
+        if (tasks != null) {
```
wilfredbtan, line 77: Can try looking up Object.requireNonNull()

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+    String getResponse(String input) {
+        try {
+            Command c = parser.parse(input);
+            c.execute(tasks, ui, storage);
+        } catch (Exception e) {
+            ui.println(e.getMessage());
+        }
+
+        String toReturn = Main.getBaos().toString();
+        Main.getBaos().reset();
```
wilfredbtan, line 113: getBaos() could be less ambiguous. Makes me hungry.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Deals with user interaction for Duke.
+ */
+public class Ui {
+    public static final String LINE = "    ____________________________________________________________\n";
+
+    /**
+     * Greets the user with default text.
+     */
+    public void greet() {
```
wilfredbtan, line 16: Interesting use of ByteArrayOuputStream to turn system.out.println in to a String. However, could it be better to just get your Ui class methods to return a string and feed it to your handleUserInput() in mainwindow?

src/main/java/task/Priority.java
```diff
@@ -0,0 +1,23 @@
+package task;
+
+import exception.DukeException;
+
+public enum Priority {
```
wilfredbtan, line 5: Good use of enums! 😀

wilfredbtan (commented):
Generally good use of enums. Except maybe can try refactoring to use String instead of print statements. 😀

PR #18, opened 2019-08-23
[Chen Su] Duke Increments

chensu2436:
Increment for Level 1-6.


chensu2436:
Increments up to Level 9.

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+        ui.greet();
+        String fullCommand = ui.readInput(new Scanner(System.in));
+
+        while (!fullCommand.equals("bye")) {
+            try {
+                Parser parser = new Parser(fullCommand);
+                boolean valid = parser.checkValidity();
+
+                // Proceed only if the full command is valid.
+                if(valid) {
```
lightz96: There should be whitespace between if and bracket.

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+                        storage.updateFile(tasks);
+                        break;
+                    }
+                }
+            } catch (DukeException | IOException ex) {
+                System.out.println(ex.getMessage());
+            }
+
+            // Scans the next command line.
+            fullCommand = ui.readInput(new Scanner(System.in));
```
lightz96: Creating one scanner object for each line of text will cause unnecessary overhead.

I feel that you should perform `sc.nextLine()` in this block of code using one scanner object. Or perhaps, you could pass the same scanner object as an argument.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Returns the index of the targeted task when there is a Done/Delete command.
+     * @return integer of the task index.
+     */
+    public int getIndex() {
+        return Integer.parseInt(part[1]);
+    }
+
+    public String getKeyword() {
+        return fullCommand.substring(fullCommand.indexOf(" ")+1);
```
lightz96: Missing whitespace

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Returns true if the full command is a valid command, or false otherwise.
+     * @return boolean variable representing the validity of the command.
+     * @throws DukeException if the command is invalid.
+     */
+    public boolean checkValidity() throws DukeException {
+        switch (getCommandType()) {
+        case "list":
+            break;
+        case "done": case "delete":
```
lightz96: There are missing whitespaces in line 47 and 55. 

According to the coding standards, if there is no `break` statement, you need to include `// fallthrough` comment.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+        list = tasks;
+    }
+
+    /**
+     * Prints out the tasks in the list currently.
+     */
+    public void showTasks() {
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < list.size(); i++) {
+            System.out.println((i+1) + ". " + list.get(i));
```
lightz96: Missing whitespaces in line 37 and 49

src/main/java/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+                break;
+            }
+        }
+        return tasks;
+    }
+
+    /**
+     * Saves the tasks in the list whenever there is any change.
+     */
+    public void updateFile(TaskList tasks) throws IOException {
```
lightz96: Great idea to update your local file after user had added a task.

lightz96 (commented):
Overall, your code is neat, easy to read and had made great use of OOP. However, I noticed some coding standards violations (esp. missing whitespaces).

klementtan:
No specific comments but I like the way you use `switch` it makes the code very readable compared to `if else` statements. 

I see that you are throwing similar errors. You can consider creating more custom error classes to increase abstraction
ie
```
DukeEmptyParamsException(String missingParams, String class)
DukeEmptyParamsException("description", "todo") => "☹ OOPS!!! The description of a todo cannot be empty"
DukeEmptyParamsException("description", "deadline") => "☹ OOPS!!! The description of a deadline cannot be empty"
```

In general great job!

PR #19, opened 2019-08-23
[Larry Law] Duke Increments

larrylawl:


src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        new Duke("/Users/larrylaw/Learn/Uni/Y2S1/CS2103/duke/src/main/data/duke.txt").run();
+    }
+
+    /**
+     * Driver method.
```
shiyao821, line 50: Solid Documentation

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+    }
+
+    public Duke() {
+        tasks = new TaskList();
+    }
+
+    /**
+     * Main method.
+     */
     public static void main(String[] args) {
```
shiyao821, line 45: main method perhaps not needed given GUI launcher

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+    /**
+     * Returns response to be shown to user
+     * @param input
+     * @return response to be shown to user
+     */
+    public String getResponse(String input) {
+        try {
+            Command command = Parser.parse(input);
+            String response = command.execute(tasks, ui, storage);
+            return response;
```
shiyao821, line 80: Simple yet effective implementation :O

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Parses full command string and outputs relevant command.
+     *
+     * @param fullCommand Full Command
+     * @return Command
+     * @throws IllegalArgumentException Command is not valid
+     * @throws ParseException Parsing error
+     */
+    public static Command parse(String fullCommand) throws IllegalArgumentException, ParseException {
```
shiyao821, line 43: Clear breakdown of logic

shiyao821 (commented):
Logical and easy-to-follow OOP structure and neat documentation. Lots I can learn from!

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+import duke.ui.UI;
+
+import java.io.FileNotFoundException;
+import java.text.ParseException;
+import java.util.Scanner;
+
+/**
+ * The one and only Duke!
+ * (Essentially a to do list...)
+ */
```
sandydays, line 14: Haha nice comment!

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
 public class Duke {
+
+    private TaskList tasks;
+    private UI ui;
```
sandydays, line 18: I think it is good coding practice not to name classes by their abbreviations as `UI` in all caps. By convention, better to have just named your class as `Ui`.

src/main/java/Duke.java
sandydays: Agreed, I don't think it is needed as well. In fact, your `run` method down below also should be obsolete and hence can be removed. It is good to remove code you don't use anymore! 

But just curious do you actually make use of the `run` method? And even the `isExit` method that you have in all your Command classes?

src/main/java/Main.java
```diff
@@ -0,0 +1,28 @@
...
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private Duke duke = new Duke("/Users/larrylaw/Learn/Uni/Y2S1/CS2103/duke/src/main/data/duke.txt");
```
sandydays: This is a nice idea. Instead of making an no-argument Duke class constructor, you just used the existing constructor that takes in a String. However, couldn't you have used a relative path, like `"data/duke.txt"`, and in your `Storage` class done something like this to get the parent directory of the path `"data/duke.txt"`:

```java
File file = new File(filePath); // this file path would be your "data/duke.txt"
file.getParentFile().mkdirs();
```

This avoids hard-coding the whole path. If at some point you change your files around, such as moving `CS2103` folder into another sub-folder, for example, then your program would break. Getting the parent path using this method is a better idea.

src/main/java/duke/commands/DoneCommand.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param storage storage
+     * @return Done message or error message
+     */
+    public String execute(TaskList tasks, UI ui, Storage storage) {
+        try {
+            String taskMessage = tasks.done(index);
+            storage.save(tasks.getTasks());
+            return ui.getDoneMessage(taskMessage);
+        } catch (IOException e) {
+            return e.getMessage();
```
sandydays, line 38: This is interesting that you put all code into the `try` block. Do all statements throw the `IOException` or only one of them do? Is this better practice? Because what I did was if I know a piece of code won't throw an error, I didn't put in the `try` block. But I am not sure if it is better to put all of it in the `try` block just because they are in the same method.

src/main/java/duke/parser/Parser.java
sandydays: Agreed! Though I have a suggestion, instead of declaring `Command command;` at the start of your code, and then assigning to it later in your switch statements, you could just return the `new XXXCommand()` in each case. Also avoids the need to have a `break` statement after the body of each case and reduces risk of having an unassigned `command`.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,146 @@
...
+            if (arr.length != 2) {
+                throw new IllegalArgumentException(emptyToDoErrorMessage);
+            }
+
+            command = new ToDoCommand(new ToDo(arr[1]));
+            break;
+        case "deadline":
+            task = arr[1].split("/by")[0].trim();
+            date = readFormat.parse(arr[1].split("/by")[1].trim());
+            command = new DeadlineCommand(new Deadline(task, date));
```
sandydays, line 71: Nice! But I think you can use even more **SLAP**! Instead of creating your deadline here in the `Parser` class and passing it directly to the constructor of `DeadlineCommand`, you could extract the relevant description and date for the deadline here and pass it as arguments to the `DeadlineCommand` constructor, and leave the creation of the `new Deadline(...)` to the constructor of or method in the `DeadlineCommand` class. Feel that would make your code flow better logically!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Parses full task string and returns a task instance.
+     *
+     * @param fullTask Full Task
+     * @return task
+     * @throws ParseException Parsing Error
+     */
+    //TODO: task initialised hard coded!
+    public static Task parseTask(String fullTask) throws ParseException {
+        Task t;
```
sandydays, line 111: Same recommendation as for `command` in your `Parser` class! It would be better to not separate the declaration of the `Task t` from the assignment of the `Task t`. Rather just return the task you create under each if-else condition or set a default value for `Task t` at the top before you reassign it to something below.

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+    public String find(String kw) {
+        StringJoiner result = new StringJoiner("\n");
+        for (int i = 0; i < tasks.size(); i++) {
+            Task t = tasks.get(i);
+            if (t.toString().contains(kw)) {
+                result.add(String.format("%d.%s", i + 1, t));
+            }
+        }
+
+        return result.toString();
```
sandydays, line 106: Nice idea to return the string directly of all the tasks found with the specific keyword instead of return the tasks found and then print them! I did the latter because I felt it was better in terms of code design as I already had a method in place to print the tasks, but I had to add dummy tasks when doing that to also keep index of each found task. Your solution avoids that completely and makes code a lot cleaner!

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package duke.tasks;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
```
sandydays, line 4: Good use of these Java classes for `Date` etc. I wasn't aware I could use them!

src/main/java/duke/ui/UI.java
```diff
@@ -0,0 +1,131 @@
+package duke.ui;
+
+import java.util.StringJoiner;
```
sandydays, line 3: Nice use of already available Java class `StringJoiner`! Feel it is a lot better to use its `add` method instead of `+` for concatenation of strings!

src/main/java/duke/ui/UI.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param message message
+     * @return padded message
+     */
+    public String padMessage(String message) {
+        StringJoiner result = new StringJoiner("\n");
+        result.add("____________________________________________________________");
+        result.add(message);
+        result.add("____________________________________________________________");
+        return result.toString();
```
sandydays, line 127: This is a nice idea to abstract away the adding of the lines above and below Duke's message into a separate method!

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests Parser class.
+ */
+public class ParserTest {
+    private Parser parser;
+
+    @BeforeEach
```
sandydays, line 20: I like that you went beyond doing the usual `@Test`! How does `@BeforeEach` work?

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+    public void parse_listCommand() throws ParseException {
+        String input = "list";
+        parseAndAssertCommandType(input, ListCommand.class);
+    }
+
+    /**
+     * Parses task string correctly.
+     *
+     * @throws ParseException Parse Exception
+     */
```
sandydays, line 40: I like that you added Javadoc comments to your Test class methods too!

sandydays (commented):
Overall, really good job! Code flows nicely and is very easy to read and understand the moving parts! I have also put some minor suggestions that you can take a look at :)

PR #21, opened 2019-08-23
[Ooi Ming Sheng] Duke Increments

ooimingsheng:
CS2103 Individual project increments


SendorasRight:
Looks like you really started on the GUI early. I particularly like your suffix generator for dates, since to my knowledge there exists no such equivalent in Java standard SDK (how did that even happen?). Overall the source code looks great. Here are some of my comments/suggestions:

* Some of your JavaDoc comments don't actually follow the CS2103 Javadoc style (for example, that fact that they must start with a verb in singular form eg. "Returns a value.." instead of "Return a value"). I found a  few examples of these in a few places (eg. Ui, Task etc) while skimming through. It's just minor thing, but you might want to take a note of it.

* Is there a reason why you didn't make your Command and Task classes abstract? Since there is no actual use case where a Command object or a Task object needs to be instantiated directly in your design, I would suggest making them abstract. Although you making the constructor private might achieve the same thing for the current iteration of Duke, declaring a class as abstract is much more readable and make the purpose much clearer to future reviewers/programmers who might be working on your code. Declaring the class as abstract also shifts the responsibility of checking this to the Java compiler/IDE pre-compilation static analyzer.

* Correct me if I'm wrong (I might have skimmed over it by accident), but I'm not sure if your Duke is able to handle invalid inputs - eg. if the user just types "deadline", it seems like your program might crash with an unhandled IndexOutOfBoundsException during one of the parsing methods. It would be good to add such exception handlers to Duke especially since we don't expect a chat assistant to suddenly die on us because we made a typo. Although I think it's fine if you're just trying for the minimal version of the course requirements (nothing wrong with that!).

* Is there a reason why you made a separate taskFactory() method in the Task class for parsing the creation of tasks? Since you already have a Parser class that was made specifically for Duke, I think a better OOP abstraction would be to shift the burden of parsing a Task object to the Parser class, so that it better separates the roles of a Task and a Parser. It would also greatly improve readability as it was initially a little hard for me to follow the logic of the code (which passes the input String from Parser for initial parsing, then into TaskList as a request to add object, then finally into Task to parse again). The burden of parsing a String into a Task sub-class object is spread out over three classes and method calls. Reducing the amount of dependencies would make it easier for you to tweak the parsing code in the future (eg. if a sudden change to the input string format was required, like using "8 done" instead of "done 8").

* Also, it seems like you accidentally overwrote one of the tutorial.md files (textUiTestingTutorial.md) which is why your PR has conflict issues with the cs2103 master. Since this can be resolved from your side without involving the duke repo admins, you might want to look into reverting or replacing that file. (EDIT: my bad, just went back to look at the commit history of the file. Seems like it was just an updated copy that conflicts with the one you're making a PR to.)

kenneth-fung:
You appear to have gotten much of the work done at a surprisingly early date, which is great. Overall, I found the code base to be quite understandable. Here are a few things I found noteworthy:

- This might be a mistake on my part, so correct me if I'm wrong. I noticed that your file path for storing tasks (in ```Duke.java```) seems to be hard-coded to your computer. This might potentially result in the program being unable to save or load files to the disk on any other computer (to my knowledge). Perhaps it would be better if the file path was changed to reference just the app's root directory? To do this, I would suggest replacing your current file path with something like ```.\\data\\duke.txt```.

- There appears to be some leftover code from before you implemented the GUI through JavaFX. Nothing seems to be wrong in terms of functionality (I think, but I'm not 100% sure), but if you have the time, perhaps it might be a good idea to remove the unneeded code? A few examples I noticed included the ```main``` and ```run``` methods in ```Duke.java```, as well as several methods in ```Ui.java``` making calls to ```System.out.println()``` (which, if I understand correctly, are no longer necessary with the GUI). This doesn't need to be particularly high on the priority list, but it might help out any future developers who take a look at your code, especially if they are unfamiliar with the original project.

- I noticed your ```Command``` class has a private constructor to prevent instantiation. Might I suggest making the class abstract instead? All other Command-type classes already inherit from ```Command``` ; together with the need to prevent instantiation, it could prove useful to declare ```Command``` as abstract. Ultimately, however, it does work as it currently is, so this is more a design consideration than anything and it's up to you!

Other than those points, I found the rest of your code base to be alright. Good work!

kenneth-fung (commented):
You appear to have gotten much of the work done at a surprisingly early date, which is great. Overall, I found the code base to be quite understandable. Here are a few things I found noteworthy:

- This might be a mistake on my part, so correct me if I'm wrong. I noticed that your file path for storing tasks (in ```Duke.java```) seems to be hard-coded to your computer. This might potentially result in the program being unable to save or load files to the disk on any other computer (to my knowledge). Perhaps it would be better if the file path was changed to reference just the app's root directory? To do this, I would suggest replacing your current file path with something like ```.\\data\\duke.txt```.

- There appears to be some leftover code from before you implemented the GUI through JavaFX. Nothing seems to be wrong in terms of functionality (I think, but I'm not 100% sure), but if you have the time, perhaps it might be a good idea to remove the unneeded code? A few examples I noticed included the ```main``` and ```run``` methods in ```Duke.java```, as well as several methods in ```Ui.java``` making calls to ```System.out.println()``` (which, if I understand correctly, are no longer necessary with the GUI). This doesn't need to be particularly high on the priority list, but it might help out any future developers who take a look at your code, especially if they are unfamiliar with the original project.

- I noticed your ```Command``` class has a private constructor to prevent instantiation. Might I suggest making the class abstract instead? All other Command-type classes already inherit from ```Command``` ; together with the need to prevent instantiation, it could prove useful to declare ```Command``` as abstract. Ultimately, however, it does work as it currently is, so this is more a design consideration than anything and it's up to you!

Other than those points, as well as what SendorasLeft has mentioned, I found the rest of your code base to be alright. Good work!

PR #22, opened 2019-08-23
[Chow Yi Wai] Duke Increments

chowyiwai:
Add Level-1 to Level-6


t-cheepeng:
Hi Yi Wai,

This iP is really well done! There was clear distinction between the different parts of the software and packages have meaningful names that clearly convey their meaning without having to dig too deep into the documentation. The documentation were also of high quality having different markups in them to make certain parts stand out more than the others. Also, they were descriptive and helps to clarify the motivation and what the method/class does. Also, it was impressive that you have implemented your own Date/Time system that does its own parsing and throws specific relevant exception messages to the user. Could certainly learn from the manual parsing of the Date/Time to throw more specific error messages to user and improve the UX of it. Overall, i think its a job well done!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,86 @@
...
+    public static void main(String[] args) {
+        new Duke("C:\\Users\\Yi Wai\\Documents\\Year 2 Semester 1\\CS2103\\duke\\data\\duke.txt").run();
+    }
+
+    /**
+     * You should have your own function to generate a response to user input.
+     * Replace this stub with your completed method.
+     */
+    public String getResponse(String input) {
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
t-cheepeng: I think its interesting how you have intercepted the output stream of your Ui class to redirect it to your GUI so you did not have to refactor the Ui class. Definitely would not have thought of this

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,74 @@
...
+import duke.command.DoneCommand;
+import duke.command.ExitCommand;
+import duke.command.FindCommand;
+import duke.command.ListCommand;
+import duke.exception.DukeException;
+
+/**
+ * Represents a parser that reads and parses raw command strings.
+ */
+public class Parser {
```
t-cheepeng, line 20: Maybe it would be more meaningful if each check/parse was given a method name of its own so we can quickly find how each command is being parsed.

src/main/java/duke/calendar/Date.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Checks if provided day is valid.
+     * @param inputDay Number corresponding to specific day
+     * @param inputMonthNum Number corresponding to specific month. E.g. 1 for January, 2 for February, etc.
+     * @param inputYear Number corresponding to year.
+     * @return True if day is valid for the given month and year.
+     */
+    protected boolean isValidDay(int inputDay, int inputMonthNum, int inputYear) {
+        switch (inputMonthNum) {
```
t-cheepeng, line 86: What's the purpose of allowing the case to flow? Maybe there is a better way to refactor this portion to better reflect the purpose of this method

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,86 @@
+package duke;
+
+import duke.command.Command;
+import duke.exception.DukeException;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+/**
+ * Represents a personal assistant chat bot. A <code>Duke</code> object corresponds to a specific <code>Storage</code>,
```
weigenie, line 10: I did not know you could do this "<code>". Good to know!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,86 @@
...
+                ui.printExceptionMessage(exception);
+            } finally {
+                ui.printLine();
+            }
+        }
+
+    }
+
+    public static void main(String[] args) {
+        new Duke("C:\\Users\\Yi Wai\\Documents\\Year 2 Semester 1\\CS2103\\duke\\data\\duke.txt").run();
```
weigenie: Instead of using a fixed filepath, perhaps you can consider using a filepath from the project root instead.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Parses raw commands and returns a type of <code>Command</code> correspondingly.
+     * @param rawCommandDetails Details of the command provided by user.
+     * @return A type of <code>Command</code> that corresponds to the raw command.
+     * @throws DukeException If the raw commands are invalid, in the wrong format or are lacking required information.
+     */
+    static Command parse(String rawCommandDetails) throws DukeException {
+        String commandDetails = rawCommandDetails.replaceAll("\\s+", " ");
+        ;
+        if (commandDetails.equals("bye")) {
```
weigenie: Perhaps you can use switch cases instead

src/main/java/duke/calendar/Date.java
```diff
@@ -0,0 +1,147 @@
+package duke.calendar;
+
+import duke.exception.DukeException;
+
+import java.time.Month;
+
+/**
+ * Represents a date. A <code>Date</code> object corresponds to a time represented by a day, month and year.
+ */
+public class Date {
```
weigenie: Interesting that you used your own Date class, but perhaps you can use Date from java.util.Date instead.

src/main/java/duke/calendar/Date.java
weigenie: You can use SimpleDateFormat instead to parse the String for a Date object, skipping all these.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,27 @@
...
+import duke.TaskList;
+import duke.Ui;
+import duke.exception.DukeException;
+
+/**
+ * Represents a command.
+ * Abstract parent class of all specific types of <code>Command</code>.
+ * A <code>Command</code> object corresponds to an action that can be executed.
+ */
+public abstract class Command {
```
weigenie, line 13: I like how you used Command as an abstract class for other child command classes.

weigenie (commented):
In general, I think that your code is very neat and tidy, adhering to most, if not all, of the OOP principles. Great job!

PR #24, opened 2019-08-23
[Keng Jun Xian] Duke Increments

jxofficial:
Completed Level 1 - 9


cheongliyuan:
Hi Jun Xian,

I did not find any obvious coding standard violations. I really enjoyed reading through your code! It is very readable, with clear level of abstraction between Duke>Command>Ui.  The _show()_ method in Ui Class is pretty ingenious, reducing the _println("  ___")_ code repetition. Very cool!
I like to ask: why are the _command_ and _exception_ methods all in different java files? Is it a preference, or it is following a standard file structure? (Edit: Just realised this is mentioned in the extra-OOP part. opps!) While the _Tasklist_ methods are all in the 1 java file. My guess is that you are trying to balance between each file's LoC and number of files in your project? 
I also like the output given for possible accepted date formats, when the input date isn't as expected.
Looking forward to hearing your thoughts.
-Li Yuan

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public abstract void execute(TaskList tasks, UI ui, Storage storage) throws DukeException;
+
+
+    /**
+     * Informs the caller if the command should result in program termination
+     *
+     * @return A boolean specifying if program should terminate after command execution
+     */
+    public abstract boolean isExit();
```
wpuffs: I noticed that for most of the command classes (i.e. CreateDeadlineCommand, CreateEventCommand, etc.), the isExit() method returns false. Would it be easier to just implement the isExit() method in the Command class and just have the ExitCommand class override the isExit() method to return true?

src/main/java/duke/task/TypeOfTask.java
```diff
@@ -0,0 +1,20 @@
+package duke.task;
+
+/**
+ * Enum that contains the various types of tasks accepted by Duke
+ * e.g.,<code>todo</code>, <code>deadline</code>, <code>event</code>
+ */
+public enum TypeOfTask {
```
wpuffs, line 7: This isn't really a suggestion but just a question! I'm just curious to know why you created an enum class for the different types of task if there are already classes for the different types of task which inherit the Task class.

wpuffs (commented):
Hey Jun Xian,

Overall a job well done! I liked how there was extensive OOP usage and how you've packaged the related classes together. You've also done a good job at documenting your code, making it easy to understand. I do have a few questions regarding certain parts of your code, and I have made comments on the specific lines 😄

C-likethis123:
Hi Jun Xian, 

In general, I like that the Javadoc comments were clear and how you packaged your classes. I do have some questions regarding your code. 

1. Why do you have an InvalidCommand class? Would it be easier to throw an InvalidCommandException in the parse() method?

2. I noticed that there are some lines of your code where you would refer to instance-level variables as this.<variable>, even though there is no shadowing. Would it be easier to just omit the 'this' keyword?

-Jia Ying

PR #27, opened 2019-08-23
[Goh Rui Zhi] Duke Increments

mirozo:
Increments level 1 - 9


src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+    /**
+     * Converts a Date object to a String object in dd/MM/yyyy HHmm format
+     *
+     * @param date Date object
+     * @return String
+     */
+    public static String dateToStringConverter(Date date) {
+        SimpleDateFormat sdf = new SimpleDateFormat("dd/MM/yyyy HHmm");
+        String sDate = sdf.format(date);
+        return sDate;
```
geshuming: This is a really weird place to implement this function. There is cyclic dependency as a result i.e. `Duke` relies on `Storage` and `Storage` relies on `Duke`. Considering it is a static function anyway, maybe you can move this into the `Storage` class, or a `Utilities` class if multiple classes require it?

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+    private TaskList tasks;
+    private Ui ui;
+
+    public Duke(String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.getTasks());
+        } catch (Exception e) {
+            ui.showLoadingError();
```
geshuming, line 16: As a fallback, maybe you can continue the process by initializing `tasks` with an empty `ArrayList`?

src/main/java/Storage.java
```diff
@@ -0,0 +1,125 @@
...
+        return f;
+    }
+
+    /**
+     * Updates the .txt with the most recent changes.
+     *
+     * @param filePath location of the file as a string.
+     * @param list The existing ArrayList of Tasks after user has made changes to it.
+     */
+    public void updateTasks(String filePath, ArrayList<Task> list) throws IOException {
```
geshuming: I think it is a bit weird that you need a new `filePath` for the instance method when the instance is constructed with a `filePath`. Why don't you use the instance's `filePath`?

src/main/java/Storage.java
```diff
@@ -0,0 +1,125 @@
...
+        FileWriter fw = new FileWriter(filePath);
+        String result = "";
+        for (int i = 0; i < list.size(); i++) {
+            Task task = list.get(i);
+            String c = task.getClass().toString();
+            if (c.equals("class Task.Todo")) {
```
geshuming: Can I recommend using `instanceof` keyword to check class type? It is more reliable and easier to implement than string comparison. [Reference](https://www.javatpoint.com/downcasting-with-instanceof-operator)

geshuming (commented):
Overall looks good.

src/main/java/Parser.java
```diff
@@ -0,0 +1,119 @@
...
+        for (int i = 1; i < words.length; i++) {
+            if (i == words.length - 1) {
+                taskDetails += words[i];
+            } else {
+                taskDetails += words[i] + " ";
+            }
+        }
+        this.taskDetails = taskDetails;
+    }
+    /**
```
irene-lly: Remember to add a line here to separate the constructor and the java docs.

src/main/java/Parser.java
```diff
@@ -0,0 +1,119 @@
...
+        if (taskDetails.equals("")) {
+            if (command.equals("bye") || command.equals("list") ) {
+
+            } else {
+                ui.showDescriptionEmptyError();
+                return;
+            }
+        }
+        switch (command) {
+            case "list":
```
irene-lly: Cases in switch should not be indented (for this module).

src/main/java/Storage.java
irene-lly: Yah I also recommend using instanceof here it can improve readability.

irene-lly (commented):
Overall good effort, the way the methods are named makes it very easy to follow the code, everything is also very well organised into their respective classes and packages. There are some however styling errors and I would recommend putting header comments to all public classes too.

src/main/java/Storage.java
mirozo: Good idea! Thanks

src/main/java/Storage.java
mirozo: Good idea! Thanks

src/main/java/Duke.java
mirozo: Good idea! Thanks

PR #28, opened 2019-08-23
[Zi Wei] Duke Increments

halfwhole:
Add Level-1 to Level-6


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,77 @@
...
+    public static void main(String[] args) {
+        new Duke().run();
+    }
+
+    /**
+     * Returns whether the program should exit
+     * @param input Input string
+     * @return Boolean indicating of program should exit
+     */
+    public boolean respond(String input) {
```
SimNee, line 77: This is a very good method to use. Because the website itself only show the hint for the front part and you adding this is a very smart move :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+            } else {
+                this.slashKeyword = inputAfterSlash.split("\\s+", 2)[0];
+                this.slashArgs = inputAfterSlash.split("\\s+").length >= 2
+                        ? inputAfterSlash.split("\\s+", 2)[1]
+                        : null;
+            }
+        }
+    }
+
+    private static void validateIsPositiveInteger(String s, String message) throws DukeException {
```
SimNee, line 56: Maybe this could be in the UI since it is interpreting the content of the input from users to see if there is any errors? I am not very sure but I think this is very minor :)

SimNee:
Overall, I feel that this is a very good set of codes like it is almost perfect but I added in some personal opinions as single comments inside. 
There were a great use of OOP, inheritance and abstraction. Thank you :)

src/main/java/duke/Duke.java
ernestyyh: This seems to be a good method to use for the CLI. Will it be applicable as well for the GUI?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+    public static class ParsedCommand {
+        public final String input;
+        public final String keyword;
+        public final String args;
+        public final String beforeSlashArgs;
+        public final String slashKeyword;
+        public final String slashArgs;
+
+        /**
+         * Creates a parsed command, whereby the input command as a string is parsed into its components.
```
ernestyyh, line 25: This is well-crafted documentation of your method.

PR #29, opened 2019-08-23
[Kwan Chen Sheng] Duke Increments

mrchensheng21:


czkay:
I think your code was pretty easy to follow, and most of your methods were clean and not too large. Thanks for that!

Regarding your command.contains(...) lines in your Duke.java, perhaps change it to .startsWith() or use regex instead? If not, "blahblahblah todo" could end up being seen as a Todo task by your program.
You can import LocalDateTime or Calendar to help format your deadline/event timings for you, then you will be able to cut out lots of your long getDate() method.
You can also cut out a lot of code if you use an ObjectInput/OutputStream to directly write/load your TaskList object to your hard drive, instead of formatting your tasks manually into a .txt file.
And lastly, I think there's supposed to be an empty line between your method description and your @param line in your Javadoc headers.

Overall nice job, thanks!

src/main/java/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Store the tasks list into a .txt file after User exits the Duke application.
+ * It also loads up the tasks list from the selected .txt file when User starts up
+ * the Duke application.
+ */
+public class Storage {
+    private String filepath;
+    private ArrayList<Task> taskList;
+    /**
```
marcusteh1238, line 17: There should be a extra empty line between the attribute declarations and the first function declaration.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,111 @@
...
+        } else {
+            String[] splitString = command.split(" ");
+            int taskDone = Integer.valueOf(splitString[1]);
+            taskList.get(taskDone - 1).markAsDone();
+            return taskList.get(taskDone - 1).getDescription();
+        }
+    }
+
+    /**
+     * Other classes can retrieve the tasks list
```
marcusteh1238: Grammar error: should be "task list".

marcusteh1238 (changes requested):
The code is easy to follow. It also seems to be clear of any coding style violations, apart from an error here and there, so good job regarding that! However, I think it would be good if you could handle more errors so users would not crash out of the program if they input some bad command. For instance, I tried to input the bad command `deadline test /at 123` and it threw a `ArrayIndexOutOfBoundsException`. To help with debugging, you should also write more specific `DukeException` messages, rather than having it be "Something went wrong!".

Additionally, for the `getDate()` method in `Deadline.java` I think it would make your code a lot simpler if you made use of the `Integer.parseInt()` function to parse integers rather than using 12 case/switches. Here is the [Java documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#parseInt(java.lang.String)) for it!

You could also use the `SimpleDateFormat` java class to make your life way easier. You should also consider using regex to simplify your code from lines 36 to 52. Take a look at [my implementation](https://github.com/nus-cs2103-AY1920S1/duke/pull/185) of Deadline.java if you want to see how I handled it, and the [Java documentation](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) for it.

For your `Events` class, the `at` attribute is pretty redundant as it is only used in the constructor. As the `datetime` attribute is used more often, you could consider removing the `at` attribute entirely and just leave it as a parameter in the constructor for producing the `datetime` attribute.

I highly recommend that you add Gradle to your project, as it really helps you to automate tests, finding bugs in the most unexpected of places, and checks your code style with a simple command.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,39 @@
+import java.text.ParseException;
+
+
+/**
+ * Creates a Deadline object which extends from the Task class.
+ * A <code>description</code> is passed into this class to
+ * instantiate a Deadline task.
+ */
+public class Deadline extends Task {
+    private String[] deadline;
```
crazoter, line 10: This is vague. What is deadline? Is it the description?

src/main/java/Events.java
```diff
@@ -0,0 +1,38 @@
+import java.text.ParseException;
+
+/**
+ * Creates a Events object which extends from the Task class.
+ * A <code>description</code> is passed into this class to
+ * instantiate a Events task.
+ */
+public class Events extends Task {
```
crazoter, line 8: Hello :)
Perhaps you can put your class name in singular rather than plural. Because when I use the class, I'm making an Event, not an "Events".

src/main/java/Events.java
```diff
@@ -0,0 +1,38 @@
...
+    private String event[];
```
crazoter: Hello :)

I don't know what events mean as a variable. Perhaps you could be clearer.

src/main/java/Parser.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Making sense of the User's input, and to throw an IllegalCommandException when User enter an invalid
+     * command.
+     * @param command Inputs of the User.
+     */
+    protected String parse(String command) {
+        try {
+            if (command.contains("todo") || command.contains("deadline") || command.contains("event")) {
```
crazoter, line 28: Hello :)

You may want to use String.startsWith() method instead of the String.contains() method. I've not seen the rest of the code, but I assume from the code here if type "haha todo" it will create a todo task.

src/main/java/Parser.java
```diff
@@ -0,0 +1,57 @@
...
+            } else if (command.contains("done")) {
+                String taskDoneStr = this.tasksList.doneTask(command);  // retrieve the task that is done.
+                return ui.getDoneMessage(taskDoneStr);
+            } else if (command.contains("list")) {
+                return ui.showList(this.tasksList.getTaskList());
+            } else if (command.contains("find")) {
+                ArrayList<Task> foundTask = this.tasksList.findTasks(command);
+                return ui.showFoundMessage(foundTask);
+            } else {
+                return "I'm sorry, but I don't know what that means :-(";
```
crazoter, line 46: Hello :)

This is a magic variable. May want to put as a constant somewhere e.g. at the top or in a class.

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Store the tasks list into a .txt file after User exits the Duke application.
+ * It also loads up the tasks list from the selected .txt file when User starts up
+ * the Duke application.
+ */
+public class Storage {
+    private String filepath;
```
crazoter, line 14: Hello :)

In one of your methods below you referred to filepath as "filePath" as a parameter. Both are okay, but would be nice to be consistent.

P.S: Would be nice to separate static methods and variables with object methods and variables. Moving this object variable after all the static methods and variables would reduce confusion by readers.

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /**
+     * Save the tasks into the .txt file allocated in the directory.
+     * @param filepath Directory for the .txt file.
+     * @param taskList ArrayList of tasks that the User inputs.
+     * @throws IOException if there is an error of copying the list to the file.
+     */
+    private static void WriteTextToFile(String filepath, ArrayList<Task> taskList)
```
crazoter: Hello :)

As mentioned above would be nice to make consistent filepath and filePath

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+    private ArrayList<Task> taskList;
+    /**
+     * Load the file contents of the .txt file and to input the saved tasks into
+     * an Arraylist.
+     * @param filePath Directory of the .txt file.
+     * @param storingList An Arraylist to store the tasks in the .txt save file.
+     * @throws FileNotFoundException if there are no .txt file found in the directory
+     * @throws DukeException if the .txt file is an empty file
+     */
+    private static void loadFileContents(String filePath, ArrayList<Task> storingList)
```
crazoter: Hello :)

Would be nice to see List instead of ArrayList used; programming to an interface is encouraged. This is not a big issue though :)

src/main/java/Task.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * A flag to toggle when a task is done.
+     */
+    protected void markAsDone() {
+        taskIsDone = true;
+    }
+
+    /**
+     * Get the date from a numeric format eg. (23/05/2019) and convert it
```
crazoter, line 75: For JavaDocs your comment should have a space separating description and the @params @return etc.

src/main/java/ToDos.java
```diff
@@ -0,0 +1,25 @@
+/**
+ * Creates a ToDos object which extends from the Task class.
+ * A <code>description</code> is passed into this class to
+ * instantiate a todos task.
+ */
+public class ToDos extends Task {
```
crazoter, line 6: Hi :)

ToDos perhaps can be named "Todo".

src/main/java/Ui.java
```diff
@@ -0,0 +1,117 @@
...
+        return scanner.nextLine();
+    }
+
+    /**
+     * Display the list of tasks to the User.
+     * @param taskList The tasks list.
+     * @return the tasks list.
+     */
+    protected String showList(ArrayList<Task> taskList) {
+        String showList = "";
```
crazoter, line 49: Hi :)

Perhaps this can be put into a ShowListCommand or something.

crazoter (commented):
Greetings :)

I have reviewed your code together with my teammates during tutorial. Generally readable code but it can be improved.

Thank you :)

PR #37, opened 2019-08-23
[Zhang Xuan] Duke Increments

zhangxuan97:


src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
+import command.Command;
+import main.*;
```
TeoShyanJie: I think maybe can try to import the specific task

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,81 @@
+package command;
+
+import main.DukeException;
+import main.Storage;
+import main.TaskList;
+import main.Ui;
+import task.*;
```
TeoShyanJie: I think maybe try to import specific task class

TeoShyanJie (changes requested):
I have left some comment regarding the import of class. I think the overall program is really good and have a good OOP.

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,41 @@
...
+     * Executes the command to mark a task as complete.
+     *
+     * @param tasks     The existing task list
+     * @param ui        The Ui object which interacts with the current user
+     * @param storage   The Storage object which reads and writes to a specified file
+     * @throws DukeException If task ID provided is out of bounds
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        if (this.taskID > tasks.size()) {
+            throw new DukeException("Invalid ask ID. Please Enter a task ID between 1 and " + tasks.size());
```
Auxinnn: Does 'ask' means 'task'?

Auxinnn (commented):
I think your codes are really good with nice abstraction.

src/main/java/Duke.java
zhangxuan97: Thanks for pointing it out 👍

src/main/java/command/DoneCommand.java
zhangxuan97: Oh yea it does! Thanks for pointing it out 👍

PR #38, opened 2019-08-23
[Jessica Phua Shu Xin] Duke Increments

jessicax941:


src/main/java/TaskList.java
```diff
@@ -0,0 +1,57 @@
...
+        taskDone.markAsDone();
+        System.out.println("\tNice! I've marked this task as done:\n\t\t" + taskDone.toString());
+    }
+
+    public void deleteTask(int taskNumber) {
+        Task task = taskList.remove(taskNumber - 1);
+        printDeletedTask(task);
+    }
+
+    private void printAddedTask(Task task) {
```
jeunhoe: I understand that you did the printing in the TaskList class as you might not want to introduce unnecessary dependencies to other classes. However, I'm not sure it is appropriate for the list to do the printing. I'm unsure myself of how to handle the printing but what do you think of printing classes directly through the command itself, or doing the printing through the Ui (introducing a method that takes in a Command as it's argument)?

src/main/java/Ui.java
```diff
@@ -0,0 +1,40 @@
...
+    public void takeInUserInput() {
+        Scanner scanner = new Scanner(System.in);
+        Parser parser = new Parser();
+        while (scanner.hasNext()) {
+            String command = scanner.nextLine();
+            command = command.trim();
+
+            try {
+                parser.executeCommand(command);
+            } catch (InvalidCommandException | MissingInputException | MissingDescriptionException e) {
```
jeunhoe: Great use of inheritance for exception handling! Makes your code look very neat.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public LocalDateTime getDeadline() {
+        return this.deadline;
+    }
+
+    public void setDeadline(String deadline) {
+        this.deadline = stringToDateTime(deadline);
+    }
+
+    private LocalDateTime stringToDateTime(String deadline) {
```
jeunhoe: Great use of existing Java libraries to handle date & time formats! I manually coded the string output for every month and day which made my code look very messy... Learnt something new!

src/main/java/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+                int taskNumber2 = Integer.parseInt(commandWords[1]);
+                taskList.deleteTask(taskNumber2);
+                break;
+            case "todo":
+            case "deadline":
+            case "event":
+                addTask(command, commandType);
+                break;
+            case "bye":
+                Ui ui = new Ui();
```
jeunhoe: This line of code causes the execution of Duke to create 2 UIs, one on start of Duke and one when "bye" is called. The code should work but I feel like it would cause a bit of confusion in the future, especially when you add on extensions to the UI. Is there anyway you can pass the execution of exit back to the first UI instead of creating a new one?

jeunhoe (commented):
Great pull request, learnt a lot good practices from reviewing.

src/main/java/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+                break;
+            case "done":
+                int taskNumber = Integer.parseInt(commandWords[1]);
+                taskList.markAsDone(taskNumber);
+                break;
+            case "delete":
+                int taskNumber2 = Integer.parseInt(commandWords[1]);
+                taskList.deleteTask(taskNumber2);
+                break;
+            case "todo":
```
joannasara: I just realize you can do this, thanks!

src/main/java/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public Parser() {}
+
+    public void executeCommand(String command) throws InvalidCommandException, MissingInputException,
+            MissingDescriptionException {
+        String[] commandWords = command.trim().split(" ");
+        String commandType = commandWords[0];
+        TaskList taskList = new TaskList();
+        switch (commandType) {
+            case "list":
```
joannasara: You forgot to remove the indentation here

src/main/java/Storage.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public ArrayList<Task> load() throws FileNotFoundException {
+        File file = new File(filePath);
+        ArrayList<Task> list = new ArrayList<>();
+        Scanner sc = new Scanner(file);
+        while (sc.hasNext()) {
+            String line = sc.nextLine(); // eg. T,0,read book
+            String[] taskDetails = line.split(",");
+            String taskType = taskDetails[0];
+            int booleanInt = Integer.parseInt(taskDetails[1]);
```
joannasara: Just a matter of preference, but I think its better to change this into boolean directly here

PR #39, opened 2019-08-23
[Koh Zheng Qiang Shawn] Duke Increments

shawnkoh:
Let's go! :)


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,56 @@
+package duke;
```
pohlinwei, line 1: It's great that you packaged your classes (:

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,56 @@
...
+     * Runs the {@link Duke} application.
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                Command c = Parser.parse(fullCommand);
+                c.setData(tasks, ui, storage);
```
pohlinwei: It might be me. However, I feel that `setData` is unnecessary since you always call `execute` afterwards. Instead, it might be more convenient if you just use `execute` and it adds a layer of abstraction to your code. Anyway, this isn't a huge issue. (:

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return {@link Command} representing the command.
+     * @throws DukeException if the command has parser errors.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        String[] arr = fullCommand.trim().split(" ", 2);
+        String commandWord = arr[0];
+
+        try {
+            switch (commandWord) {
+            case DeadlineCommand.COMMAND_WORD:
```
pohlinwei: This is an interesting way of doing things. However, why not consider using `enum` instead? If I had no access to the code for the other classes and I looked at this, it is a little confusing.  

Besides, using `enum` will save you a lot of code (you don't have to keep typing `public static final String COMMAND_WORD`. Isn't that great? :D

src/main/java/duke/command/DeadlineCommand.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.Date;
+
+/**
+ * Represents a {@link Command} to add a {@link Deadline}.
+ */
+public class DeadlineCommand extends AddCommand {
+
+    public static final String COMMAND_WORD = "deadline";
+
+    public DeadlineCommand(String description, Date by) {
```
pohlinwei: `by` seems a little ambiguous to me

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,29 @@
...
+import duke.exception.DukeException;
+import duke.task.Task;
+
+/**
+ * Represents a {@link Command} to delete a {@link Task}.
+ */
+public class DeleteCommand extends Command {
+
+    public static final String COMMAND_WORD = "delete";
+    private final int toDelete;
```
pohlinwei, line 19: Nice use of `final` 😄

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + description;
+    }
+
+    /**
+     * Returns a stringified version of the {@link Task} that is meant to be used in the {@link duke.storage.Storage}.
+     *
+     * @return a stringified version of the {@link Task} that is meant to be used in the {@link duke.storage.Storage}.
+     */
+    public String stringify() {
```
pohlinwei, line 52: The method name `stringify` is good. However, I feel that in the context where this method is called, `stringify` doesn't seem to convey what the method actually does. 😅

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Date;
+
+/**
+ * Represents an event as a {@link Task}.
+ */
+public class Event extends Task {
+    protected Date at;
+
+    public Event(String description, Date at) {
```
pohlinwei: `at` might come across as ambiguous

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,85 @@
...
+        return sb.toString();
+    }
+
+    /**
+     * Returns a new {@link TaskList} with {@link Task}s that contains the keyword.
+     *
+     * @param keyword the string to search for.
+     * @return a new {@link TaskList} with the filtered copy.
+     */
+    public TaskList find(String keyword) {
```
pohlinwei, line 79: It is interesting that you returned a new task list when user wants to `find`! (: I wouldn't have thought of that

src/test/java/task/TestDeadline.java
```diff
@@ -0,0 +1,29 @@
...
+    private Deadline deadline;
+
+    @BeforeEach
+    public void setUp() throws DukeException {
+        deadline = new Deadline("return a book", DateUtil.parse("1/12/2019 1800"));
+    }
+
+    @Test
+    public void toString_normal_noException() {
+        assertEquals(deadline.toString(), "[D][✘] return a book (by: 01/12/2019 1800)");
```
pohlinwei: This is not a major problem. However, according to the syntax for `assertEquals`, the expected output should be the first argument while the actual output should be the second argument. 😅

pohlinwei (commented):
Great job! 😄 Overall, great Javadoc and good coding style. I like the use of access modifiers. However, do take note of how you use `assertEquals`. (I have pointed it out for one line of your code but it applies to the rest as well) Well done (:

src/main/java/duke/task/Task.java
shawnkoh: Haha i agree, I was trying to think of a better name but resorted to using stringify following how JSON did it https://www.w3schools.com/js/js_json_stringify.asp

src/main/java/duke/Duke.java
shawnkoh: Just in case we ever have to perform some mutations on the command before executing it :)

src/main/java/duke/parser/Parser.java
shawnkoh: https://stackoverflow.com/questions/9969690/whats-the-advantage-of-a-java-enum-versus-a-class-with-public-static-final-fiel
There’s an entire discussion on this here - it is true that enums have advantages, but I don’t really need them in this context. I wanted to keep all the relevant information about a Command in its implemented class, but I agree with your points.
On hindsight, using enums will allow me to point other command words to a Command - but I’m not sure if that’s a good thing to have since it can be misleading.

PR #43, opened 2019-08-23
[Matthew Lee] Duke Increments

crazoter:


jcjjjared (commented):
Hi, your code is very clean and the classes are well-organised! Keep up the good work. Generally no syntax violations too! Easy to read and understand.😄 Cheers!

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void execute(Duke duke, TaskList taskList, Ui ui, Storage storage) throws DukeException {
+        ui.println("____________________________________________________________\n"
+                + "Here are the matching tasks in your list:\n");
+        //I'm assuming index matches the index in the list
+        for (int i = 0; i < taskList.size(); i++) {
+            String taskDetails = taskList.get(i).toString();
+            if (taskDetails.contains(remainingLine)) {
+                ui.println(i + "." + taskDetails);
```
dalisc: It wasn't explicitly specified in the instructions whether to assign new indexes to the matching tasks (numbering them from 1 again), but it does imply that in the instructions.

As for me, I kept the original indexes of the task so that the user can more easily refer to their index if they want to issue a Done or Delete command.

The instructions are ambiguous and this is something I will clarify with Prof and my tutor, so just a heads up to keep a lookout for clarifications regarding this!

src/main/java/duke/command/IndexBasedCommand.java
```diff
@@ -0,0 +1,17 @@
+package duke.command;
+
+import duke.*;
```
dalisc: Do avoid using * in imports!

dalisc:
Hi! Your code is very clean and readable. I like that you went the extra mile to do the optional increments of making it even more OOP style and packaging them.

I have pointed out some small errors such as the use of asterisks. Nothing serious - just coding standard stuff. Also, commit messages should be in present tense in an imperative style (as taught in Week 3 lecture), so that is something to take note of in the future. Aside from that, you could also write documentation at the class level when time allows. 

All in all, really neat and solid work! Keep it up!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,110 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Ui {
+    private boolean enabled;
```
jcjjjared: Could use "isEnabled" instead to make it sound more like a boolean

PR #47, opened 2019-08-23
[Chan Kok Jing, Daryl] Duke Increments

dvrylc:


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,50 @@
+package duke;
+
+import java.util.Date;
+
+class Deadline extends Task {
```
yan-wl: Nice

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,110 @@
+package duke;
+
+import java.text.SimpleDateFormat;
+
+public class Duke {
+    // Configuration
+    static final SimpleDateFormat dateFormatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
yan-wl: I believe this identifier should be all caps as it is a constant

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,110 @@
...
+                    System.exit(0);
+                    break;
+                }
+
+                case "list": {
+                    this.ui.showTasks(this.tasks.toString());
+                    break;
+                }
+
+                case "done":
```
yan-wl: Needs a _//fallthrough_ to comply with coding standards/style

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,106 @@
...
+     * Access underlying ArrayList for export purposes.
+     *
+     * @return ArrayList of tasks.
+     */
+    ArrayList<Task> getTasks() {
+        return this.tasks;
+    }
+
+    /**
+     * Generates a numbered list of tasks matching a query.
```
yan-wl: This description does not seem to tally with what the method is doing

yan-wl (commented):
Good job overall!

src/main/java/duke/TaskList.java
dvrylc: Hmm, could you elaborate on this please?

yan-wl: Ah, I was mistaken. The return description gave me the impression that it would return a List object instead of the String representation of a list.

PR #49, opened 2019-08-23
[Ellie Yee Wen Na] Duke Increments

Ellieyee:


build.gradle
```diff
@@ -0,0 +1,38 @@
...
+version '0.1.0'
+
+repositories {
+    mavenCentral()
+}
+
+application {
+    // Change this to your main class.
+    //mainClassName = "seedu.duke.Duke"
+    mainClassName = "Duke"
```
ang-zeyu: Use of default package can result in naming collisions when (for e.g. importing an external library that also uses the default package )

From how your code is structured, I don't think the following applies, but I ran into it early on in my implementation enough to follow it.
Quote from 2103 week 3 topics : "_If you do not use a package statement, your type doesn't have a package -- a practice not recommended (except for small code examples) as it is not possible for a type in a package to import a type that is not in a package._"

A stackoverflow post summarising the two points:
https://stackoverflow.com/questions/7849421/is-the-use-of-javas-default-package-a-bad-practice

src/main/java/DataStorage.java
```diff
@@ -0,0 +1,84 @@
+import duke.task.*;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+public class DataStorage {
+    private static final String PATH = "C:\\Users\\ellieyeewenna\\Desktop\\Ellie\\NUS\\CS2103\\IP\\data\\duke.txt";
```
ang-zeyu: Good idea to use Paths api to make the storage path platform independent!

What if the user's username is not 'ellieyeewenna' though? 
( or if the os isn't windows, Paths library won't be able to resolve "C:\\.." )
What you can do:
- use System.getProperty("user.dir") to find the present working directory of the application
- use Paths api's Paths.get(\<pwd from above\>, \<desired relative path to data directory\>)

src/main/java/UI.java
```diff
@@ -0,0 +1,17 @@
+public class UI {
```
ang-zeyu: Seems to me to be a class of _static UI-related utility methods_.

What if you want to change the application to use a gui (Jfx) though?
Each of your calls to System.out.println would have to be refactored to use the gui's call to display a message.

Instead, UI could be a more concrete (non-utility, maybe just housing some) instance abstraction of a user interface, where you would only have to make the gui library implement / extend the UI class. You would then only have to change the passed UI instance into the commands / parser / exceptions.

Unrelated to duke \:
Another benefit to this is the possibility to easily implement multiple UI instances ( think file > open in a new window ) in applications that need this!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * Adds given task into tasklist
+     * @param taskList - list containing all existing tasks
+     */
+    @Override
+    public void execute(TaskList taskList) {
+        taskList.add(this.task);
+        this.printSuccessfulAddMessage(taskList.size());
```
ang-zeyu: From level-7 task requirements \: _'Save the tasks in the hard disk automatically **whenever** the task list changes.'_

Just add your DataStorage.storeTaskList(taskList) command here! (Would need to be in a package to import)
Otherwise a clean implementation 👍 

_Suggestion\:_
Since many of the commands need a save to disk operation after executing, 
What I did was use a WritableCommand extends Command abstraction that has 3 methods - execute, run and validate. Execute is implemented to run \: `validate(); run(); storage.saveToDisk(taskList)`.
(ignore the validate() call since your validation logic is inside your parsers)
Then simply extend from WritableCommand for any tasks that need saving after running.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Execute delete command on given task and save into tasklist
+     * @param taskList - list containing all existing tasks
+     */
+    @Override
+    public void execute(TaskList taskList) {
+        Task task = taskList.get(this.idx);
+        taskList.delete(idx);
+        this.printSuccessDeleteTaskMessage(task, taskList.size());
```
ang-zeyu: _Same as before:_ 

From level-7 task requirements : 'Save the tasks in the hard disk automatically whenever the task list changes.'

Just add your DataStorage.storeTaskList(taskList) command here! (Would need to be in a package to import)
Otherwise a clean implementation 👍

_Suggestion\:_
Since many of the commands need a save to disk operation after executing, 
What I did was use a WritableCommand extends Command abstraction that has 3 methods - execute, run and validate. Execute is implemented to run \: `validate(); run(); storage.saveToDisk(taskList)`.
(ignore the validate() call since your validation logic is inside your parsers)
Then simply extend from WritableCommand for any tasks that need saving after running.

src/main/java/duke/command/DoneCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Execute done command on given task and save into tasklist
+     * @param taskList - list containing all existing tasks
+     */
+    @Override
+    public void execute(TaskList taskList) {
+        Task task = taskList.get(idx);
+        taskList.done(this.idx);
+        this.printSuccessfulDoneMessage(task);
+    }
```
ang-zeyu, line 36: _Same as before:_

From level-7 task requirements : 'Save the tasks in the hard disk automatically whenever the task list changes.'

Just add your DataStorage.storeTaskList(taskList) command here! (Would need to be in a package to import)
Otherwise a clean implementation 👍

_Suggestion\:_
Since many of the commands need a save to disk operation after executing, 
What I did was use a WritableCommand extends Command abstraction that has 3 methods - execute, run and validate. Execute is implemented to run \: `validate(); run(); storage.saveToDisk(taskList)`.
(ignore the validate() call since your validation logic is inside your parsers)
Then simply extend from WritableCommand for any tasks that need saving after running.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class FindCommand extends Command {
+    private String taskName;
+
+    /**
+     * Constructor
+     * @param taskName - Name of task to search for
+     */
+    public FindCommand(String taskName) {
+        super(null);
```
ang-zeyu: I noticed many of your command abstractions make calls to `super(null)`.

Maybe it would be easier to introduce an additional layer of abstraction, say, 'AddTaskCommand extends Command' and only commands that make use of the task instance variable would have it.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        this.taskName = taskName;
+    }
+
+    /**
+     * Execute find command on given task and save into tasklist
+     * @param taskList - list containing all existing tasks
+     */
+    @Override
+    public void execute(TaskList taskList) {
```
ang-zeyu: Effective and clean implementation! 👍

src/main/java/duke/exception/IncompleteCommandError.java
```diff
@@ -0,0 +1,22 @@
+package duke.exception;
+
+public class IncompleteCommandError extends UnknownCommandException {
```
ang-zeyu: Regarding packaging exceptions into a package:

I think this stackoverflow post would explain it better than me:
(Examples: java.**io**.IOException, java.**sql**.SQLException)
https://stackoverflow.com/questions/33480666/java-custom-exceptions-package-convention

ang-zeyu: Regarding errors vs exceptions \:
https://stackoverflow.com/questions/5813614/what-is-difference-between-errors-and-exceptions

(e.g. stackOverflowError is something that is unreasonable for the application to catch since it depends on the defined  maximum stack size of the platform the application runs on ) ( assuming a well programmed recursive process )

src/main/java/duke/parser/ParserManager.java
```diff
@@ -0,0 +1,81 @@
...
+        String[] commandDescription = fullCommand.trim().split("\\s+", 2);
+        commandDescription[0] = commandDescription[0].toLowerCase();
+        String taskName = commandDescription[0];
+        try {
+            // Special case since only word "list" is required
+            if(taskName.equals("list")) { return ListCommandParser.parse(fullCommand); }
+            // Checks if command is empty
+            this.checkCommandEmpty(commandDescription);
+
+            switch (taskName) {
```
ang-zeyu: Since commands are limited to a fixed set of values, maybe it would be a good idea to use an enum?

I think it also offers a trivial runtime benefit as you only call CommandsEnum.valueOf(taskName) once ( and then catch IllegalArgumentException if invalid), and compare for enum equality. 
( In C / C++ / C# enums are fundamentally implemented with integers, _I believe_ it is the same for Java but I cant find a good source on it thus far )

(instead of comparing string equality till the correct string is reached)

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,68 @@
+package duke.task;
+
+import duke.task.ITask;
+
+public class Task implements ITask {
```
ang-zeyu: Since only one task implements ITask, and Task holds some common implementations to all Tasks (markDone, getName, ...) you could consider making Task an abstract class instead.

ang-zeyu (commented):
👍 Good implementations and clean code overall! 👍

src/main/java/duke/task/Task.java
Ellieyee: Thank you for the detailed and insightful comments, definitely learnt a lot :)! Greatly appreciate the time you took to look through my code.

src/main/java/DataStorage.java
Ellieyee: Thanks for the comment! I've taken reference to your code at https://github.com/ang-zeyu/duke/blob/master/src/main/java/duke/storage/Storage.java (in the setFilePath function). I added credit for your code and the link to your github, do let me know if you are okay with it.

ang-zeyu: Yup no problem! 👍

src/main/java/duke/command/Commands.java
```diff
@@ -0,0 +1,5 @@
+package duke.command;
+
+public enum Commands {
```
JohnNzj, line 3: Good use of enum here since the commands are limited and fixed.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,52 @@
...
+    }
```
JohnNzj, line 33: Similar to what zeyu said, adding storeTaskList(taskList) here might help!

src/main/java/duke/parser/ParserManager.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param fullCommand - command given by user input
+     * @return Optional containing either valid command or null (when exception is thrown)
+     */
+    public Optional<Command> parseCommand(TaskList taskList, String fullCommand) {
+        this.taskList = taskList;
+        try {
+            String[] commandDescription = fullCommand.trim().split("\\s+", 2);
+            Commands commandType = checkValidCommand(commandDescription);
+            switch (commandType) {
+            case LIST:
```
JohnNzj: good compatibility of enum with switch case.
looks neat and readable

JohnNzj (commented):
Overall, neat codes with good use of abstraction and well design of classes.
Compliant with coding standards and java doc comments except for a few classes without header comments.

src/main/java/duke/parser/ParserManager.java
```diff
@@ -0,0 +1,54 @@
+package duke.parser;
+
+
+import duke.command.Command;
+import duke.command.Commands;
+import duke.command.UnknownCommandException;
+import duke.task.TaskList;
+
+public class ParserManager {
+
```
tohcejasmine, line 10: Nice abstraction to have a main class for parsing that allocates the user input to more specific classes for parsing the different commands 😄

tohcejasmine (commented):
Really good use of abstraction with short and readable methods 👍

PR #50, opened 2019-08-23
[Lew Kai Ming] Duke Increments

KxxMxxx:


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.text.ParseException;
+import java.util.Scanner;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * A Parser class parses all the user input.
+ */
+public class Parser {
```
chanjunren, line 22: Hi bro, 
it might be better if you make use of a command class rather than using the parser to call the other classes :D But your code is very clean and easy to read

src/main/java/Duke.java
```diff
@@ -1,10 +1,67 @@
...
+    public void run() {
+        boolean isExit = false;
+        Scanner sc = new Scanner(System.in);
+
+        System.out.println(ui.showWelcome());
+
+        while(!isExit) {
+            String input = sc.nextLine();
+            if(input.equals("bye")) {
+                System.out.println(parser.parse(input));
```
xinmingzh: Why not consolidate duplicate code?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                    break;
+                case "E":
+                    Event event = new Event(line[2], Parser.parseDate(formatter, line[3]));
+                    if (line[1].equals("1")) {
+                        event.markDone();
+                        tasks.add(event);
+                    } else {
+                        tasks.add(event);
+                    }
```
xinmingzh, line 77: Why not consolidate all the tasks.add(events) ie. call the method once after the switch?

src/main/java/Duke.java
KxxMxxx: By that do you mean to abstract out the System.out.println()? Could you suggest how I can go about doing that? Thanks!

src/main/java/duke/storage/Storage.java
KxxMxxx: That's a good suggestion! I could use polymorphism and use a Task reference to refer to the object instances and add it after the switch.

xinmingzh (commented):
There seems to be some duplicate code which I prefer to consolidate. Is there any considerations when you did it that way? Otherwise, everything seems clean and organised.

src/main/java/Duke.java
xinmingzh: I am referring to line 47 and 50 can be simplified to one line above the if else clause.

KxxMxxx:
I used to do the printing in the UI class, however, to compromise for the addition of JavaFx, I had to return Strings instead. That's why I had to put the System.out.println() in the Duke class instead.

src/main/java/Duke.java
KxxMxxx: Oh ok I see what you meant, thanks!

chanjunren (commented):
(Sorry I forgot to press submit review)

PR #51, opened 2019-08-23
[Sahil S/O Sanjeev Gathani] Duke Increments

Sahilgat:


yeodonghan:
- I appreciate how you used LocalDateTime object to store your dates. It is simple and clear.
- Your exceptions are well handled. 
- Some of the longer lines of your codes can be addressed using the coding standards.

kxyoke:
- I like how you packaged all your classes. May I ask how you decide to which classes to package into run?
- I like how you made Ui methods static, which makes it easy to access. (I made them instance methods, following the example provided in the task description.) That said, may I ask would writing static methods or instance methods be better?
- You mainly tested Parser class on JUnit. Perhaps you could consider testing some of the other classes as well (as practice to get used to for future).
- Code is compliant with coding standards
- I also like how you have a command for each kind of task creation commands, which makes it clear what would happen when add commands are called

PR #52, opened 2019-08-23
[Murugesan Karthika] Duke increments

Kzrthikz:


lumwb (commented):
Great that you implemented gradle, checkstyle. Perhaps you could write a few more unit tests. Good job!

olivercheok20:
Good job overall! Your code is understandable and makes sense on first glance. I'm a bit confused why the merges are mostly from the 'gradle' branch instead of the respective Level- or A-branches though. Keep it up man :)

PR #53, opened 2019-08-23
[TeoShyanJie] Duke Increments

TeoShyanJie:


src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,51 @@
...
+import duke.database.Storage;
+
+/**
+ * This Command class is the abstract class with execute method
+ * and isExit() method.
+ *
+ * @author TeoShyanJie
+ *
+ */
+public abstract class Command {
```
Auxinnn, line 14: Nice use of abstraction class.

Auxinnn (commented):
I think they are really nice codes with well-organized structure.

src/main/java/duke/command/Command.java
TeoShyanJie: Thank you

src/main/java/duke/database/Storage.java
```diff
@@ -0,0 +1,215 @@
...
+     * @param ui Ui of Duke Constructor.
+     */
+    public Storage(Ui ui) {
+        this.ui = ui;
+    }
+
+    /**
+     * To get the project root and path to create the txt file or check if text file is available.
+     */
+    public void initialise() {
```
zhangxuan97, line 43: I like the way you designed your Storage class, especially this method to create a new directory if one does not already exist 👍

zhangxuan97 (commented):
Your code is very well written and documented. Good job!

PR #54, opened 2019-08-23
[Liu  Xiaoyu] Duke Increments

xiaoyu-nus:


xiaoyu-nus:
Duke Increment

src/main/java/TimeManager.java
```diff
@@ -0,0 +1,83 @@
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Contains utility functions for managing LocalDataTime object.
```
xiaoyu-nus: This class is somehow confusing and hard to understand.
Here, "getTime" method actually allows user to input incomplete date and time information and then auto-complete using current date (if date/month/year is missing) or 1200am (if time is missing).
I will add comments and modification to make this more readable in later increments.

src/main/java/IllegalTimeFormatException.java
```diff
@@ -0,0 +1,14 @@
...
+
+    /**
+     * Constructs the IllegalTimeFormatException object.
+     *
+     * @param msg Message of the exception.
+     */
+    public IllegalTimeFormatException(String msg) {
+        super(msg);
+    }
+}
```
francislow: Nice, i think it is a good idea to specify more exception variety like u did, it really helps readability :)

src/main/java/Storage.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+            this.tasks.createNewFile();
+            FileWriter fw = new FileWriter(tasks, true);
+            Path p = Paths.get(pathToFile);
+            for (Task tsk : lst) {
+                fw.write(tsk.toString());
+            }
+            fw.close();
+        } catch (IOException e) {
+            throw new DukeIoException("oops, something went wrong :(");
```
francislow: Wow, you did a really good job catching all sorts of exceptions in ur codes, I think it is good to specify when throwing duke exception that the problem occurs when saving, else it will be hard to track during runtime

francislow:
Hi Xiaoyu, 

Great job in design and abstraction even specifying additional classes such as sheet class. Ur codes are very readable. I think u did an excellent work :)

Francis

src/main/java/Storage.java
xiaoyu-nus: Thanks for your feedback! Jiayouzzzzz

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,43 @@
...
+import duke.sheet.Sheet;
+import duke.storage.Storage;
+import duke.ui.Ui;
+
+public class Duke {
+
+    private Storage storage;
+    private Sheet sheet;
+    private Ui ui;
+    private MainWindow mw;
```
ShuTingY, line 26: Naming can be improve!

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,30 @@
+import duke.command.*;
```
ShuTingY: I think need to avoid * despite you import most of the class in the package

ShuTingY (commented):
You have done good job! Keep the good work

PR #57, opened 2019-08-23
[Hoon Chee Ping] Duke Increments

hooncp:


ryanYtan:
## Code

### `duke.task`
Good inheritance done on the classes here, encapsulation also done by `TaskList`

### `duke.command`
Instead of an `abstract` method `boolean isExit()` that has to be implemented by all subclasses of `Command`, you may want to implement `boolean isExit()` as non-abstract in the `Command` base class that returns `false`, and then have `ByeCommand` override it to return `true` instead.

### `Duke.java`
* In `Duke.java`, you can combine the `catch` statements together using the `|` operator:  

   ```
   try {
       // ...
   } catch (Exception1 | Exception2 | Exception3 e) {
       // ...
   }
   ```

### `Storage.java`
Not sure if `rewriteData()` will write to the file if it doesn't exist, I remember having some trouble with that on my end.

### `Ui.java`
My approach to `Ui` was to have a general `public void print(String... args)` function so any method with access to `Ui` could print whatever they wanted. This makes it so that you don't have to change the `Ui` class if one of your `Command` subclasses needs to print a different thing (also because I had some additional formatting provided by the `Ui` class), or if you need to implement another `Command` subclass.

Not sure if it's any better than your method (since my method still requires a reference to a `Ui` instance *anyway*, and you'll be hardcoding the strings anyway). No comments otherwise.

### `Parser.java`
* Using `switch...case` may look cleaner rather than continuous `if else`
* It might not be recommended to import using `duke.command.*` (even though you actually have to import everything...)

## Style
Style guide followed, JavaDocs done, etc. I spotted one very tiny mistake where in `Storage.java`, there is no space between `IOException` and opening `{` on line 31, but that's about it.

Yen-Peng:
### Code Review
**Duke.java**

- Good packaging into command/task.

**Command.java**

- Good abstraction of command class and inheritance for child classes.

**Todo.java**

- May be better to insert a space between `Task` and `{` in line 8.

**Code**

- Good documentation of classes and methods.

hooncp:
> you may want to implement boolean isExit() as non-abstract in the Command base class that returns false, and then have ByeCommand override it to return true instead.

 Refactored code for command abstract class.

>combine the catch statements together using the | operator:

 Completed.

>Parser.java
It might not be recommended to import using duke.command.* (even though you actually have to import everything...)

 Updated.

PR #59, opened 2019-08-23
[Chong Guang Jun] Duke Increments

tcgj:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+            taskList = storage.load();
+
+            ui.sendMessage("Hello! I'm Duke",
+                    "What can I do for you?");
+            int returnCode = CODE_CONTINUE;
+            while (returnCode != CODE_EXIT) {
+                returnCode = executeMainFlow();
+            }
+            storage.save(taskList);
+        } catch (IOException e) {
```
honhaochen: Would it be better if you catch DukeException instead since this is a Duke App?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,106 @@
...
+    private TaskList taskList;
+
+    /**
+     * Creates a new <code>Duke</code> instance. The specified file path is used in
+     * loading and saving the <code>TaskList</code> for this <code>Duke</code>.
+     *
+     * @param filePath Specifies the path where the <code>TaskList</code> is loaded from.
+     */
+    public Duke(String filePath) {
+        this(new SimpleCliUi(), new Storage(filePath));
```
honhaochen: Would it be better if you just assign <new UI> and <new Storage> in this instance of class Duke? That way you don't need to have two constructors, which might cause some confusion

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Creates a new storage with the specified file path and character encoding.
+     *
+     * @param path the path of the file for reading and writing.
+     * @param charset the character encoding to be used.
+     */
+    public Storage(String path, Charset charset) {
+        filePath = Path.of(path).normalize();
```
honhaochen: Could it be just fine with only one constructor?
Pardon me because I personally think with just one will work perfectly fine

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+        this.charset = charset;
+    }
+
+    /**
+     * Reads the task list from file. If the file does not exist, an empty task list is returned.
+     *
+     * @return the task list read from file, or a new task list if the file does not exist.
+     * @throws IOException If the file input cannot be read.
+     */
+    public TaskList load() throws IOException {
```
honhaochen: would it be better to throw DukeException instead?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+                            break;
+                        case "E":
+                            Date datetime = Parser.parseDate(data[3]);
+                            list.add(new EventTask(data[2], datetime, isDone));
+                            break;
+                        default:
+                            break;
+                        }
+                    } catch (DukeParserException ignored) {
+                        // If date cannot be parsed, too bad, we shall ignore that task.
```
honhaochen: Hmmm I guess if you can store the task, loading it back should be fine

src/main/java/duke/command/CommandBuilder.java
```diff
@@ -0,0 +1,34 @@
+package duke.command;
```
honhaochen: This file is missing JavaDocs, did you forget about it?

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,94 @@
+package duke.command;
```
honhaochen: Parser.java seems to be not part of the commands, might be better if you put into another directory?

honhaochen (commented):
Comments on Command classes:
- Nice use of abstract classes and inheritance features, it really amazed me in a way, including the CommandBuilder.

Exception Handling:
- I'll suggest that the Duke.java should only deal with Duke exceptions, it should be abstracted away from any java's exceptions (IO Exception for example). One way to deal with it would be catching the exception in the class and throw another Duke-defined exception. It would be useful to deal with multiple types of Exception and would be easier to maintain the code in the future.

Overall:
- Great use of constant declarations to deal with repeated values.
- Great coding conventions, just some JavaDocs are missing.

For details, please look into the specific comments in your codes.
Anyway, great job completing the milestones :)

src/main/java/duke/Duke.java
tcgj: In this case I suppose the intention was so that someone creating a new `Duke` instance could choose to select which `filePath` they wanted to use, if they wanted to specify something rather than the default. Or even more specifically, they could supply the Ui and the Storage themselves.

src/main/java/duke/command/CommandBuilder.java
tcgj: Ah, there are no public methods here, so I didn't see a need to add JavaDoc comments here.

src/main/java/duke/storage/Storage.java
tcgj: In this case I'm just handling a case where an implementer might want to use a different encoding than UTF-8 haha

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,211 @@
...
+public class Duke extends Application {
+    /**
+     * Specifies the date format used by <code>Duke</code> for receiving date input.
+     */
+    public static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("ddMMyyyy HHmm");
+
+    /**
+     * Specifies that <code>Duke</code> should continue to read the next command.
+     */
+    public static final int CODE_CONTINUE = 0;
```
aang114: good to see you are following naming covention!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Represents a task list chatbot named Duke.
+ *
+ * @author Terence Chong Guang Jun
+ */
+public class Duke extends Application {
+    /**
+     * Specifies the date format used by <code>Duke</code> for receiving date input.
+     */
+    public static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("ddMMyyyy HHmm");
```
aang114: maybe it would be better if you made a seperate class to handle converting strings to dates

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,211 @@
...
+     *           Reads and prints replies to the user.
+     * @param storage the storage for this <code>Duke</code>.
+     *                Reads and writes the <code>TaskList</code> to file.
+     */
+    public Duke(Ui ui, Storage storage) {
+        this.ui = ui;
+        this.storage = storage;
+    }
+
+    private int executeMainFlow() throws IOException {
```
aang114: good structuring!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,211 @@
...
+        Label userText = new Label(userInput.getText());
+        Label dukeText = new Label(this.getResponse(userInput.getText()));
+        dialogContainer.getChildren().addAll(
+                new DialogBox(userText, new ImageView(user)),
+                new DialogBox(dukeText, new ImageView(duke))
+        );
+        userInput.clear();
+    }
+
+    private String getResponse(String input) {
```
aang114: it seems you have not made the gui user interactive yet?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,211 @@
...
+        return "Duke heard: " + input;
+    }
+
+    public static void main(String[] args) {
+        String path = "data/duke.txt";
```
aang114: maybe you can make this a final variable?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Represents a command entered by the user. Abstract base class to be inherited by
+ * more specific command types.
+ *
+ * @author Terence Chong Guang Jun
+ */
+public abstract class Command {
+    protected List<String> arguments;
+
+    Command() {
```
aang114: this seems redundant

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Date;
+import java.util.HashMap;
+
+/**
+ * Represents a parser for string inputs.
+ *
+ * @author Terence Chong Guang Jun
+ */
+public class Parser {
+    private static HashMap<String, CommandBuilder> commandMap;
```
aang114: very cool that you are using a HashMap to parse!

src/main/java/duke/command/TodoCommand.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+
+/**
+ * Represents a todo command entered by the user.
+ *
+ * @author Terence Chong Guang Jun
+ */
+public class TodoCommand extends AddTaskCommand {
+    TodoCommand() {
```
aang114: this seems redundant?

src/main/java/duke/ui/SimpleCliUi.java
```diff
@@ -0,0 +1,31 @@
+package duke.ui;
+
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Represents a simple command line user interface.
+ *
+ * @author Terence Chong Guang Jun
+ */
+public class SimpleCliUi extends Ui {
```
aang114: Wow! Very smart and unique way to represent the command line user interface

src/test/java/duke/command/ParserTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.api.Test;
+
+import duke.exception.DukeParserException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class ParserTest {
+    @Test
+    public void parseDate_validDateFormat_success() throws Exception {
```
aang114: if i am not wrong, it is not recomended to have methods with '_' (it's against coding standards i think)

aang114 (commented):
Overall your code is very OOP friendly, very easy-to-read and very well implemented - although there were a few places where the coding standard was not followed. Kudos!

src/test/java/duke/command/ParserTest.java
tcgj: Haha, perhaps you might wanna look at [https://nus-cs2103-ay1920s1.github.io/website/schedule/week3/topics.html#tools-junit-junit-basic](url)

src/main/java/duke/command/Command.java
tcgj: Yes I wish it was also. I needed a package private constructor for the prototype instances of my concrete classes inheriting this. Could do without, but that would not be very good in terms of visibility.

PR #61, opened 2019-08-23
[Evon Dong] Duke Increments

EvonDong:
Add bye


src/main/java/seedu/duke/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Writes new list of tasks to the datafile, overwriting previous tasks.
+     *
+     * @param taskList the arraylist of tasks to be written
+     * @throws IOException on wrong input
+     */
+    public void writeToFile(TaskList taskList) throws IOException {
+        FileWriter fw = new FileWriter(filePath);
```
cutieprobe, line 90: I noticed you used BufferedReader in `readFile`, maybe consider also using `BufferedWriter` here to improve IO performance?

src/main/java/seedu/duke/Task.java
```diff
@@ -0,0 +1,81 @@
...
+        return (isDone ? "[✓]" : "[✗]");
+    }
+
+    /**
+     * Sets the isDone variable to true and the doneIcon to 1.
+     * Prints the done message.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+        this.doneIcon = 1;
```
cutieprobe, line 54: Is it better to reuse the `updateDone` function here to avoid repetitive code?

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,72 @@
...
+    private Ui ui;
+
+    /**
+     * Constructor for class Duke.
+     * Loads the data if datafile exists.
+     *
+     * @param filePath the path to the file where the list of tasks is stored
+     * @throws IOException On input error
+     */
+    public Duke(String filePath) throws IOException {
```
cutieprobe, line 32: Since `DukeException` is the base class of all exceptions thrown in Duke, is it better to catch the `IOException` and only throw `DukeException`? Similarly for `ParseException`.

src/main/java/seedu/duke/Storage.java
EvonDong: Noted, thank you.

src/main/java/seedu/duke/Task.java
EvonDong: Noted, I will update it soon.

src/main/java/seedu/duke/Command.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Returns whether the program should exit.
+     *
+     * @return the exit variable
+     */
+    public boolean isExit() {
+        return exit;
+    }
+
+}
```
Kyzure, line 30: Great use of Abstract Class instead of interface as a base class for the actual Commands.

src/main/java/seedu/duke/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+        FileWriter fw = new FileWriter(filePath, true);
+        if (taskList.getSize() == 1) {
+            fw.write(taskList.getTask(taskList.getSize() - 1).writeToFile());
+        } else {
+            fw.write(System.lineSeparator() + taskList.getTask(taskList.getSize() - 1).writeToFile());
+        }
+        fw.close();
+    }
+
+}
```
Kyzure, line 117: Overall you used Buffered Reader, File Writer and File Reader to encode your code. Perhaps you may be interested in [Serializable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html) as it may much simpler to serialize the code. Otherwise good job on encoding the code.

src/main/java/seedu/duke/Task.java
Kyzure: Agreed, or another possibility would be to `markAsDone` only, but no need for the return value, if necessary that is.

src/main/java/seedu/duke/Task.java
```diff
@@ -0,0 +1,81 @@
...
+        doneIcon = 0;
+    }
+
+    /**
+     * Return the status icon depending on whether the task is marked done or not.
+     *
+     * @return the status icon
+     */
+    public String getStatusIcon() {
+        return (isDone ? "[✓]" : "[✗]");
```
Kyzure, line 45: Good to see usage of ternary operators instead of is and else in shorter codes.

src/main/java/seedu/duke/TaskList.java
```diff
@@ -0,0 +1,85 @@
...
+        return taskList.size();
+    }
+
+    /**
+     * Loops through the list of tasks to find the tasks containing the keywords.
+     *
+     * @param keyword the keyword that need to be matched
+     * @return an arraylist tasks containing the keywords
+     */
+    public ArrayList<Task> findMatching(String keyword) {
```
Kyzure, line 83: Instead of splitting the String up for each task, you could use `taskList.get(i).getDescription().contains(keyword)`. Similarly, you might want to consider lower and upper case for when searching for keyword, unless you want a strict keyword finding feature.

src/main/java/seedu/duke/Deadline.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Date;
+
+/**
+ * Represents a Deadline task.
+ * A <code>Deadline</code> object corresponds to a task with a command that starts with "deadline" and contains
+ * a description and a timestamp e.g., <code>deadline project 3/12/2022 0212</code>
+ */
+public class Deadline extends Task {
+
+    protected SimpleDateFormat dateFormat = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
Kyzure: Maybe `SimpleDateFormat` could be a constant instead of a variable, and placed in Task.java instead, as this is repeated in Event.java as well.
I.e. `protected static final DATE_FORMAT`

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Returns the indexOfByAt variable of the command.
+     *
+     * @return the indexOfByAt variable of the command
+     */
+    public int getIndexOfByAt() {
+        return this.indexOfByAt;
+    }
+
+}
```
Kyzure, line 190: Good job on parsing the inputs by the user, but maybe using Enums for the different commands may make the code slightly more consistent and understandable, thought longer.

src/main/java/seedu/duke/ListCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Prints all the tasks in the list.
+     *
+     * @param list the TaskList object that is handling the arraylist of the datafile
+     * @param ui the UserInterface object that handles the interaction with users
+     * @param storage the Storage object that stores and handles the path to datafile
+     * @return the output string
+     */
+    public String execute(TaskList list, Ui ui, Storage storage) {
+        return ui.printList(list);
```
Kyzure, line 24: In my opinion, scanning through the list should not be the job of the UI, instead, it should be here. Otherwise the logic behind the printing is correct.

src/main/java/seedu/duke/Ui.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Returns the error message.
+     *
+     * @param s the error message
+     */
+    public String printErrMsg(String s) {
+        return s + "\n";
+    }
+
+}
```
Kyzure, line 176: For certain methods (The ones with the for loops), the Ui should not handle any searching, but instead, handle the printing and scanning of the output and input. Apart from that, great work on the overall ui management.

text-ui-test/runtest.sh
```diff
@@ -0,0 +1,34 @@
...
+# compare the output to the expected output
+diff ACTUAL.TXT EXPECTED.TXT
+if [ $? -eq 0 ]
+then
+    echo "Test result: PASSED"
+    exit 0
+else
+    echo "Test result: FAILED"
+    exit 1
+fi
```
Kyzure, line 34: Good work on using Text Ui for testing the actual code, as well as JUnit for pinpointing and testing certain classes.

Kyzure (commented):
Overall good job on A-More-OOP and A-JUnit as well. Everything that is necessary to Java-Docs is being documented in well-written English. The logic and flow behind the code is wisely defined, and even extra exception classes such as `EmptyDescriptionException` is being used to make the code more understandable.

PR #63, opened 2019-08-23
[Lee Shi Jie Shawn] Duke Increments

shawnlsj97:
Completed requested features


shawnlsj97:
Completed up to Level-9

src/main/java/DateTime.java
```diff
@@ -0,0 +1,121 @@
...
+        case 11:
+            month = "November";
+            break;
+        case 12:
+            month = "December";
+            break;
+        }
+    }
+
+    public void processTime(String t) {
```
lawncegoh: The amount of effort put in here should be applauded but there is a simpler way in the one that I did, in that you can use the dateFormat class and the methods within and it makes all these date conversions quite automatic

src/main/java/DateTime.java
```diff
@@ -0,0 +1,121 @@
...
+    String day;
+    String month;
+    String year;
+    String time;
+
+    public DateTime(String dateTime) {
+        this.dateTime = dateTime;
+    }
+
+    public void processDate(String d) {
```
lawncegoh: Perhaps the 'd' character in the params of this can be replaced with something more meaningful according to the coding standards convention

src/main/java/TaskList.java
```diff
@@ -0,0 +1,30 @@
+import java.util.ArrayList;
+
+public class TaskList {
```
lawncegoh: TaskList class is succinct and sufficient

src/main/java/Ui.java
```diff
@@ -0,0 +1,185 @@
...
+        System.out.println(logo + '\n' + greetMsg);
+    }
+
+    public String getUserInput() {
+        String userInput = sc.nextLine();
+        return userInput;
+    }
+
+    public static void handleBye() {
+        System.out.println("    ____________________________________________________________\n" +
```
lawncegoh: The '___________' dividers can probably be included in a showLine() method like I have done in my code, to make everything neater and not exceed the amount of characters in a specific line

src/main/java/Ui.java
```diff
@@ -0,0 +1,185 @@
...
+        try {
+            throw new DukeException("    ____________________________________________________________\n" +
+                    "     ☹ OOPS!!! I'm sorry, but I don't know what that means :-(\n" +
+                    "    ____________________________________________________________\n");
+        } catch(DukeException ex) {
+            System.err.println(ex.getMessage());
+        }
+    }
+
+    public static void handleDone(String input, TaskList tasks, Storage storage) throws DukeException {
```
lawncegoh: I understand how all these handlers are meant to separate the commands but I feel that the UI class should solely be for user input and that a Parser class can be used to parse these user inputs and transform them into commands that will be executed, exactly like how I have done it in my code. Everything will seem less messy too

src/main/java/Duke.java
```diff
@@ -4,177 +4,188 @@
...
-                } else if (input_string[0].equals("delete")) {
-                    int task_num = Integer.parseInt(input_string[1]);
-                    int new_total_tasks = tasks.size() - 1;
-                    Task t = tasks.get(task_num - 1);
-                    tasks.remove(task_num - 1);
-                    printDelete(t);
+                userInput = sc.nextLine();
+                inputStringArr = userInput.split(" ");
+                String command = inputStringArr[0];
+                if (command.equals("bye")) {
```
lawncegoh: The several lines of if and else if can perhaps be swapped for switch, case

src/main/java/DateTime.java
shawnlsj97: yeah i agree should've search for a class that would do this automatically

src/main/java/Ui.java
shawnlsj97: oh yeah i alr did that but seems like my PR did not update :/

src/main/java/DateTime.java
shawnlsj97: true esp for team project

src/main/java/Ui.java
shawnlsj97: ah ok cuz i read the task specification which said it deals with interactions with the user so i thought this was the class that would handle printing as well. will re-implement this. thanks!

src/main/java/Duke.java
```diff
@@ -1,10 +1,80 @@
...
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (FileNotFoundException ex) {
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * This method contains the main logic of the program. Program is executed until exit command is issued
```
Seris370: I thought the first sentence of documentation for public methods and classes should always start with a verb.

src/main/java/duke/date/DateTime.java
```diff
@@ -0,0 +1,98 @@
...
+            simpleDateFormat.setLenient(false);
+            Date date = simpleDateFormat.parse(dateTime);
+            day = new SimpleDateFormat("d").format(date);
+            month = new SimpleDateFormat("MMMM").format(date);
+            year = new SimpleDateFormat("yyyy").format(date);
+            time = new SimpleDateFormat("h:mma").format(date).toLowerCase();
+            processDay(day); // adds st, nd, rd, th suffix to day
+        } catch (ParseException ex) {
+            System.err.println(ex);
+            System.exit(1);
```
Seris370: Nice.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,56 @@
...
+    public String getType() {
+        return type;
+    }
+
+    /**
+     * This method is used to get the string representing the deadline task.
+     * @return String This returns the string representing the deadline task
+     */
+    @Override
+    public String getTypeIcon() {
```
Seris370: Good idea.

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public TaskList() {
+        tasks = new ArrayList<>();
+    }
+
+    /**
+     * Class constructor that is called when user has tasks stored from previous session.
+     * @param storage Arraylist of tasks retrieved from storage on hard disk.
+     */
+    public TaskList(ArrayList<Task> storage) {
```
Seris370: Can use a different name as there is a storage class.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,254 @@
...
+     * Print message that a task has been marked as completed.
+     * @param input User input
+     * @param tasks List of tasks
+     * @param storage Update tasks stored persistently
+     * @throws DukeException
+     */
+    public static void handleDone(String input, TaskList tasks, Storage storage) throws DukeException {
+        String[] inputStringArr = input.split(" ");
+        if (inputStringArr.length > 1) {
+            int taskNum = Integer.parseInt(inputStringArr[1]);
```
Seris370: Will NumberFormatException be handled here?

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,254 @@
...
+     */
+    public static void handleTodo(String input, TaskList tasks, Storage storage) throws DukeException {
+        String[] inputStringArr = input.split(" ");
+        if (inputStringArr.length > 1) {
+            String taskName = ((input.split(" ", 2))[1]).strip();
+            Todo t = new Todo(taskName);
+            tasks.add(t);
+            System.out.println("     Got it. I've added this task: \n" +
+                    "       [T]" + "[ ]" + ' ' + t+ '\n' +
+                    "     Now you have " + tasks.size() + " tasks in the list.");
```
Seris370: "Tasks" should be in singular form if there  is only 1 task.

Seris370 (commented):
Well-written code with super detailed documentation. Have learnt a lot from it.

PR #66, opened 2019-08-23
[Bahety Arjavi] Duke Increments

arjavibahety:


src/main/java/Duke.java
```diff
@@ -42,22 +40,22 @@ public void run() {
...
-                Command c = (new Parser()).parse(echo);
+                Command cmd = (new Parser()).parse(echo);
                 ui.showLine();
                 try {
                     if (echo.equals("list")) {
                         ui.showTasksInList(tasks);
-                    } else if (c.getCommandType().equals("find")) {
-                      ui.showFoundItems(c.getInstruction(), tasks);
-                    } else if (c.getCommandType().equals("done")) {
+                    } else if (cmd.getCommandType().equals("find")) {
```
heze8: Hi looks a bit long here, maybe you could consider using a switch case format?

heze8:
Hello Arjavi!

Good job overall, your code is neat and follows a good OOP style. Also no serious problems with code style. I've listed a few comments on your code for your review. Also some major things I've noticed is that you have not packaged your classes yet. And your compiled classes are in the same folder as your .java files, maybe consider moving them to another folder to keep it tidy? Otherwise well done so far!

src/main/java/Duke.java
```diff
@@ -112,12 +110,12 @@ public void run() {
...
-                        } else if (c.getCommandType().equals("todo")) {
+                        } else if (cmd.getCommandType().equals("todo")) {
                             try {
-                                if (c.getInstruction().equals("") || c.getInstruction().equals("todo")) {
+                                if (cmd.getInstruction().equals("") || cmd.getInstruction().equals("todo")) {
                                     throw new DukeException("☹ OOPS!!! The description of todo cannot be empty");
                                 } else {
-                                    tasks.addItemToList(new Todo(c.getInstruction()));
+                                    tasks.addItemToList(new Todo(cmd.getInstruction()));
                                     try {
```
heze8: Perhaps you could reduce the lines and improve clarity by removing the multiple try catch clauses into one larger try bracket in the method? Also, another approach is to maybe have all methods throw exceptions and have one class handle all exceptions?

src/main/java/Storage.java
```diff
@@ -25,39 +25,45 @@ public Storage(String filePath) {
...
         ArrayList<Task> list = new ArrayList<>();
-        Scanner s = new Scanner(f);
-        while(s.hasNextLine()) {
-            String[] task = s.nextLine().replaceAll(", ", ",").split(",");
-            Task tsk;
+        Scanner sc = new Scanner(file);
+        while(sc.hasNextLine()) {
+            String[] task = sc.nextLine().replaceAll(", ", ",").split(",");
+
             if(task[0].equals("T")) {
```
heze8: Spaces after if and while to follow the codestyle

src/main/java/Duke.java
```diff
@@ -1,10 +1,194 @@
...
+                                    storage.update(tasks.getList());
+                                } catch (Exception exp) {
+                                    ui.showErrorMsg(exp.getMessage());
+                                }
+                            }
+                        } catch (DukeException exp) {
+                            ui.showErrorMsg(exp.getMessage());
+                        }
+
+                    } else if (cmd.getCommandType().equals("delete")) {
```
ybchen97: Maybe instead of manually checking for the command type you could further abstract these commands out into their own respective classes? So each command type would have their own execute() function, and you would just call command.execute() in your run() loop after parsing the user input. In this sense, the execution logic would all be abstracted into their own subclass, so it won't have to appear in this one run() funciton. This would also reduce the need for so many try catch clauses as all execute methods would return their own kinds of DukeExceptions, and only one try catch clause is needed to catch all the varying exceptions. Just something to consider!

src/main/java/Duke.java
```diff
@@ -1,10 +1,194 @@
...
+                            }
+                        } else if (cmd.getCommandType().equals("event") || cmd.getInstruction().equals("event")) {
+                            try {
+                                if (cmd.getInstruction().equals("")) {
+                                    throw new DukeException("☹ OOPS!!! The description of event cannot be empty");
+                                } else if (!echo.contains("/at")) {
+                                    throw new DukeException("☹ OOPS!!! The event must be at by a certain date");
+                                } else {
+                                    try {
+                                        SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
ybchen97: Maybe you can consider shifting more logic related to creating and adding tasks to addItemToList() method of TaskList, this can help to shorten the length of your run function.

ybchen97 (commented):
Hi Arjavi! Looks pretty good so far! Just a few things to consider, such as have more abstraction and splitting up your logic into further subclasses so your run section can be a little cleaner. Apart from that, great job!

src/main/java/Command.java
```diff
@@ -0,0 +1,41 @@
+public class Command {
+    private String commandType;
```
zhangj1an: Since you set commandType to private, I'm guessing maybe you could also set it to final. But this is not a violation haha, it's already good.

src/main/java/Parser.java
```diff
@@ -0,0 +1,27 @@
...
+    public Command parse(String command) {
+        String commandType = command.split(" ")[0];
+        String instruction = "";
+        String date = "";
+
+        if (commandType.equals("todo") || commandType.equals("delete") || commandType.equals("done")
+                || commandType.equals("find")) {
+            instruction = command.substring(command.indexOf(" ") + 1);
+        } else if (commandType.equals("deadline") || commandType.equals("event")) {
+            if (command.contains("/")) {
```
zhangj1an, line 40: This contains() method wouldn't work for commands like `event task_description /to time_date` whereby the "/at" condition was not met. Although lengthy, it would be clearer to split "event" and "deadline" into 2 different sections. 
Import java.util.regex package and use codes like `regex.matches(.\/at*)` might help. But this version is already good enough! 💃I like the clear conditions.

src/main/java/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+
+            if (task[0].equals("T")) {
+                Task tsk = new Todo(task[2]);
+                if (task[1].equals("1")) {
+                    tsk.markAsDone();
+                }
+                list.add(tsk);
+            } else if (task[0].equals("D")) {
+                try {
+                    SimpleDateFormat formatter = new SimpleDateFormat("EEEEE MMMMM dd HH:mm:ss z yyyy");
```
zhangj1an, line 42: Good use of APIs. I didn't know about this haha

src/main/java/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+                    Task tsk = new Deadline(task[2], formatter.parse(task[3]));
+                    if (task[1].equals("1")) {
+                        tsk.markAsDone();
+                    }
+                    list.add(tsk);
+                } catch (java.text.ParseException exp) {
+                    exp.printStackTrace();
+                    break;
+                }
+            } else {
```
zhangj1an, line 52: A extended if-elseif-else outer loop with a try-catch inner loop here. Perhaps can extract the inner loop into another function to enhance readability. But code is already clear with reference to JavaDoc written.

src/main/java/Task.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Marks the task as done.
+     */
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    /**
+     * Stringified representation of the task.
```
zhangj1an, line 47: `Stringifies` you mean. Other than that your JavaDoc is impeccable. 👍💯

src/main/java/Storage.java
```diff
@@ -0,0 +1,131 @@
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Scanner;
+
```
zhangj1an, line 8: "Every class should be part of some package", says Java coding standard. It is optional but I feel you can try it out if you feel like it. (Because tbh I think you has already achieved all other requirements of Java coding standard.)

zhangj1an (commented):
Hi Bahety, you have very well written code. It is very clear and is easy to understand for peer workers. I tried but cannot find any fault in your coding standard. Good work! Keep it up.

PR #67, opened 2019-08-23
[Roby Tanama] Duke Increments

tanamaroby:
Hello! Welcome to my Pull Request for the Duke project.
This request contains the code that I have made for my Individual Project in the module CS2103T in NUS year 2019/2020. Feel free to leave constructive comments and suggestions about my code. Should you see any part of my code that could use some improvements.

Happy reviewing!


src/main/java/Duke.java
```diff
@@ -1,10 +1,36 @@
...
+        }
+
+        this.ui = new UI(taskList);
+        this.parser = new Parser(taskList, storage, ui);
+    }
+
+    public void run() throws Exception {
+        ui.printIntro();
+        storage.loadTasks();
+        storage.loadTasks();
```
JunHongT: Is there a reason to repeat the same process a second time?

src/main/java/Duke.java
```diff
@@ -1,10 +1,36 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+    private Storage storage;
+    private Parser parser;
+    private UI ui;
+    private TaskList taskList;
+
+    public Duke(String filepath) throws Exception {
```
JunHongT: Will it be a better idea to throws a specific exception instead of a general Exception?

src/main/java/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+    private UI ui;
+
+    public Parser(TaskList tasks, Storage storage, UI ui) {
+        this.br = new BufferedReader(new InputStreamReader(System.in));
+        this.listOfTasks = tasks;
+        this.storage = storage;
+        this.ui = ui;
+    }
+
+    public void parse() throws Exception, DukeException {
```
JunHongT: Might it be possible to segregate your method to smaller methods? The method will works fine without it but i believe that overall it will looks neater and more concise.

JunHongT:
In general, your code flow is smooth and allows reader to gain a basic idea of what is going on with your code. However, in the event that someone who is not taking part in our module, it might be a great idea to make use of comment/javadoc for them to have an easier time understanding the code.

It is a great idea to segregate all the different message into your Ui class. Personally, i will recommend the use of StringBuilder or other String formatter rather than System.out.println() everything in the Ui class. I do not think that this will be a problem for Duke but you may consider trying out StringBuilder in the event of future project?

src/main/java/Duke.java
tanamaroby: Dear Jun Hong T,

I believe this was a careless mistake, which I will rectify immediately.
Thank you Jun Hong T, very cool!

Best regards
Roby Tanama

src/main/java/Duke.java
tanamaroby: Dear Jun Hong T, 

I agree that throwing a specific exception would be better. However, I believe that it is imperative for me to cover every base should there be any exception that passed my notice. Though this could just be unnecessary paranoia. Despite that, thank you for pointing this out to me! My sincerest gratitude for your review Jun Hong T, very epic!

Best regards
Roby Tanama

src/main/java/Parser.java
tanamaroby: Dear Jun Hong T,

There is definitely a better way of handing the method, I will do what I can to segregate this specific method and make it more readable. Great point to make Jun Hong T, amazing comment!

Best regards
Roby Tanama

robytanama:
> In general, your code flow is smooth and allows reader to gain a basic idea of what is going on with your code. However, in the event that someone who is not taking part in our module, it might be a great idea to make use of comment/javadoc for them to have an easier time understanding the code.
> 
> It is a great idea to segregate all the different message into your Ui class. Personally, i will recommend the use of StringBuilder or other String formatter rather than System.out.println() everything in the Ui class. I do not think that this will be a problem for Duke but you may consider trying out StringBuilder in the event of future project?

tanamaroby:
> In general, your code flow is smooth and allows reader to gain a basic idea of what is going on with your code. However, in the event that someone who is not taking part in our module, it might be a great idea to make use of comment/javadoc for them to have an easier time understanding the code.
> 
> It is a great idea to segregate all the different message into your Ui class. Personally, i will recommend the use of StringBuilder or other String formatter rather than System.out.println() everything in the Ui class. I do not think that this will be a problem for Duke but you may consider trying out StringBuilder in the event of future project?

Dear Jun Hong T,

I am delighted to receive your review concerning my code for this module CS2103T, and will be more than happy to rectify the issues that you have pointed out. 

I thoroughly enjoyed reading the review you have given me and will take those as a learning experience to further improve my code in the future. Many of your suggestions have been noted and will be implemented if possible, in future commits. 

Should you have any more questions or reviews regarding my code, please feel free to leave them here. I am looking forward to future reviews and suggestions. Thank you very much for your time and best of luck in your future endeavors.

Best regards
Roby Tanama

ChengGeng97:
Overall, I think your code is clean and readable. It's easy to see what each block of code is meant to do. That said, I think more comments could be added to guide a collaborator along. Even if the code is readable, it's always good to know what the code does without having to read it.

With regards to readability, one small change that could be quite beneficial to a collaborator is spelling out bufferedReader and fileReader as variable names rather than using br and fr. Being that they are class-wide properties, a collaborator might be slightly inconvenienced to have to scroll up and down to find out what they are.

This one is a personal preference, but I spotted some instances where classes that don't strictly have to store state hold on to state nonetheless. For example, Parser and Duke hold the same references to TaskList, Storage and UI. The reference to the TaskList is also shared with Parser. This way, the coder (or I guess in this case, you) will be forced to pass the TaskList/Storage/UI when you call the method, which would make it easier for a reader to understand the dependencies of a particular method.

If you get rid of state, you can also make methods like Parser's parse and UI's print methods static, allowing you to call them from a static instance. Then you could reduce the number of references you hold.

Overall Rating: Kaguya-sama is a great manga with likable and relatable characters, and keeps its formula fresh as we see the characters grow and change. 11/10.

tanamaroby:
> Overall, I think your code is clean and readable. It's easy to see what each block of code is meant to do. That said, I think more comments could be added to guide a collaborator along. Even if the code is readable, it's always good to know what the code does without having to read it.
> 
> With regards to readability, one small change that could be quite beneficial to a collaborator is spelling out bufferedReader and fileReader as variable names rather than using br and fr. Being that they are class-wide properties, a collaborator might be slightly inconvenienced to have to scroll up and down to find out what they are.
> 
> This one is a personal preference, but I spotted some instances where classes that don't strictly have to store state hold on to state nonetheless. For example, Parser and Duke hold the same references to TaskList, Storage and UI. The reference to the TaskList is also shared with Parser. This way, the coder (or I guess in this case, you) will be forced to pass the TaskList/Storage/UI when you call the method, which would make it easier for a reader to understand the dependencies of a particular method.
> 
> If you get rid of state, you can also make methods like Parser's parse and UI's print methods static, allowing you to call them from a static instance. Then you could reduce the number of references you hold.
> 
> Overall Rating: Kaguya-sama is a great manga with likable and relatable characters, and keeps its formula fresh as we see the characters grow and change. 11/10.

Dear Mr Cheng Geng 97,

I very much appreciate the comment you made on my code. Some of the suggestions that you have given me are actually very useful and would greatly improve both efficiency and readability of my code. Due to the massive amount of changes to be done, it will be a while before your suggestions come to fruition in my code. However, I will not falter at such monumental task and will continue to further enhance my work. As Kaguya-sama used to say : "All I can do is behave the way I know how. I don't know any other ways. This is the only way I can do it."

Overall your review has a little something for everyone.

Best regards
Roby Tanama

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
...
+     */
+    private ScrollPane scrollPane;
+
+    /**
+     * The dialog window of the application.
+     */
+    private VBox dialogContainer;
+
+    /**
+     * The box for the user input;
```
TeoShyanJie: Slight error for the comment. I think should be a "."

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,104 @@
...
+        String tempInfo = "";
+        for (int i = marker + 1; i < inputMessage.length; i++) {
+            if (i == inputMessage.length - 1) {
+                tempInfo += inputMessage[i];
+            } else {
+                tempInfo += inputMessage[i];
+                tempInfo += " ";
+            }
+        }
+        String[] dateAndTime = tempInfo.split(" ");
```
TeoShyanJie, line 58: The function is very long perhaps it is better abstracted.

TeoShyanJie (commented):
Overall has a good Javadoc and has a very nice OOP style.

PR #71, opened 2019-08-23
[Sim Jun Yuen, Darren] Duke Increments

Kalsyc:


src/main/java/Duke.java
```diff
@@ -1,10 +1,196 @@
-public class Duke {
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+import javafx.application.Application;
```
gohsnn: I like how your import statements are grouped according to their functionality which makes it easier to read!

src/main/java/slave/command/DoneCommand.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Executes by marking a particular task as done and prints to the user.
+     *
+     * @param tasks List containing current tasks.
+     * @param ui User interface.
+     * @throws DukeException For tasks that cannot be found or have already been done.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui) throws DukeException {
```
gohsnn: Your error catching here is specific and easy to identify. I feel that this would help me improve my code in the future.

src/main/java/slave/elements/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+public class Parser {
+
+    /**
+     * Static method which parses input by user and returns the corresponding command object.
+     *
+     * @param fullCommand Input by user.
+     * @return Corresponding command based on user input.
+     * @throws DukeException For invalid input.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
```
gohsnn: Following the logic of your code here, would it be better to remove your enum class entirely?

src/main/java/slave/elements/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+            return new ClearCommand();
+        case "find":
+            return new FindCommand(fullCommand.substring(5));
+        case "done":
+            checkValidity("done", fullCommand, tokens);
+            return new DoneCommand(Integer.parseInt(tokens[1]));
+        case "delete":
+            checkValidity("delete", fullCommand, tokens);
+            return new DeleteCommand(Integer.parseInt(tokens[1]));
+        case "todo":
```
gohsnn: This method may seem a bit lengthy, would it be better to abstract the logic of adding the different tasks into an AddCommand class, which would then have your AddToDoCommand, AddDeadlineCommand and AddEventCommand extend from? It can potentially be helpful since both deadlines and events require formatting of the date.

src/main/java/slave/elements/Storage.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Encodes task and add it to .txt file.
+     *
+     * @param task Task to store in .txt file.
+     * @throws DukeException Throws if input went wrong in writing file.
+     */
+
+    void addTask(Task task) throws DukeException {
```
gohsnn: I like how your methods for formatting the tasks in and out of the text file is so clean!

src/main/java/DialogBox.java
```diff
@@ -0,0 +1,46 @@
...
+import javafx.collections.ObservableList;
+
+import javafx.geometry.Pos;
+
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+public class DialogBox extends HBox {
```
le0tan, line 22: Actually popup windows can also be written in FXML and be loaded using an FXML loader.

src/main/java/Duke.java
```diff
@@ -23,22 +41,137 @@
...
     }
 
+    /**
+     * Iteration 1:
+     * Creates a label with the specified text and adds it to the dialog container.
+     * @param text String containing text to add
+     * @return a label with the specified text that has word wrap enabled.
+     */
+    private Label getDialogLabel(String text) {
+        // You will need to import `javafx.scene.control.Label`.
```
le0tan: Maybe consider removing the comment from the tutorial?

src/main/java/Duke.java
```diff
@@ -23,22 +41,137 @@
...
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(userText, new ImageView(user)),
+                DialogBox.getDukeDialog(dukeText, new ImageView(duke))
+        );
+        userInput.clear();
+    }
+
+    /**
+     * You should have your own function to generate a response to user input.
+     * Replace this stub with your completed method.
```
le0tan: Same as the last comment.

PR #73, opened 2019-08-23
[Wang Xueting] Duke Increments

ShirleyWangxt:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,53 @@
...
+            tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    public static void main(String[] args) {
+
+        Duke duke = new Duke("../data/duke.txt");
```
garylyp: Hi Xueting, I thought that the file path could be passed into Storage directly in the Duke constructor (without Duke having to take in an additional argument). Perhaps, you were coming from the direction that different Duke users would have different file paths, which is a good consideration, but I felt that the storage directory isn't usually a customizable attribute since we don't need users to know where we store the temp files.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+                    task.markAsDone();
+                }
+
+                tasks.add(task);
+            }
+
+            return tasks;
+
+        } catch (FileNotFoundException ex) {
+            throw new DukeException("File is not found");
```
garylyp, line 59: Hi, instead of throwing a new exception upon catching one, I would suggest catching both exceptions separately with multiple catch blocks and then specifying the course of action when the specific exception is encountered. 

Also, it is good practice to catch your exceptions where you expect it to be thrown, rather than making the method throws the exception and handing the responsibility to higher level code to catch it. This may sometimes make it difficult for other users to find the precise location of the error when testing your code. For your case, you caught the DukeException in the Duke constructor. Fortunately, there is only one statement in the "try" block so it is still not too difficult to identify the source of the thrown exception.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,51 @@
+package duke;
```
garylyp: Small issue here but I thought it would be more apt to place the TaskList.java in your Task package where all the Task related classes are.

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class AddDeadlineCommand extends Command {
+    public AddDeadlineCommand(String[] description) {
+        super(description);
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        if (super.description.length < 2) {
+            throw new DukeException("☹ OOPS!!! The description of an event cannot be empty.");
```
garylyp, line 20: Hi, since this exception is in relation to user input, I felt it might be more apt if you catch it at the earliest possible point, which is at Parser itself. This applies to other subclass of your Task as well. Again, it is good habit to catch the exception where you throw it, even though the IDE may sometimes offer suggestions for you to add the "throws" clause in the method signature and ask you to catch them where you call the method.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public void showLoadingError() {
+        System.out.println("File cannot be loaded. A new task list is created.");
+    }
+
+    public void showError(String msg) {
+        showLine();
+        System.out.println("     " + msg);
+        //showLine();
+    }
```
garylyp: I felt that it was good that you specify the names of the errors when you throw the DukeException so that they can be printed out using the showError method. It really helps users find out what is the specific cause of the the exceptions thrown.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,34 @@
...
+import duke.command.*;
+import duke.exception.DukeException;
+
+/**
+ * Represents a parser to parse user command.
+ */
+public class Parser {
+
+    public static Command parse(String fullCommand) throws DukeException {
+        String[] description = fullCommand.trim().split(" ", 2);
```
garylyp, line 13: As a parser, I feel that this class could be further developed more to help check the format and requirements of the user input rather than pushing that role down to the Command class to handle.

garylyp (commented):
Good job implementing all the features! But do take note of your placement of try-catch blocks and when you include "throws" in your methods. :)

ReignOfComputer:
Looks good to me.

src/main/java/duke/Main.java
```diff
@@ -0,0 +1,32 @@
...
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private Duke duke = new Duke();
```
xinmingzh, line 17: Will this initialization be more suitable to be in the start method?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,85 @@
+package duke;
+
+import duke.task.*;
```
xinmingzh, line 3: Import packages individually

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+                task = new ToDo(lines[2]);
+                break;
+            case "D":
+                task = new Deadline(lines[2], lines[3]);
+                break;
+            case "E":
+                task = new Event(lines[2], lines[3]);
+                break;
+            default:
+                throw new DukeException("The task does not exist.");
```
xinmingzh, line 76: Consider using more specific Exception types that extends DukeException.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,44 @@
+package duke.task;
+
+import java.util.ArrayList;
+
+/**
+ * Represents a tasklist object which handles list update.
+ */
+public class TaskList {
+
+    static ArrayList<Task> list;
```
xinmingzh: Consider using the interface List<Task> instead of the implementation class ArrayList<Task>.

xinmingzh: Consider removing static as every TaskList should contain a different list.

PR #74, opened 2019-08-23
[Lim Jun Hup] Duke Increments

junhuplim:
Level-1 to Level-6 Week2


src/Duke.txt
```diff
@@ -0,0 +1,3 @@
+T | 1 |  read
```
seanlowjk, line 1: Is it really necessary for this to be shown to the user? 

I might put this in my .gitignore

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,60 @@
+package main.java.command;
+
+import main.java.components.Storage;
+import main.java.components.TaskList;
+import main.java.exception.InvalidInputException;
+
+/**
```
seanlowjk: Why is everything indented by 8 spaces?

I would prefer the indentation to be of 4 spaces to increase readability!

src/main/java/command/Command.java
```diff
@@ -0,0 +1,61 @@
...
+import main.java.exception.InvalidDescriptionException;
+import main.java.exception.InvalidInputException;
+import java.io.IOException;
+import java.text.ParseException;
+
+/**
+ * An abstract Command class with abstract method execute and isExit.
+ * This Command class handles the command given by the user.
+ */
+public abstract class Command {
```
seanlowjk: Good Job! 

I like the implementation of OOP in this class!

src/main/java/components/Duke.java
```diff
@@ -0,0 +1,58 @@
...
+import main.java.exception.DukeException;
+import main.java.command.Command;
+import java.io.IOException;
+
+/**
+ * This class is the main class of the program. It helps users to help track their tasks
+ * by creating a task list for them when users key in the respective input. The task list
+ * is then automatically saved into a txt file which allows the users to view and keep track.
+ */
+public class Duke {
```
seanlowjk: Clean code!

src/main/java/components/Storage.java
```diff
@@ -0,0 +1,312 @@
...
+import main.java.tasks.Task;
+import main.java.tasks.Deadline;
+import main.java.tasks.Event;
+import main.java.tasks.Todo;
+import java.io.*;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.regex.Pattern;
```
seanlowjk: Interesting! 

I have never thought of using Regex!

src/main/java/components/Ui.java
```diff
@@ -0,0 +1,58 @@
...
+	public void showError(Exception e) {
+		if (e instanceof InvalidInputException) {
+			System.out.println("OOPS!!! I'm sorry, but I don't know what that means :-(");
+		} else if (e instanceof EmptyDescriptionException) {
+			System.out.println(String.format("OOPS!!! The description of a %s cannot be empty.", e.getMessage()));
+		} else if (e instanceof InvalidDescriptionException) {
+			System.out.println(String.format("OOPS!!! Invalid input! Make sure your %s has a description and required" +
+					" data after /at for Event or /by for Deadline.\n", e.getMessage()));
+		} else if (e instanceof ParseException) {
+			System.out.println(String.format("Please write your deadline/event date in this format: dd/MM/yyyy HH:mm," +
```
seanlowjk: Try to keep your lines to a character limit of 80!

seanlowjk (commented):
Overall, I like your code and the clean state is has, which allows me to run through it without any difficulties! 

However, as mentioned, you might want to start improving on it by standardizing the indentation to 4 spaces and the character limit per line to 80 (if I recall correctly). For this, checkstyle will be really useful here to make sure your code is clean and consistent throughout. 

Not only that, if you were to include your "tasks.txt" file inside, I would be good to just leave it blank before committing. My suggestion will be to just clear the file every time you finish every test round with Duke. You could also add Classes for the handling and parsing of the date (and / or) time of the Deadline and Event tasks.

Finally, I would suggest catching another Exception just in case the tasks.txt file is not found. This would be very useful just in case the user accidentally deletes the file which stores the user data. 

I have also learnt much from your code, such as the usage of regex.Pattern. 

One more thing to add: Great usage of Short and Sweet commit names (even though or commit names were not that great at the start...) and consistent tagging of commits!

In conclusion, styling issues can be fixed and Date and Time could be handled by specific classes just for them. 

Good Job and Keep up the good work!

Rating: 9/10

-Sean

src/main/java/components/Ui.java
junhuplim: Is it 80? From the [Java Coding Standard](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html) it states that the hard limit can be 120.

src/main/java/command/Command.java
junhuplim: Thank you!

src/main/java/components/Duke.java
junhuplim: Thank you!

src/main/java/components/Storage.java
junhuplim: Thank you!

src/main/java/command/AddCommand.java
junhuplim: How do I actually check the indentation though? It is 4 spaces in my Intellij and I have also set the settings in the editor accordingly.

src/main/java/components/Ui.java
seanlowjk: Ahhh thanks for the clarification! 

The hard limit is indeed 120, my bad!

src/main/java/command/AddCommand.java
seanlowjk: Hmmm... 

That's weird....
I would suggest opening it in visual studio code to double check the indentation! 

That's my suggestion...

src/main/java/command/Command.java
```diff
@@ -0,0 +1,61 @@
...
+	public abstract boolean isExit();
+	
+	/**
+	 * Handles the exception thrown by respective Command execution.
+	 * Takes care of InvalidInputException, EmptyDescriptionException, InvalidDescriptionException,
+	 * Parse Exception and other unhandled Exceptions.
+	 *
+	 * @param e Exception thrown when user input is wrong.
+	 */
+	public static void handleException(Exception e) {
```
chowyiyin: You may want to let Ui handle all System.out messages? Just a suggestion :)

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,43 @@
+package main.java.command;
+
+import main.java.components.Storage;
+import main.java.components.TaskList;
+/**
```
chowyiyin: You might want to standardise the spacing between your import statements and Javadocs

src/main/java/components/Storage.java
```diff
@@ -0,0 +1,312 @@
...
+	
+	/**
+	 * This method stores the event task keyed in by the user into the file.
+	 *
+	 * @param input the description of the event task
+	 * @throws EmptyDescriptionException if the description is empty for the event task.
+	 * @throws InvalidDescriptionException if the format of the event keyed in by the user is wrong.
+	 * @throws ParseException if the date format for the event keyed in by the user is wrong.
+	 */
+	public void storeEvent(String input) throws EmptyDescriptionException, InvalidDescriptionException, ParseException{
```
chowyiyin: Maybe you could let your exceptions extend DukeException then you wouldn't have to specify so many exceptions

src/main/java/components/Storage.java
```diff
@@ -0,0 +1,312 @@
...
+	 * This method completes the task as requested by the user input in the file. It rewrites
+	 * everything from the saved file and writes from the updated task list in the program.
+	 *
+	 * @param input the command given by the user input which includes the index of the
+	 *              array list that user wants to complete.
+	 */
+	public void updateComplete(String input) {
+		try {
+			BufferedReader br = new BufferedReader(new FileReader(filePath));
+
```
chowyiyin: You might have forgotten to delete this empty line.

chowyiyin (commented):
Good job 👍 Your Javadoc comments are succinct and your code is neat. One thing you could do though is to try to limit your interaction with the user to the Ui class (all the System.out.println(…)). For coding style, like what Sean mentioned, you might want to check the indentation settings. 

Also, just a personal opinion, when parsing the input commands, it might be easier to use the split method for Strings instead so that you don't have to take care of the substring indexes but it's just a personal preference because I had some experience of dealing with strings using substrings and the indexes gave me quite a few off by one errors.

PR #76, opened 2019-08-23
[SHUI YAO] Duke Increments

shuiyao-sg:


Parcly-Taxel (commented):
There is a dedicated [DateTimeFormatter](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html) in Java that can replace the custom date-handling code you have written. Wouldn't you agree that's better?

You can also fold `{Todo,Event,Deadline}Command` into a single `AddCommand` holding the `Task` to be added, for cleaner/less repetitive code.

Your command classes also include the system's responses, hard-coded. Have you also thought of splitting them into the `TextUi` class so that the messages could potentially be swapped in the future?

Lastly, a bit more code could be added to handle cases like no matching tasks, no tasks in the list, etc. – a little more friendly to the user.

src/main/java/cs2103t/duke/Deadline.java
```diff
@@ -0,0 +1,52 @@
...
+        this.by = MyDate.genMyDate(by);
+    }
+
+    /**
+     * Generates a cs2103t.duke.Deadline task from input String.
+     *
+     * @param s deadline date and time represented by String.
+     * @return cs2103t.duke.Deadline task.
+     */
+    public static Deadline genDeadline(String s) {
```
eejian97: Hi there, i think it is a very good idea to have a factory method before instantiating the object. This will allow you to catch any exceptions in a very elegant way.

src/main/java/cs2103t/duke/date/MyDate.java
```diff
@@ -0,0 +1,102 @@
+package cs2103t.duke.date;
+
+/**
+ * Encapsulates a date and time.
+ */
+public class MyDate {
```
eejian97: Hi there, can i ask why you decided to create your own MyDate object instead of simply using the LocalDate / LocalTime that is already pre-defined by java? That would save you alot of trouble . Thanks!

eejian97 (commented):
Overall the code is very well structured and very readable! Very clean . I have learnt alot . Thanks!

PR #78, opened 2019-08-23
[Lee Ee Jian] Duke Increments

eejian97:
Includes duke increments up to level9.


src/main/java/Command.java
```diff
@@ -0,0 +1,7 @@
+public abstract class Command {
+
+    abstract boolean isExit();
```
Parcly-Taxel, line 8: There is only one command that leads to the program exiting, which is `bye`. As such, it would be better to have a Boolean variable `isExiting` that is initialised to `false` (by default) and only set to `true` in the ExitCommand.

src/main/java/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class Ui {
+    private Scanner sc;
+
+    public Ui() {
+        this.sc = new Scanner(System.in);
+    }
+
+    /**
+     * printLine() will print the separator lline from input and output.
```
Parcly-Taxel: This should be line and not **l**line. Perhaps you should use IntelliJ's code inspection utilities (Analyze → Inspect Code), so any such typos can be caught before they are committed?

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,18 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class DeadlineTest {
+
+    @Test
+    void toDataFormatTest() {
+        Deadline d = new Deadline("return book","2/12/2019", "1800");
+        assertTrue(d.toDataFormat().equals("D | 0 | return book | 2/12/2019, 1800"));
+    }
+    /*
```
Parcly-Taxel, line 11: It's always better to include as many tests as you can, rather than leavinng them commented out, don't you agree?

src/main/java/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.FileWriter;
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.PrintWriter;
+
+public class Storage {
+    String filePath;
+
+    /**
+     * Storage constructor.
```
Parcly-Taxel: According to the CS2103 coding standard, a line should be left between the description in a Javadoc comment and the @param/@return fields.

Parcly-Taxel (commented):
When it comes to methods in classes, there should not be a space immediately after the class's declaration and immediately before its closing brace. Your code is inconsistent with regards to this.

Later on in the project you should be putting the classes into packages, and that will cause some access problems – `description` in the `Task` class is `protected`, but once that is moved into its own package you should make it private and implement a getter method to access the description (for the `find` command).

I suggest that the logic for each command be within the corresponding command class, rather than concentrated in the `TaskList` class. You're "jumping through too many hoops" this way. Do you think the code would be simpler following the alternative?

There are also several lines of code dedicated to handling input, which I think could be done away with. You are in control of the style – it will be all the better if you can strip away some of the parsing code in the pursuit of elegance.

shuiyao-sg (commented):
Regarding the JavaDoc, in [CS2103T Coding Standard](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#comments), it is suggested that:

>In method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)

Therefore, instead of commenting like:

> execute() will add newTask to the TaskList.

You may want to write "Adds new Task to the TaskList."

In addition, "/**" should reside on a separate line for block comments, meaning that you are advised to start your first sentence of the block comment from the second line (you may refer to your Deadline.java file).

Lastly, it would be better if you can add JavaDoc comments for each public class.

Your design looks good!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public Deadline(String description, String date, String time) {
+        super(description);
+        this.date = LocalDate.parse(date, DateTimeFormatter.ofPattern("d/MM/yyy"));
+        this.time = LocalTime.parse(time, DateTimeFormatter.ofPattern("HHmm"));
+    }
+
+    @Override
+    public String toDataFormat() {
+        return "D | " + super.toDataFormat() + " | " + date.format(DateTimeFormatter.ofPattern("d/MM/yyyy"))
```
Yehezkiel01, line 66: I would suggest that you create a class for formatting date. Therefore, you do not need to change `Deadline` and `Event` classes when the date format changes.

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param cmd is the full command String
+     * @return the appropriate Task
+     * @throws InputMismatchException when user does not type correct input
+     */
+    private static Task handleNewTask(String keyword, String cmd) throws InputMismatchException {
+        String date = "";
+        String time = "";
+        String desc = "";
+        if (keyword.equalsIgnoreCase("deadline")) {
+            String descriptionAndTime = cmd.substring(8);
```
Yehezkiel01: I would suggest you to break down this method into several methods such as handleDeadline, handleEvent, etc.

src/main/java/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+     * load() will read the data file.
+     * @return will return an ArrayList of Task after reading data file
+     */
+    public ArrayList<Task> load() {
+        try {
+            File dataFile = createOrRetrieve(filePath);
+            ArrayList<String> stringOfTasks = readFile(dataFile);
+            ArrayList<Task> arrOfTasks = new ArrayList<>();
+
+            String keyword = "";
```
Yehezkiel01: I would suggest to move these variables into the appropriate scope.

PR #79, opened 2019-08-23
[Yoke Kai Xin] Duke Increments

kxyoke:


yeodonghan:
- There is a good number of test methods and classes. 
- Code is clean and compliant with the coding standards. 👍
- I would like to understand more as to why you choose to package your files into separate packages.

Sahilgat:
- I liked your use of enums and switch-case to handle the logic of which command was inputted, it is much neater than my implementation of if-else and makes the code more readable

- I liked how you used "String.format("%d.%s", listIndex, task.toString()" to print the tasks instead of the messier (and possibly inefficient) String concatenation I was using and I'll likely implement this as well! 

- Your javadoc comments are descriptive and code complies with coding standards

- Tests are very comprehensive! I didn't know about the @BeforeEach feature which would have been very useful to use and I'll try to implement it from now on

- Exceptions are well structured as well

- For your file writing, you used BufferedWriter to write each task to the file in a for loop. However, I used a for loop and created the String using string builder then wrote it to the file in one instance. I'm wondering which method is more efficient and I think it was interesting you thought of that approach as writing to the file line by line seems more intuitive and I'm not sure how I didn't think of something so seemingly obvious!

PR #81, opened 2019-08-23
[Lim Yu Hui] Duke Increments

mannggoo:


src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+        //return "Duke heard: " + input;
+        return run(input);
+    }
+
+    /**
+     * Generate Duke's welcome message.
+     * @return a string representing the welcome message.
+     */
+    public String showWelcome() {
+        String s1 = ui.showWelcome();
```
chensu2436: One-level abstraction

src/main/java/Parser.java
```diff
@@ -0,0 +1,330 @@
...
+
+
+
+    /**
+     * Generates the appropriate task that matches the command.
+     * @param task command to create the task.
+     * @return Task that is generated.
+     * @throws DukeException Exceptions specific to Duke
+     */
+    public static Task generateNewTask(String task) throws DukeException {
```
Luk723, line 110: Consider shortening this method.  It is lengthy

src/main/java/Parser.java
```diff
@@ -0,0 +1,330 @@
...
+    /**
+     * Processes the Command given by the user to Duke.
+     * @param command String representing the command given to Duke.
+     * @return A string representing the reply from Duke.
+     */
+    public String processCommand(String command) {
+        String reply = "";
+
+        // When command is 'list'
+        if (command.equals("list")) {
```
SendorasRight: Is there a reason why you opted to use multiple if-else blocks? a switch statement might be better choice for such implementations as it implicitly uses a hash table to perform the checks

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+        String s2 = "Here is the list of tasks from where you've left off: ";
+        String s3 = ui.showList(taskList);
+        return s1 + "\n" + s2 + "\n" + s3;
+    }
+
```
cheongliyuan: Hi Yu Hui, 
consider using different names for Duke's showWelcome() and ui.showWelcome(). After a while, you might forget the naming. Different naming might help with clarity.

src/main/java/Parser.java
```diff
@@ -0,0 +1,330 @@
...
+            }
+
+            Task newTask = new Task("dummy");
+
+            // Create the appropriate Task type
+            if (type.equals("todo")) {
+                newTask = new ToDo(taskDescription);
+            }
+
+            try {
```
Luk723: The nesting of tries is too deep.

SendorasRight: Is there a reason why you you used multiple consecutive try-except blocks that are executed right after each other to create tasks? The code inside each of them are very similar and should be abstracted out into a function if possible. Moreover, it's possible to catch and handle multiple exception in one try-catch block, which makes reading the code easier

src/main/java/Storage.java
```diff
@@ -0,0 +1,129 @@
...
+        if (desc.contains("tags")) {
+            hasTags = true;
+            desc = desc.replace("(", "");
+            desc = desc.replace(")", "");
+            String[] processed = desc.split("tags:");
+            desc = processed[0];
+            tag = processed[1];
+        }
+
+        if (type.equals("T")) {
```
SendorasRight: Is there a reason why your mutually exclusive condition checking is performed in separate if-blocks here? instead of using multiple if blocks here, you should change them to if-else or switch statements so that it's clear to a third party that these conditions (equals "T", equals "D" etc.) are mutually exclusive events and that only one branch should be taken

src/main/java/Duke.java
mannggoo: ohhh that's a good point, I didn't see that, thanks!

PR #83, opened 2019-08-23
[Bryan Lim Yan Jie] Duke Increments

blimyj:


src/main/java/Todo.java
```diff
@@ -0,0 +1,17 @@
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+        this.taskType = TaskType.TODO;
+    }
+
+    @Override
+    public String toString() {
+        String str = "["
```
moziliar, line 9: Would it be better to use a one-liner to code this part?
`String str = "[T][" + this.getStatusIcon()+ "] "
	                + this.getDescription();`
Personally I did something like this. Could you share with me if you think otherwise is better?

src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
+    public Duke() {
+        
+        //Initialize FileHandler
+        fileHandler = new FileHandler("../data", "save1.txt");
+
+        //Try to load old data.
+        try {
+            tasks = fileHandler.loadData();
+        } catch (DukeException e) {
+            //Does nothing for now
```
moziliar: Could you elaborate on this comment?

src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
+            tasks = new TaskList(new LinkedList<Task>());
+        }
+    }
+    
+    public void main(String[] args) {
```
moziliar: Try not to leave empty class with overloaded name around. It creates both confusion and potential hazard.

src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
+        } catch (DukeException e) {
+            LinkedList<String> msg = new LinkedList<>();
+            msg.add(e.getMessage());
+            output = this.formattedPrint(msg);
+        }
+        
+        return output;
+    }
+    
+    public String formattedPrint(LinkedList<String> strings) {
```
moziliar: As UI is meant for interaction with the user and thus the iostream, and Parser is, namely, meant for parsing of text, would it be a good idea to put this inside either ui or parser (personally prefer parser) as a static method?

src/main/java/FileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    
+    public void saveTasks(TaskList tasks) throws DukeException {
+        
+        String strToAdd;
+        String filePath = saveLocation + "/" + fileName;
+        //Future: Check if filePath is valid?
+
+        try {
+            //Check if dir exists, creates if dir does not exist.
+            File dataDir = new File(saveLocation);
```
moziliar, line 32: Wow good practice:)

src/main/java/FileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+            if (!dataDir.exists()) { 
+                dataDir.mkdirs(); 
+            }
+            //Future: Check if parent dir exists?
+
+            //Check if file exists, creates if file does not exist.
+            File saveFile =  new File(saveLocation + "/" + fileName);
```
moziliar, line 39: Before new there were 2 spaces.

src/main/java/FileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+            if (!saveFile.exists()) { 
+                saveFile.createNewFile();
```
moziliar, line 41: Good:)

src/main/java/FileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+            //Check if file exists, creates if file does not exist.
+            File saveFile =  new File(saveLocation + "/" + fileName);
+            if (!saveFile.exists()) { 
+                saveFile.createNewFile(); 
+            }
+            
+            //Initialise FileWriter
+            FileWriter fw = new FileWriter(filePath);
+            Task task;
+            for (int i = 0; i < tasks.getTaskListLength(); i++) {
```
moziliar, line 47: Good use of methods to achieve abstraction. Will learn from you. Kudos

src/main/java/FileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                    newTask = new Todo(taskArr[2]);
+                    break;
+                case "DEADLINE":
+                    newTask = new Deadline(taskArr[2], taskArr[3]);
+                    break; 
+                case "EVENT":
+                    newTask = new Event(taskArr[2], taskArr[3]);
+                    break;
+                default:
+                    newTask = new Todo("Error Task");
```
moziliar, line 93: Not very good to generate such task I think?

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
+import java.util.LinkedList;
+
+public class Parser {
+    
+    public static LinkedList<String> parseAndExecute(String input, TaskList tasks, FileHandler fileHandler)
```
moziliar, line 19: Personally, I think Parser only processes an unformatted String, i.e. a user input that must be broken down, and return one single type of object (in my case, a Command, because a user input **is** a command). With that, the Parser knows very little about other classes like TaskList and FileHandler, thus decoupling it from the rest. This is one of the software engineering principle which will help when you try to test Parser. This is so because you would only need to ensure the command is correct before you can test it properly. In the current state when you want to test Parser, you will have to make sure that TaskList and FileHandler are running correctly, which complicates the work. (Might not be depending on how you order the testings) But generally decoupling is encouraged. Hope this helps! You can also share your view on this so that we could discuss together.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+        //Split input incase it has a command and argument
+        String[] inputArray = input.split(" ", 2);
+
+        String[] cmdArgs = new String[0];
+        Task newTask;
+
+        LinkedList<String> result = new LinkedList<>();
+
+        switch (inputArray[0]) { 
+        case "list":
```
moziliar, line 30: Would it be better to have a Command instance to take care of this instead? Having a Parser to add tasks into the task list would not be favourable, given that a Parser should only parse the text.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+            Task currTask;
+
+            //Append to result
+            for (int i = 1; i <= tasks.getTaskListLength(); i++) {
+                currTask = tasks.getTask(i - 1);
+                result.add(i + ". " + currTask);
+            }
+
+            break;
+        case "done":
```
moziliar, line 40: Same as above. What I did was to generate a DoneCommand and let it take care of all the following lines in `public void execute()` method. You can read more about it by searching Command pattern, one of the software engineering patterns (in CS3219 I think).

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+
+            //Saving
+            fileHandler.saveTasks(tasks);
+
+            //Append to result
+            result.add("Nice! I've marked this task as done:");
+            result.add("  " + modifiedTask.toString());
+
+            break;
+        case "todo":
```
moziliar, line 59: Same as above.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+            //Saving
+            fileHandler.saveTasks(tasks);
+
+            //Append to result
+            result.add("Got it. I've added this task:");
+            result.add("  " + newTask.toString());
+            result.add("Now you have " + tasks.getTaskListLength() + " task(s) in the list.");
+
+            break;
+        case "deadline":
```
moziliar, line 94: Same as above.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+            //Saving
+            fileHandler.saveTasks(tasks);
+
+            //Append to result
+            result.add("Got it. I've added this task:");
+            result.add("  " + newTask.toString());
+            result.add("Now you have " + tasks.getTaskListLength() + " task(s) in the list.");
+
+            break;
+        case "event":
```
moziliar, line 75: Same as above.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+            //Saving
+            fileHandler.saveTasks(tasks);
+
+            //Append to result
+            result.add("Got it. I've added this task:");
+            result.add("  " + newTask.toString());
+            result.add("Now you have " + tasks.getTaskListLength() + " task(s) in the list.");
+
+            break;
+        case "delete":
```
moziliar, line 112: Same as above.

src/main/java/Task.java
```diff
@@ -0,0 +1,66 @@
+public abstract class Task {
+    protected String description;
+    protected boolean isDone;
+    protected TaskType taskType;
+    
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        this.taskType = TaskType.UNDECIDED;
```
moziliar, line 14: If Task is abstract (which it should be), do we have to initialise TaskType since we would not instantiate a Task Object?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,26 @@
+import java.util.LinkedList;
+
+public class TaskList {
```
moziliar, line 3: Personally, I use TaskList to do a lot of processing and also the generation of reports (formatted text to be stored inside the data.txt file and output list).

src/main/java/TaskType.java
```diff
@@ -0,0 +1,3 @@
+public enum TaskType {
```
moziliar, line 1: Good call:) Easier to manage task types.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
+import java.time.LocalDateTime;
+import java.time.DateTimeException;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
+    protected LocalDateTime byDateTime;
```
moziliar: Would names like "deadline" be better? I was a little bit puzzled by the name haha.

src/main/java/UI.java
```diff
@@ -0,0 +1,67 @@
...
+        startMessage.add("|____/ \\__,_|_|\\_\\___|");
+        startMessage.add("Hello! I'm Duke");
+        startMessage.add("What can I do for you?");
+        
+        this.formattedPrint(startMessage);
+        
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+        
+        while (!input.equals("bye")) {
```
moziliar: Ugh not sure how this works with the GUI. I could not generate the welcome message somehow. But I abandoned while loop and the program still runs on the GUI. I guess the Main/Launcher/Application class has something like that to keep it going?

moziliar (commented):
It would be better to make folders to package the different classes so that they don't look that messy? Would be good to package files that are related to each other when you try to limit access with protected or default access modifier. Also, Javadoc not in place. I know it is a pain and it took me about close to half an hour to finish all the methods Javadoc last week, but it is important. 

Good job overall! A lot of learning points for me as well. Will try to learn from your coding style as it looks more secure than mine! Cheers:)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
+    protected LocalDateTime byDateTime;
+
+    public Deadline(String description, String byDateTime) throws DukeException {
+        super(description);
+        this.taskType = TaskType.DEADLINE;
+        try {
+            this.byDateTime = LocalDateTime.parse(byDateTime, DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm"));
```
ryantay232: Thanks for telling me about this tip! Didn't realise there is an easier way to format date and time.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public String toSaveString() {
+        String saveString = super.toSaveString() + "@@@" + this.getByDateTimeString();
+        return saveString;
+    }
+    
+    @Override
+    public String toString() {
```
ryantay232, line 47: Looks like a clean way to seperate strings and methods to be added into a string.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+                        + "you want to mark as done!");
+            }
+            int taskNum = Integer.parseInt(inputArray[1]);
+            int taskIndex = taskNum - 1;
+
+            Task modifiedTask = tasks.getTask(taskIndex);
+            modifiedTask.markAsDone();
+
+            //Saving
+            fileHandler.saveTasks(tasks);
```
ryantay232, line 52: Would it be better that the saving is outside of the switch case? So there will be lesser repeated lines.

ryantay232 (commented):
Learned a lot from reading your code! Overall clean code and very understandable.

src/main/java/Parser.java
blimyj: I like the idea, in this case, I was considering the possibility of other commands that might not modify the TaskList so it made more sense to me to explicitly call saveTasks() only for commands that needed it.

src/main/java/Deadline.java
blimyj: Thanks! :D

src/main/java/Todo.java
blimyj: I think either way works, but I did it in multiline in case I would have to append other strings in the future. To me, if it grows, it would be easier to maintain the coding style of 120chars and clearer to interpret.

src/main/java/Duke.java
blimyj: Heh, caught red-handed...
I removed the comment.

src/main/java/Duke.java
blimyj: Thanks for the tip! 😄 
I removed it.

src/main/java/Duke.java
blimyj: Good idea! Implemented.

src/main/java/FileHandler.java
blimyj: Thanks! 😄

src/main/java/Task.java
blimyj: I was thinking that I might initialize Task via an anonymous class for stubbing in the tests. 
I have no idea if this is good practice or not. Eager to hear your input! :D

src/main/java/TaskType.java
blimyj: Thanks!

src/main/java/Deadline.java
blimyj: Implemented. Appreciate for the feedback! 😄

src/main/java/UI.java
blimyj: Heh, caught red-handed once again... (was rushing for the deadline).
I removed the bulk of the UI class and replaced it with static methods following your suggestion above.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+     * Creates a deadline event.
+     * @param description task description
+     * @param byDeadline deadline for task
+     * @throws DukeException when there is a problem parsing the date time string provided by user
+     */
+    public Deadline(String description, String byDeadline) throws DukeException {
+        super(description);
+        this.taskType = TaskType.DEADLINE;
+        try {
+            this.byDeadline = LocalDateTime.parse(byDeadline, DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm"));
```
moziliar, line 19: Would it be a better idea to save the pattern as a final static constant (preferably in Parser class so that you can call it by Parser.DATE_PATTERN)?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+public class Deadline extends Task {
+    protected LocalDateTime byDeadline;
+
+    /**
+     * Creates a deadline event.
+     * @param description task description
+     * @param byDeadline deadline for task
+     * @throws DukeException when there is a problem parsing the date time string provided by user
+     */
+    public Deadline(String description, String byDeadline) throws DukeException {
```
moziliar, line 15: Over here, Deadline takes in a string of deadline. Would it be alright to take in a LocalDateTime instead, as the parsing role should be Parser's?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+                    + "format." 
+                    + e.getMessage());
+        }
+    }
+
+    /**
+     * Converts DateTime to String in the format of dd/MM/yyyy HH:mm.
+     * @return DateTime as String in the format of dd/MM/yyyy HH:mm
+     */
+    public String getByDeadlineString() {
```
moziliar, line 31: Could it be a better idea to return a LocalDateTime and let the user of its return value decide how to format it? Let's say your task list is using it, it could decide on which pattern to display the date.

src/main/java/Event.java
```diff
@@ -0,0 +1,60 @@
...
+public class Event extends Task {
+    protected LocalDateTime atDateTime;
+
+    /**
+     * Creates an Event Object.
+     * @param description task description
+     * @param atDateTime time of event
+     * @throws DukeException when there is a problem parsing the date time string provided by user
+     */
+    public Event(String description, String atDateTime) throws DukeException {
```
moziliar, line 16: Same as Deadline.

src/main/java/Task.java
moziliar: Could you elaborate more on the anonymous class part? Thanks.

src/main/java/UI.java
```diff
@@ -0,0 +1,24 @@
...
+    
+    //public static final String ENDMESSAGE = "Bye. Hope to see you again soon!";
+
+    /**
+     * Appends a newline after every String for Duke to output.
+     * @param strings LinkedList of Strings to concatenate
+     * @return String of concatenated strings with newline after each string has been concatenated.
+     */
+    public static String formattedPrint(LinkedList<String> strings) {
+        String output = "";
```
moziliar, line 17: Would be nice to use a StringBuilder here. The equivalent code would be 
```StringBuilder output = new StringBuilder();
for (String string: strings) {
    output.append(string);
    output.append("\n");
}

return output.toString();
```

moziliar (commented):
Good refactoring. Not much major redesign of logics, but the codes are more readable now.

src/main/java/Parser.java
```diff
@@ -0,0 +1,189 @@
...
+
+            //Append to result
+            result.add("Noted. I've removed this task:");
+            result.add("  " + removedTask.toString());
+            result.add("Now you have " + tasks.getTaskListLength() + " task(s) in the list.");
+
+            break;
+        case "remind":
```
moziliar, line 129: Would it be a better idea to extract all these codes into a separate method? It looks super lengthy and affects readability. This code doesn't fulfil Single Responsibility Principle as it has several reasons to change. Let's say the execution in case "remind" gets changed, your code in this parseAndExecute would have to change (because the remind case must change); then when execution of delete changes, this method would also have reason to change. It increases the chances of having bugs in this method and unit test for this method will be complicated. Instead, my approach is to have handleRemind() and handleDelete() etc. to handle all these executions. These methods can be tested separately and parseAndExecute would have only 1 reason to change: that is, when there are more types of command/execution to be added (say, listFull) and only that.

moziliar (commented):
Coupling of methods is surging and it will make the unit test harder. Would recommend refactoring to decouple all the methods such that they can fulfil SRP.

PR #86, opened 2019-08-23
[Koh Wee Lun Clarence] Duke Increments

fatclarence:
In Lecture pull request


src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    protected int deletedTask;
+
+    /**
+     * Class constructor.
+     *
+     * @param deleted The zero-based index of the task to be deleted from the task list.
+     */
+    public DeleteCommand(int deleted) {
+        super();
```
wh-coder-2025, line 28: super() constructor not required here.

src/main/java/duke/command/DoneCommand.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    protected int doneTask;
+
+    /**
+     * Class constructor.
+     *
+     * @param doneTask The zero-based index of the task to be marked as done from the task list.
+     */
+    public DoneCommand(int doneTask) {
+        super();
```
wh-coder-2025, line 27: super() not required.

src/main/java/duke/command/FindTaskCommand.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    protected String keywords;
+
+    /**
+     * Class constructor specifying the keywords to search for from the tasklist.
+     *
+     * @param keywords The keywords to be searched.
+     */
+    public FindTaskCommand(String keywords) {
+        super();
```
wh-coder-2025, line 31: super() not required.

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Clarence Koh
+ * @version 1.0
+ * @since 29th August 2019
+ */
+public class ListCommand extends Command {
+    /**
+     * Class constructor.
+     */
+    public ListCommand() {
```
wh-coder-2025: Since the superclass has a parameter-less constructor, we do not need to write a parameter-less constructor for the subclass that calls super() only. Removing this should still allow the code to be compiled since Java will insert a parameterless constructor into the subclass during compile time as long as there is a parameterless constructor in the superclass.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param userInput The user's input string.
+     * @throws DukeException If there is no specified date or task.
+     */
+    private static void deadlineCheck(String[] tasks, String userInput) throws DukeException {
+        if (tasks.length <= 1) {
+            throw new DukeException("OOPS!!! The description of a deadline cannot be empty.");
+        } else if (!userInput.contains("/by")) {
+            throw new DukeException("OOPS!!! Deadline must include /by (date to complete task).");
+        } else if (userInput.substring(userInput.indexOf("/by") + 3).equals("")
+                || userInput.substring(userInput.indexOf("/by") + 4).equals("")) {
```
wh-coder-2025: Don't you will need to perform the second check. If the user had input a date after "/by", then the substring that is 3 index after "/by" should contain some items. As an example,

"deadline homework /by           02/09/2019" 
-> userInput.substring(userInput.indexOf("/by") + 3) should return "           02/09/2019" and so it will return as false.
So, the second check is not necessary for this case. 

Likewise, if the second check is true, then the first check should return true too unless the user decided to input "/bya" or any extra character after "/by".

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param tasks The user's input string in separated into an array.
+     * @param userInput The user's input string.
+     * @throws DukeException If there is no specified date or task.
+     */
+    private static void eventCheck(String[] tasks, String userInput) throws DukeException {
+        if (tasks.length <= 1) {
+            throw new DukeException("OOPS!!! The description of a event cannot be empty.");
+        } else if (!userInput.contains("/at")) {
+            throw new DukeException("OOPS!!! Event must include /at (time of event).");
+        } else if (userInput.substring(userInput.indexOf("/at") + 3).equals("")
```
wh-coder-2025: The same issue here.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+                    return new DeleteCommand(taskNumber - 1);
+                } catch (ArrayIndexOutOfBoundsException e) {
+                    throw new DukeException("OOPS!!! Provide a task number to delete!");
+                }
+            } else {
+                switch (instruction) {
+                case "todo": {
+                    try {
+                        todoCheck(task);
+                        Task todo = new Todo(command.substring(5));
```
wh-coder-2025, line 206: What happens if the user decides to enter "todo __________ sleep" instead of "todo sleep"? "_" indicates a whitespace.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+                        return new AddCommand(todo);
+                    } catch (DukeException e) {
+                        System.err.println("Something went wrong: " + e.getMessage());
+                        break;
+                    }
+                }
+                case "deadline": {
+                    try {
+                        deadlineCheck(task, command);
+                        Task deadline = new Deadline(command.substring(9, command.indexOf("/by") - 1),
```
wh-coder-2025, line 215: Same issue as the code for todo.
Would it work if the user enters "deadline2103T iP /by 20/10/2019 2000"?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+                        return new AddCommand(deadline);
+                    } catch (DukeException e) {
+                        System.err.println("Something went wrong: " + e.getMessage());
+                        break;
+                    }
+                }
+                case "event": {
+                    try {
+                        eventCheck(task, command);
+                        Task event = new Event(command.substring(6, command.indexOf("/at") - 1),
```
wh-coder-2025, line 225: Same issue as above.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+                }
+                    case "find": {
+                        try {
+                            findCheck(task);
+                            return new FindTaskCommand(command.substring(5).trim());
+                        } catch (DukeException e) {
+                            System.err.println("Something went wrong: " + e.getMessage());
+                            break;
+                        }
+                    }
```
wh-coder-2025, line 238: Just noticed that you enclosed each of the case blocks in braces. Not necessary to do so

wh-coder-2025 (commented):
Good points:
1. Neat code! I can understand what you are trying to do easily.
2. I like the way you modelled the different "objects" in play.

Some suggestions:
Do think of more ways in which the user may accidentally break your code.

src/main/java/duke/parser/Parser.java
fatclarence: Good point but I do not actually agree with this, for if a user were to do that, he might be keen to format it like that on purpose. If I want to include 8 spaces just because I am keen to do so, but the programme keeps restricting me from doing so, then it is a bit too annoying. Most to-do lists out there do not have such a behaviour.

src/main/java/duke/parser/Parser.java
fatclarence: Does it affect the performance of the programme? :O Thanks for this though!

wh-coder-2025: I don't think it affects the performance. Maybe it is my first time seeing braces being used to enclose an environment within each case statement, so I had a difficult time looking for the matching braces :x

src/main/java/duke/parser/Parser.java
wh-coder-2025: That's actually a good point that you raised. At first, I was thinking that it would be better to check whether commands like "todo __________ sleep" would be valid. But to think about it, it might actually be possible for us to accept such commands while ensuring that the output syntax aligns with how it should be like. One possible way which I thought of was:

String toDoName =command.substring(5).trim();  // trim() helps to remove any leading and trailing whitespaces.
Task todo = new Todo(toDoName);

Sounds like a good thing for us to implement. Instead of outright rejecting the user's input, we can help them to trim it as long as they input the correct details in the correct order.

src/main/java/duke/parser/Parser.java
```diff
@@ -78,6 +78,19 @@ private static void eventCheck(String[] tasks, String userInput) throws DukeExce
...
+     * a keyword to search for from the task list.
+     *
+     * @param tasks The user's input split into an array of strings.
+     * @throws DukeException If there is no keyword to search for.
+     */
+    private static void findCheck(String[] tasks) throws DukeException {
+        if (tasks.length <= 1) {
+            throw new DukeException("OOPS!!! Please tell me what you want to find!!");
+        }
+    }
```
bakwxh, line 128: Quite interesting that you have a separate methods to check the validity of a command, as opposed to checking the validity within your FindTaskCommand/AddCommand and throwing your DukeException there. I didn't ever encounter a need for this approach, I'm curious is this method works better or more reliably?

PR #87, opened 2019-08-23
[Dominique Ng] Duke Increments

hellodommy:
Add error handling for todos and deadlines without specified date/time


hellodommy:
Complete Week 3 increments

src/main/java/Duke.java
```diff
@@ -1,10 +1,59 @@
...
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                ui.showLine();
+                Command c = Parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
+            } catch (DukeException e) {
+                System.err.println(e.getMessage());
```
tanlk99: Printing the exceptions to stderr is a good idea, but it might be better to abstract the error printing and formatting (the "OOPS" part of the string) to the Ui class.

tanlk99 (commented):
**General comments**

- Your code looks really clean and has followed the style guide quite well. Good job!
- While you have properly left all class variables as `private` or `protected`, I feel that there are some methods that might be better left as `default` instead of `public`, like `getTasks()` in TaskList.
- In the `DateTime` class, it might be possible to use [Calendar](https://docs.oracle.com/javase/7/docs/api/java/util/Calendar.html) which would simplify the code.

**Specific comments**
https://github.com/nus-cs2103-AY1920S1/duke/blob/ea3bd1619e7bc248350949b8816c170b5de4f56a/src/main/java/DateTime.java#L51-L57
Nice attention to detail by adding the date suffixes to the time format. However, I've noticed that the 11th, 12th and 13th will be displayed incorrectly under this logic.

https://github.com/nus-cs2103-AY1920S1/duke/blob/ea3bd1619e7bc248350949b8816c170b5de4f56a/src/main/java/Duke.java#L52
Printing the output from the `DukeExceptions` to `System.err` is a good idea. It may be even better to abstract the formatting (the "OOPS" and ":-(" part) to the `Ui` class instead. I've also noticed that some of your `IOExceptions` such as those in `DeleteCommand` or `AddCommand` were printed to `System.out` instead. Perhaps it would be better to switch them to `System.err` as well?

hellodommy:
@tanlk99 Thank you for your feedback! 😊
I've implemented the changes to the date suffix and `System.err`.
I'll look into the access modifiers and the `Calendar` as well, seems useful! ✌🏻

src/main/java/Duke.java
```diff
@@ -1,10 +1,65 @@
+import duke.command.Command;
+import duke.exception.DukeException;
```
ManasVegi: sfv,jbekivheto

src/test/java/DateTimeTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+    @Test
+    public void TestDateTime_validInput_success() {
+        assertEquals("31st of April 2019, 4:32pm", (DateTime.create("31/4/2019 1632")).toString());
+        assertEquals("22nd of September 2030, 12:00am", (DateTime.create("22/9/2030 0000")).toString());
+        assertEquals("3rd of March 1930, 8:20am", (DateTime.create("3/3/1930 0820")).toString());
+        assertEquals("16th of February 1966, 7:12pm", (DateTime.create("16/2/1966 1912")).toString());
+    }
+
+}
```
ManasVegi, line 15: FYI: I've been struggling to set up JUnit for so long. I observed your project structure and emulated it and it worked. So thanks a lot.

ManasVegi:
**General Comments**

- Your OOP integration is really extensive. Great job on that. Top notch modularity.
- JavaDocs and coding style seems perfect.
- It is nice that *DateTime* is separate so you are abstracting a lot. But probably using Calendar instead of all the attributes could be easier? (as mentioned by @tanlk99 )
- Would you like to organize your files into packages?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,63 @@
...
+
+import duke.datetime.DateTime;
+import duke.exception.DukeException;
+
+/**
+ * Events are tasks that start at a specific time and ends at a specific time.
+ */
+public class Event extends Task {
+
+    protected DateTime dateTime;
```
ManasVegi: Both Event and Deadline look to have the same dateTime implementation.
To differentiate it from Deadline, probably you could try implementing Event with startTime and endTime? In this way Events can represent a duration like the Events in Google Calendar.

src/main/java/duke/datetime/DateTime.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Creates a DateTime object by splitting up the date and time given in a string.
+     *
+     * @param dateTimeString Date and time given in the format DD/MM/YYYY HHMM.
+     * @return DateTime object created.
+     * @throws ArrayIndexOutOfBoundsException If date and time given is not in the accepted format.
+     */
+    public static DateTime create(String dateTimeString) throws ArrayIndexOutOfBoundsException {
```
ManasVegi: If you want to make the code more foolproof, you can throw an exception when you encounter 

1. negative date / negative time

1.  time and day and/or month that are out of the allowed range of values. Using Calendar class might help in some of these.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Reads commands give to Duke.
+     * Creates the appropriate command object.
+     *
+     * @param fullCommand Whole command where the first word will be extracted.
+     * @return Command object that is created according to the type of command.
+     */
+    public static Command parse(String fullCommand) {
+        String[] currArray = fullCommand.split("\\s+", 2);
```
ManasVegi: Nice touch on the limit for split method!

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    public void findTasks(String searchTerm, Ui ui) {
+        ArrayList<Task> searchResults = new ArrayList<>();
+        for (Task task : tasks) {
+            if (task.getDescription().contains(searchTerm)) {
+                searchResults.add(task);
+            }
+        }
+        if (searchResults.isEmpty()) {
+            ui.show("OOPS!!! You don't have any tasks containing the term \"" + searchTerm + "\".");
```
ManasVegi: Nice touch on printing the searchTerm also when reporting error.

hellodommy:
@ManasVegi Thanks for the feedback! I'm glad some parts of my implementation could be useful to you :)

src/main/java/duke/task/Event.java
hellodommy: Yep, I just realised it's a little weird that my events do not have an end time.
I'll implement that soon, thanks for the suggestion!

PR #88, opened 2019-08-23
[Wong Shun Min] Duke increments

wongsm7:


src/main/java/Duke.java
```diff
@@ -1,10 +1,163 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
-    }
-}
+        new Duke("D:\\CS2103T\\duke\\docs\\duke.txt").run();
```
Sam-limyr: You are using an absolute filepath (e.g. "D:/CS2103T/duke") but this will not apply for all users (e.g. if they have their CS2103T folder in "D:/Desktop/CS2103T") and will affect Duke's save function. Consider using a relative filepath (e.g. "./duke/doc/duke.txt") rather than an absolute one.

src/main/java/Parser.java
```diff
@@ -0,0 +1,96 @@
...
+    public static DateTimeFormatter dtf = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+
+    /**
+     * Method to parse input.
+     * @param input which is the input string
+     * @return A Command
+     */
+    public static Command parse(String input) throws DukeException {
+        String[] temp = input.split(" ");
+        if (temp[0].equals("todo")) {
```
Sam-limyr, line 21: The code sections for 'todo', 'deadline', and 'event' are quite long. Perhaps you could consider using String.split( ) methods instead, to manipulate the description strings?

src/main/java/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+     * Method to write to the file.
+     *
+     * @param tasks a list of tasks to be written to the file
+     */
+    public void writeToFile(List<Task> tasks) throws IOException {
+        File file = new File(filePath);
+        PrintWriter pw = new PrintWriter(file);
+        String output = "";
+        for (Task t : tasks) {
+            if (t.getType().equals("T")) {
```
Sam-limyr, line 81: The string manipulation to create a formatted string is done in Storage.java; however, it would be more appropriate to include the formatted String as a method "String formatString( )" in Task.java, and override it in the subclasses of Task. Then, call the formatString( ) method in Storage.java instead.

src/main/java/Event.java
```diff
@@ -0,0 +1,46 @@
...
+    public String getAt() {
+        return this.at;
+    }
+
+    /**
+     * Override toString method.
+     */
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + at + ")";
```
openorclose: I think we can reuse `this.type` instead of hardcoding "E" here? Similar for the other tasks.

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,13 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Test Class for Duke Class.
+ */
+public class DukeTest {
+    @Test
+    public void dummyTest() {
```
openorclose, line 10: Let's remove this empty test?

src/test/java/TaskListTest.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Tests removeTask method of TaskList class.
+     */
+    @Test
+    public void removeTaskTest() throws DukeException {
+        listOfTasks = new TaskList();
+        listOfTasks.addTask(new Todo("todo buy bread"));
+        listOfTasks.addTask(new Todo("todo buy butter"));
```
openorclose, line 31: Think it's good to assert that the list of tasks here has size 2. (the bug could be that somehow one of the tasks didn't get added in properly)

src/test/java/TaskTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public void getDescriptionTest() {
+        Task t = new Task("buy book");
+        assertEquals("buy book", t.getDescription());
+    }
+
+    /**
+     * Tests getType method of Task class.
+     */
+    @Test
+    public void getTypeTest() {
```
openorclose, line 39: I'm not sure about this, but should we add tests for getters and setters?

PR #90, opened 2019-08-23
[Ding YuChen] Duke Increments

dingyuchen:
Week 2


dingyuchen:
Week 3

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,25 @@
...
+    private final String heading = "Here are the matching tasks in your list: \n";
+
+    FindCommand(String description) {
+        this.description = description;
+    }
+
+    @Override
+    public String run(TodoList tasks, Storage storage) {
+        List<Task> matches = tasks.find(this.description);
+        return heading + matches.stream().map(task -> "\t" + task.toString())
```
Nanosync: This looks neat!

src/main/java/Ui.java
```diff
@@ -1,25 +1,25 @@
 import java.util.Scanner;
 
 class Ui {
-    private final String LOGO = " ____        _        \n"
+    private final String logo = " ____        _        \n"
```
Nanosync: `logo` can be marked `static` and `final`, keeping the variable name uppercased.

src/main/java/Ui.java
```diff
@@ -1,25 +1,25 @@
...
             + "|  _ \\ _   _| | _____ \n"
             + "| | | | | | | |/ / _ \\\n"
             + "| |_| | |_| |   <  __/\n"
             + "|____/ \\__,_|_|\\_\\___|\n";
-    private final String LINE = "______________________________";
+    private final String line = "______________________________";
```
Nanosync: `line` can be marked `static` and `final`, keeping the variable name uppercased.

src/main/java/Ui.java
```diff
@@ -31,6 +31,6 @@ void printErrorMessage(DukeException e) {
     }
 
     private String encase(String message) {
-        return String.join("\n", new String[]{LINE, message, LINE});
+        return String.join("\n", new String[]{line, message, line});
```
Nanosync: Could add spacing in between braces.

Nanosync (commented):
Code looks good! Some styling changes might help.

hueyy (approved):

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
...
+    void markAsDone() {
+        isDone = true;
+    }
+
+    String saveFormat() {
+        return String.format("%s|%s", this.description, this.isDone);
+    }
+
+    private String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
```
DivineDX: Very clean and good usage of a ternary operator here as opposed to an if else statement!

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[%s] %s", getStatusIcon(), this.description);
```
DivineDX: This is really good and helps to adhere to the DRY principle.

src/main/java/Event.java
```diff
@@ -0,0 +1,32 @@
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class Event extends Task {
+    private Date startPeriod;
+    private Date endPeriod;
+    private SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
+
+    Event(String description, String period) throws ParseException {
```
DivineDX: Personally feel its better to handle the Parsing in the Parser class, and then pass it directly to the constructor on line 17, hence keeping to the Single Responsibility Principle.

src/main/java/Command.java
```diff
@@ -0,0 +1,8 @@
+public abstract class Command {
+    public String run(TodoList tasks, Storage storage) {
+        storage.save(tasks);
+        return null;
+    }
+
+    public abstract boolean isExit();
```
DivineDX: Something that can be improved and eventually retired since it only returns true for one child class. I understand that it is due to the need to break out of the while loop in the Duke run() method, but perhaps a better way is to call System.exit(0) in your run() method in ExitCommand class.

src/main/java/FindCommand.java
DivineDX: Wicked Sick! TIL \t exists

src/main/java/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+        keywords.add("done");
+        keywords.add("todo");
+        keywords.add("event");
+        keywords.add("deadline");
+        keywords.add("list");
+        keywords.add("delete");
+        keywords.add("find");
+    }
+
+    static Command parse(String[] args) throws DukeException {
```
DivineDX: Would it be better if instead of the HashSet keywords, this switch case is expanded so that for the todo, events, and deadline cases they call AddCommand()? Personally, I feel that it would make more sense to throw the DukeException in the default case, which will be reached if it fails through all the other switch cases that are the "defined commands"

src/main/java/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+                }
+                String[] dates = details[1].trim().split("-");
+                Date startPeriod = formatter.parse(dates[0].trim());
+                Date endPeriod = formatter.parse(dates[1].trim());
+                return new Event(details[0].trim(), startPeriod, endPeriod);
+            }
+            default:
+            }
+        } catch (ParseException e) {
+            throw new DukeException("Dates must have proper format!");
```
DivineDX: Probably good to state what the correct format is so they can key in correctly the next time round

DivineDX (commented):
Generally a pretty good piece of code, with nifty tricks seen throughout. I learned plenty from you in the process of reviewing your code.

I believe that its best if the run() command contains the UI object as a parameter, as then you would be able to abstract out the string outputs directly to the UI class. However, I understand that it would require a significant amount of effort refactoring.

PR #91, opened 2019-08-23
[Benedict Pak Hong Jun] Duke Increments

Nanosync:
Levels 1-6


src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "x" : " "); // return tick or no symbol
+    }
+
+    public boolean getIsDone() {
```
dingyuchen, line 32: Even though method name follows convention, perhaps it is more semantic to name `isDone`

src/test/java/duke/CommandParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+import duke.command.DeleteCommand;
+import duke.command.DoneCommand;
+import duke.command.FindCommand;
+import duke.command.ListCommand;
+import duke.exception.DukeException;
+import duke.exception.InvalidInputDukeException;
+import duke.parser.CommandParser;
+import org.junit.jupiter.api.Test;
+
+class CommandParserTest {
```
dingyuchen, line 17: Very extensive tests! Nice!

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,126 @@
...
+        PrintWriter writer = null;
+        try {
+            Path path = Paths.get(filePath);
+
+            // Create directories if necessary
+            if (path.getParent() != null) {
+                Files.createDirectories(path.getParent().getFileName());
+            }
+
+            writer = new PrintWriter(new BufferedWriter(new FileWriter(filePath)));
```
dingyuchen: Might want to consider ObjectOutputStream to abstract writing and parsing objects
https://docs.oracle.com/javase/7/docs/api/java/io/ObjectOutputStream.html

dingyuchen (commented):
Code is neat, many milestones are achieved. Good job!

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,17 @@
...
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Commands are executed by a CommandParser.
+ */
+public abstract class Command {
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    public boolean isExit() {
```
DivineDX, line 14: Possible to retire this and simply call a System.exit(0) in the ExitCommand class

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,10 @@
+package duke.exception;
+
+/**
+ * Exceptions used specifically for Duke errors.
+ */
+public class DukeException extends Exception {
```
DivineDX, line 6: Great that you created various classes inheriting DukeExceptions to improve the specificity of your exceptions thrown

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+ * Tasks are used in the entire program.
+ */
+public abstract class Task {
+    /** Task description. **/
+    private String description;
+
+    /** Status of task. **/
+    private boolean isDone;
+
+    public Task(String description) {
```
DivineDX: This constructor can possibly be retired -the overloaded constructor fulfills the same functionality if called with the isDone argument as false. Any particular reason to keep this other than to have a more convenient constructor call?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,35 @@
+package duke.task;
+
+import java.util.Date;
+
+/**
+ * Tasks used for events.
+ */
+public class Event extends Task {
```
DivineDX, line 8: Looks like the implementation is wrong. Event are tasks that **start** at a specific time and **end** at a specific time. There needs to 2 attributes called 'start' and 'end', and not 'at'

src/main/java/duke/parser/TaskParser.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Handles task parsing.
+ */
+public class TaskParser {
+    /** Syntax for date. **/
+    private static final DateFormat DATE_FORMAT = new SimpleDateFormat("d/M/yyyy HHmm");
+
+    /** Syntax for deadline. **/
+    private static final Pattern DEADLINE_PATTERN = Pattern.compile("deadline (.*) /by (.*)");
```
DivineDX, line 24: Intriguing use of pattern here, nicely done!

src/main/java/duke/parser/CommandParser.java
```diff
@@ -0,0 +1,92 @@
...
+            throw new UnknownCommandDukeException();
+        }
+
+        Matcher matcher;
+
+        if (input.equals("bye")) {
+            return new ExitCommand();
+        } else if (input.equals("list")) {
+            return new ListCommand();
+        } else if (input.startsWith("done")) {
```
DivineDX: Could possibly use a switch statement from here onwards

DivineDX (commented):
Quite a remarkable piece of code. There was significant usage of unique classes such as Pattern to assist with the functionality which makes it interesting. Other than the wrong implementation of the Event class, this was quite well done.

src/main/java/duke/task/Task.java
Nanosync: The initial design was to avoid the default false parameter from propagating throughout the program as it seemed unnecessary. Thanks for the suggestion!

PR #93, opened 2019-08-23
[Jon Chua] Duke Increments

jon-chua:
- Implemented Duke Levels 1 to 6


src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import duke.main.TaskList;
+import duke.main.Ui;
+import duke.task.Deadline;
+
+import java.util.Date;
+
+public class AddDeadlineCommand extends Command {
+    private String description;
+    private String deadlineBy;
+    private Date deadlineDate;
```
LeowWB, line 14: Between `deadlineDate` and `deadlineBy`, it seems like one of them will always be null. Maybe you should consider just using a single variable?

src/main/java/duke/command/AddEventCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import duke.main.TaskList;
+import duke.main.Ui;
+import duke.task.Event;
+
+import java.util.Date;
+
+public class AddEventCommand extends Command {
+    private String description;
+    private String eventAt;
+    private Date eventDate;
```
LeowWB, line 14: Similar to `AddDeadlineCommand`.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,28 @@
...
+import duke.main.Storage;
+import duke.main.TaskList;
+import duke.main.Ui;
+import duke.task.Task;
+
+public class DeleteCommand extends Command {
+    private int zeroBasedIndex;
+
+    public DeleteCommand(int oneBasedIndex) {
+        this.zeroBasedIndex = oneBasedIndex - 1;
```
LeowWB, line 13: Have to say that I really like this method of naming your indices. Good job 👍👍👍

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,177 @@
...
+import duke.command.ListCommand;
+import duke.exception.DukeException;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+class Parser {
+    static Command parse(String fullCommand) throws DukeException {
+        if (isExitCommand(fullCommand)) {
```
LeowWB, line 21: In my opinion, this could be better expressed as a `switch` statement. Maybe you could add some `String` or `enum` identifier to each of your command classes to facilitate this.

LeowWB:
Overall thoughts:

Really like the extent of your abstraction, especially for your command classes (and how each of them had an `execute()` function). I'd say your testing is also quite on-point; you have quite a good variety of unit test cases that should prevent your individual components for breaking. One thing you may want to try to do in general is to write more documentation, as I noticed quite a few `public` methods missing JavaDoc comments.

src/main/java/duke/main/Parser.java
```diff
@@ -67,7 +67,7 @@ private static Date parseDate(String date) throws DukeException {
         }
     }
 
-    private static boolean isDate(String input) {
+    static boolean isDate(String input) {
```
caesarpjz, line 75: Great code all around! :) Maybe you could consider the edge cases for date, for example, if the user enter something like 45/52/3012 1800 by either deadline or event.

src/test/java/duke/main/ParserTest.java
```diff
@@ -18,6 +19,7 @@
     void parserTest() throws DukeException  {
```
caesarpjz, line 19: Wow! Didn't know you could check for class types. Good job!

caesarpjz:
In general, I think your code is very well documented and neatly arranged. Your choice of method names are very clear and is very easily understood by anyone reading it. It was quite impressive that you had many smaller methods for several commands. Thanks for the good read!

PR #94, opened 2019-08-23
[Chia De Xun] Duke Increments

DivineDX:


src/main/java/logic/Ui.java
```diff
@@ -26,14 +26,20 @@ public void greet() {
...
-        sb.append("Hello! I'm logic.Duke\n");
+        sb.append("Hello! I'm Duke\n");
         sb.append("What can I do for you?");
         printStr(sb.toString());
     }
 
     //returns numbered list of tasks
-    public void printList(List<Task> taskList) {
-        StringBuilder sb = new StringBuilder("Here are the tasks in your list:\n");
+    public void printList(List<Task> taskList, boolean isMatching) {
```
Nanosync: Is there a design reason to use `isMatching` here? `printList` seems to be a general method and should not know about matching tasks.

src/main/java/Parser.java
```diff
@@ -40,4 +41,64 @@ public static String printDate(LocalDateTime dateObj) {
                 + String.format("%02d", dateObj.getMinute());
     }
 
+    public static int parseTaskInt(String str, List<Task> taskList) throws DukeException {
+        int taskInt;
+        try {
+            taskInt = Integer.parseInt(str);
+            if (taskInt > taskList.size() || taskInt <= 0) {
+                throw new DukeException("That task number does not exist, please try again");
```
Nanosync: Throwing a new `NumberFormatException` here can help reduce the number of code paths to maintain.

src/main/java/Duke.java
```diff
@@ -111,22 +117,49 @@ void executeCommand(String command, String[] inputSplit) throws DukeException {
...
         }
     }
 
+    private void loadFile(String filePath) {
+        try {
+            File f = new File(filePath); //hardCoded file directory
+            Scanner sc = new Scanner(f);
+            while (sc.hasNext()) {
+                String line = sc.nextLine();
+                String[] strArr = line.split(Pattern.quote(" | "));
```
Nanosync: You can escape the line using `\\|` to avoid using Pattern.quote.

src/main/java/logic/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            Ui.printStr("Data is loaded from data/taskList.txt");
+        } catch (FileNotFoundException e) {
+            Ui.printStr("File not found, data not loaded!\nCreating new file taskList.txt");
+            File f = new File(filePath);
+
+            try {
+                f.createNewFile(); //creates new text file
+            } catch (IOException ioE) {
```
Nanosync, line 55: Might help to name the variable `e` as a standard way to read exceptions.

Nanosync (commented):
Overall, the code looks fine, though some improvements can be made!

src/main/java/logic/Parser.java
```diff
@@ -0,0 +1,144 @@
...
+            String[] date = strSplit[0].split("/");
+            String time = strSplit[1];
+
+            int year = Integer.parseInt(date[2]);
+            int month = Integer.parseInt(date[1]);
+            int day = Integer.parseInt(date[0]);
+            int hours = Integer.parseInt(time.substring(0, 2));
+            int mins = Integer.parseInt(time.substring(2));
+
+            return LocalDateTime.of(year, month, day, hours, mins);
```
dingyuchen, line 46: May consider using a built-in formatter
https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

src/main/java/logic/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+        sb.append("  " + t + "\n");
+        taskList.remove(taskNum - 1);
+        sb.append("Now you have " + taskList.size() + " tasks in the list.");
+        Ui.printStr(sb.toString());
+    }
+
+    public List<Task> findTask(String taskDescription) {
+        return taskList.stream()
+                .filter(task -> task.getDescription()
+                        .contains(taskDescription)).collect(Collectors.toList());
```
dingyuchen:
```suggestion
                        .filter(task -> task.getDescription().contains(taskDescription))
                        .collect(Collectors.toList());
```

dingyuchen: having chained methods follow the same level of indentation if they are acting on the same object improves readability

src/main/java/logic/Ui.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Wraps strings for style and formatting
+     *
+     * @param input String parameter to be wrapped and printed
+     */
+    public static void printStr(String input) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("____________________________________________________________\n");
+        sb.append(input + "\n");
+        sb.append("____________________________________________________________\n");
```
dingyuchen: you can set line as a variable to avoid re-initializing

src/main/java/task/ToDo.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder("[T]");
+        if (isDone) {
+            sb.append("[✓] ");
+        } else {
+            sb.append("[✗] ");
+        }
```
dingyuchen, line 27: a ternary statement here might be more readable

dingyuchen (commented):
I think the code is generally good, but some parts where code is copy pasted can be abstracted to look more succinct.
I can see how you decided to package the files, but `logic` might not be the best naming convention seeing how it contains `Ui` as well as `Storage`. 
config files are usually not pushed into remote. you might want to include the whole `.idea` folder in your gitignore.

src/main/java/Parser.java
DivineDX: I coded it this way as I felt there was a difference between the one cited above (where the input number is valid, but it doesnt correspond to a valid task number) whereas when a NumberFormatException is thrown, it is because something other than an integer is typed and parsed

src/main/java/Duke.java
DivineDX: Thanks, but I tried  `line.split("\\ | ");` and it does not seem to work due to the spacing. I could apply a `trim()` to all the elements in the array, but that would be an unnecessary addition of code.

src/main/java/logic/Storage.java
DivineDX: Did not name e due to the variable e in the parent scope

src/main/java/task/ToDo.java
DivineDX: Appreciate it, but it would require further refactoring as in this instance as I am not returning a value.

dingyuchen: try `sb.append( isDone ? "[✓] " : "[✗] " );`
I think that this should work

PR #96, opened 2019-08-23
[TSAI, HSIAO-HAN] Duke Increments

TSAI-HSIAO-HAN:


src/main/java/data/duke.txt
```diff
@@ -0,0 +1,6 @@
+T | 1 | read book
```
muhammadaljunied, line 1: This file should be added into your .gitignore.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,28 @@
+package duke.command;
+
+import duke.exception.DukeException;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Represents a command that Duke will operate.
+ */
```
muhammadaljunied, line 10: I think I understand why you would want to have isExit()  as part of your interface, perhaps one day there may be another command that allows the user to exit besides with the ExitCommand, though unlikely. I would remove it and have just the ExitCommand have an `boolean isExit()` method. Then all your `*Command` classes will have less boilerplate code (:

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+            char num = fullCommand.charAt(fullCommand.length() - 1);
+            int index = Character.getNumericValue(num);
+            return  new DoneCommand(index);
+        } else if (fullCommand.startsWith("delete")) {
+            char num = fullCommand.charAt(fullCommand.length() - 1);
+            int index = Character.getNumericValue(num);
+            return new DeleteCommand(index);
+        } else if (fullCommand.startsWith("todo")) {
+            if (fullCommand.trim().length() == 4) {
+                throw new DukeException(" ☹ OOPS!!! The description of a todo cannot be empty.");
```
muhammadaljunied: Store the `DukeException` String as a constant final since it is being used on Lines, 39, 46, 50, 56, 60, 68.

It will make your code a lot more readable.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+        return tasks;
+    }
+
+    /**
+     * Saves tasks to file.
+     *
+     * @param tasks The tasks to save.
+     * @throws DukeException When file does not exist.
+     */
+    public void saveFile(ArrayList<Task> tasks) throws DukeException {
```
muhammadaljunied, line 90: A few things I noticed here:

1. Method Argument could use the more generic `List<Task>` rather than a specific class `ArrayList<Task>`. This modification helps the method to accept more subclasses of `List<T>` as arguments.

2. If program terminates (assume an unhandled exception) without ExitCommand.java::execute() being called, my changes would not be saved. An easy fix would be to add the following to your Duke.java:run() try-catch block:

```
try {
...
finally {
    ui.showLine();
    database.update(taskList);
}
```
3. Should my list size grow to the thousands (though unlikely), the list would take longer to write to file.

For 3. you could write an append(int index, String data) method that appends the new task into the file. Since each line denotes one task, and they are in running order, it would be easy for us to "insert" a new task. This reduces our write operation to a single line rather than n lines.

Then calling the new append(int index, String data) after each task is added to the list, would then eliminate the issues in 2 and 3! There is a tradeoff however, there will be more IO operations.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+        try {
+            FileWriter fw = new FileWriter(filePath);
+            String data = tasks.get(0).toFile();
+            if (tasks.size() > 1) {
+                for (int i = 1; i < tasks.size(); i++) {
+                    data = data + System.lineSeparator() + tasks.get(i).toFile();
+                }
+            }
+            fw.write(data);
+            fw.close();
```
muhammadaljunied, line 105: Here it would be advisable to move `fw.close()` into a `finally` block. 

```
try {
    FileWriter fw = new FileWriter(filePath);
...
} catch (IOException e) {
...
} finally {
    fw.close();
}
```

The reason for this is, suppose the `try-catch` catches an exception after `fw` has gained file access. When the exception is handled, `fw` still has access to the file in `filePath`. The next time you try to access this file, you will encounter an IOException again.

The `finally` block always executes regardless of Exceptions caught in the `try` block.

ChrisKheng:
Comments
1. I like how you pass the file Path of the data file to the Storage object when initializing it. Also, your code generally looks clean too!

2. Regarding the enum **Type** in the task package, I think it would be good to put it as a static field in the **Task** class as it is only related to **Task** instead of writing it in a separate file.

3. For the description(which includes the dates) of **Deadline** and **Event** right, I think you can make a string variable to store the ...

4. Regarding the input processing(for example extracting the description & dates) in **Parser**, I think you can abstract them away by putting them in the respective **Command** classes to make the Parser class code cleaner. e.g. You can abstract away the input processing logic for "done" and put them in the **DoneCommand** class. Then, you would only need to pass the raw input to the DoneCommand for it to process.

5. In **Storage** class, maybe you could make the class to automatically create the data file for the user if it hasn't existed? This can make it easier for the user to start using the app. Also, maybe you could abstract away the part where you check whether the task is done and mark them as done? Cuz there's repetition and can make it quite hard to adapt to changes since you have to change all the code at once. Maybe you could put this part after the switch statement.

In both of the methods, maybe you could directly operate on the TaskList object instead of an arrayList? I think this would improve the adherence to the abstraction principle and make more good use of the **TaskList** class. 

In the method **saveFile**, maybe you could refactor the code to remove the if statement and statement before it such that only the for loop is needed to concatenate the information of all the tasks? It may be good too to use a **StringBuilder** to build the concatenated descriptions to optimize the performance since String is immutable. Also in the error message, maybe you could make it as "Data file not found!" instead? Cuz this would make it clearer for the user to know what file is missing.

6. In the methods in the **TaskList** class, it may be good for you to check if the index is less than 1 in the two methods to make the check more comprehensive. Also, in doneTask(), maybe you could retrieve the task first and assign it to a variable, then mark it as done and return it? Because using 2 get() means we have to traverse the list twice which may add more complexity to the method.

In deleteTask(), maybe you could do something like "Task removeItem = tasks.remove(n - 1)" ? Cuz remove() will return the item that is removed from the list. This can improve the performance of the method.

Maybe you could also make a size() method too to return the size of the taskList instead of retrieving its array list and get its size. This way can make your code to adhere to the abstraction principle better.

7. In the **UI** class, maybe you could assign the line to be printed to a constant? Cuz this would make it easier for you to change your code if there's a change in the requirement needed. This can be done for the indentation string too.

Also, it may be good to remove the showLoadingError() method and use a throw error statement in the respective method of the **Storage** class instead. In the showDelete() method, you may also want to consider the case where there is 1 or 0 task in the list and the message printed should be "task" instead of "tasks" in the two cases.

8. In the **Command** classes, maybe you want to consider to do the writing to the data file only at the end of the program? Because the writing actually overwrites the entire file and doing so every time when there's a change to the taskList may be quite costly. I think if you remove the updating (which is saveFile()) in every Command class **except ExitCommand** will do the suggestion I mentioned.

9. Also maybe you could store similar classes such as **Parser, Storage** in one package instead of different packages? Cuz they are like the supporting components for the Duke bot. This way it may be easier to navigate through the directories and find similar classes. For the part to print out the result list (showFind() and listTasks()), you may also want to consider abstract away the printing part and put it in one method to prevent repetition. You can then call the new method in showFind() and listTasks() to help you print the result list only (without the top description like "Here are your matching...").

10. Last note: I think it may be good for you to do level 10 first before working on the suggestions here because some of the suggestions only pertain to the program before Level 10.

Overall you did a great job in keeping the code clean, but there are some improvements that you may want to work on, for example performance. All the best and cheers!

ChrisKheng (commented):
..

src/main/java/duke/command/Command.java
ChrisKheng: In the **Command** classes (including child classes), maybe you want to consider to do the writing to the data file only at the end of the program? Because the writing actually overwrite the entire file and doing so everytime when there's a change to the taskList may be quite costly. I think if you remove the updating (which is saveFile()) in every Command class **except ExitCommand** will do the suggestion I mentioned.

src/main/java/duke/task/Type.java
```diff
@@ -0,0 +1,8 @@
+package duke.task;
+
+/**
+ * The type of the task.
+ */
+public enum Type {
```
ChrisKheng, line 6: Regarding the enum **Type** in the task package, I think it would be good to put it as a static field in the **Task** class as it is only related to **Task** instead of writing it in a separate file.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public class Storage {
+    private String filePath; // The file's path.
+
+    /**
+     * Initiates the object.
+     *
+     * @param filePath The file's path.
+     */
+    public Storage(String filePath) {
```
ChrisKheng, line 29: I like how you pass the file path of the data file to the Storage object when initializing it. Also, your code generally looks clean too!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+     * @throws DukeException When input is in wrong format.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        if (fullCommand.equals("bye")) {
+            return new ExitCommand();
+        } else if (fullCommand.equals("list")) {
+            return new ListCommand();
+        } else if (fullCommand.startsWith("done")) {
+            char num = fullCommand.charAt(fullCommand.length() - 1);
+            int index = Character.getNumericValue(num);
```
ChrisKheng: Regarding the input processing(for example extracting the description & dates) in **Parser** class, I think you can abstract them away by putting them in the respective **Command** classes to make the Parser class code cleaner. e.g. You can abstract away the input processing logic for "done" and put them in the **DoneCommand** class. Then, you would only need to pass the raw input to the DoneCommand for it to process.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+     * @return Tasks loaded from the file.
+     * @throws DukeException When cannot found file or format incorrect.
+     * @throws FileNotFoundException Would not happen.
+     */
+    public ArrayList<Task> load() throws DukeException, FileNotFoundException {
+        ArrayList<Task> tasks = new ArrayList<Task>();
+
+        File f = new File(filePath);
+        if (!f.exists()) {
+            throw new DukeException("File not found.");
```
ChrisKheng, line 46: In **Storage** class, maybe you could make the class to automatically create the data file for the user if it hasn't existed? This can make it easier for the user to start using the app. Also, maybe you could abstract away the part where you check whether the task is done and mark them as done? Cuz there's repetition and can make it quite hard to adapt to changes since you have to change all the code at once. Maybe you could put this part after the switch statement.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+     * Saves tasks to file.
+     *
+     * @param tasks The tasks to save.
+     * @throws DukeException When file does not exist.
+     */
+    public void saveFile(ArrayList<Task> tasks) throws DukeException {
+        try {
+            FileWriter fw = new FileWriter(filePath);
+            String data = tasks.get(0).toFile();
+            if (tasks.size() > 1) {
```
ChrisKheng, line 99: In the method **saveFile**, maybe you could refactor the code to remove the if statement and statement before it such that only the for loop is needed to concatenate the information of all the tasks? It may be good too to use a **StringBuilder** to build the concatenated descriptions to optimize the performance since String is immutable. Also in the error message, maybe you could make it as "Data file not found!" instead? Cuz this would make it clearer for the user to know what file is missing.

Also, in both methods of the class, maybe you could directly operate on the TaskList object instead of an ArrayList? I think this would improve the adherence to the abstraction principle and make more good use of the **TaskList** class.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Deletes task from task list.
+     *
+     * @param n The order should be delete.
+     * @return The task be deleted.
+     * @throws DukeException When n > tasklist's size.
+     */
+    public Task deleteTask(int n) throws DukeException {
+        if (n > tasks.size()) {
```
ChrisKheng, line 38: In the methods in the **TaskList** class, it may be good for you to check if the index is less than 1 in the two methods to make the check more comprehensive. Also, in doneTask(), maybe you could retrieve the task first and assign it to a variable, then mark it as done and return it? Because using 2 get() means we have to traverse the list twice which may add more complexity to the method.

Maybe you could also make a size() method too to return the size of the taskList instead of retrieving its array list and get its size. This way can make your code to adhere to the abstraction principle better.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,77 @@
...
+            throw new DukeException(" ☹ OOPS!!! Do not exist that task.");
+        }
+        Task removeItem = tasks.get(n - 1);
```
ChrisKheng, line 41: Maybe you could do something like "Task removeItem = tasks.remove(n - 1)" ? Cuz remove() will return the item that is removed from the list. This can improve the performance of the method as we only need to traverse the list once in this case.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+ * The user interaction reactor.
+ */
+public class Ui {
+    private Scanner scanner = new Scanner(System.in); //The scanner scans user command.
+
+    /**
+     * Prints welcome message.
+     */
+    public void showWelcome() {
+        System.out.println("____________________________________________________________");
```
ChrisKheng, line 19: In the **UI** class, maybe you could assign the line to be printed to a constant? Cuz this would make it easier for you to change your code if there's a change in the requirement needed. This can be done for the indentation too, i.e. make a constant indentation string and append to the sentences that need indentation so that you can change the indentations altogether easily if you need

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+     * @return User's command.
+     */
+    public String readCommand() {
+        return scanner.nextLine();
+    }
+
+    /**
+     * Prints loading error message.
+     */
+    public void showLoadingError() {
```
ChrisKheng: It may be good to remove the showLoadingError() method and use a throw error statement in the respective method of the **Storage** class instead.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Prints delete message.
+     *
+     * @param task The task be deleted.
+     * @param listSize The size of the tasklist.
+     */
+    public void showDelete(Task task, int listSize) {
+        System.out.println("Noted. I've removed this task:");
+        System.out.println(String.format("  %s", task.toString()));
```
ChrisKheng: You may want to consider the case where there is 1 or 0 task in the list and the message printed should be "task" instead of "tasks" in the two cases.

src/main/java/duke/parser/Parser.java
ChrisKheng: Also maybe you could store similar classes such as **Parser, Storage** in one package instead of different packages? Cuz they are like the supporting components for the Duke bot. This way it may be easier to navigate through the directories and find similar classes.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+        System.out.println(String.format("Now you have %d tasks in the list.", listSize));
+    }
+
+    /**
+     * Prints find message.
+     *
+     * @param tasklist The tasks contains keyword.
+     */
+    public void showFind(ArrayList<String> tasklist) {
```
ChrisKheng: For the part to print out the result list (showFind() and listTasks()), you may want to consider abstract away the printing part for the list and put it in one method to prevent repetition. You can then call the new methbod in showFind() and listTasks() to help you print the result list only (without the top description like "Here are your matching...").

e.g. 
```
public void showFind(ArrayList<String> tasklist) {
    System.out.println("Here are the matching tasks in your list:");
    printList(tasklist);
}
```

ChrisKheng (commented):
Last note: I think it may be good for you to do level 10 first before working on the suggestions here because some of the suggestions only pertain to the program before Level 10.

Overall you did a great job in keeping the code clean, but there are some improvements that you may want to work on, for example performance. All the best and cheers!

PR #98, opened 2019-08-23
[Wen Shufa] Duke Increments

woon17:


tiuweehan:
**Strengths**
I think it is quite helpful that you created different kinds of exceptions for the different types of errors a user might encounter, as they can then be caught and handled separately. It is also useful for debugging purposes as being able to tell what exception is thrown allows the developer to narrow down the possible root causes for a problem.

**Areas of improvement**
It would be clearer if you packaged your files into meaningful directories e.g. command, exception, rather than keeping a flat hierarchy. This will make the overall application more intuitive and easily navigable.

It would be good to provide more descriptive JavaDoc comments, including tags such as @param, @return and @exception, as well as a short imperative description of what a function/class does. This will improve readability and maintainability of your application.

SoilingRogue (commented):
I think you have done a great job, especially in following the proper coding standards, naming conventions and the abstraction levels you implemented. Below are a few pointers I noted, which may help improve the quality of code.

Firstly, perhaps you should consider creating methods in Ui class to handle all System.out.println() statements in other classes, such as FindCommand, ExitCommand etc. to make full use of your Ui class as its purpose is indeed to handle all output.

Furthermore, I like the idea of creating additional DukeException classes, however they may be wasted in a sense that you have to specifically throw them for it to be caught and that they are identical to one another, which is no different from throwing a DukeException directly. Instead, consider catching the viable exceptions such as InputMismatchException, IndexOutOfBoundsException, IOException etc. and throw DukeException with the relevant error messages in them instead. Below is the sample code to illustrate my idea.
eg. 
try {
    ... code here...
} catch (InputMismatchException e) {
    throw new DukeException("Oh no please enter the correct number of elements");
} catch (NoSuchElementException e) {
    throw new DukeException("Please enter the correct element");
} finally {
    ....code here...
}

Next, instead of returning the task list in TaskList in the ListCommand, consider returning the combined toString() of each task in the task list instead. This will ensure that any accidental modification of the returned list will not result in the task list in TaskList being modified.

Lastly, consider adding more classes to the AddCommand class. Since you have already parsed the commands in the Parser, ie "deadline", "event" and "todo", you might be better off calling their relevant Command classes to build the relevant Task classes instead of needing to re-parse the command again to build them.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    /**
+     * feature of find command.
+     *
+     * @return FindCommand object.
+     */
+    private FindCommand findFeature() throws DukeException {
+        if (oneLine.length == 1 || (oneLine.length == 2 && oneLine[1].isBlank())) {
```
kangyeelim, line 42: Perhaps it might be easier for understanding to abstract the condition into another method that returns a boolean so it is more readable? Same for the other methods below

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+            return new DeleteCommand(oneLine);
+        } else if (oneLine.length == 2 && oneLine[1].trim().split(" ").length != 1) {
+            throw new ExtraDescriptionException("There is extra description for delete");
+        } else {
+            throw new InvalidNumberException("the description should be a number");
+        }
+    }
+
+    /**
+     * feature of add command.
```
kangyeelim, line 88: From reading the code, I can't really make out what it is supposed to do. Perhaps more elaboration on comments and/or additional abstraction will help

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+     * @return The Command obj by the parse function.
+     */
+    Command parse(String input) throws DukeException {
+        assert input != null : "input should be empty";
+
+        Command outCommand = null;
+        String cmd = input;
+        oneLine = cmd.trim().split(" ", 2);
+        String firstWord = oneLine[0];
+        if (firstWord.equals("exit")) {
```
kangyeelim, line 114: Perhaps consider using switch cases?

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,95 @@
...
+            DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+            return true;
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws Exception {
+        assert tasks != null;
```
kangyeelim, line 46: Perhaps you can consider adding a message so that its easier to locate where the assert fails?

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,95 @@
...
+     * check if it is a valid time format.
+     *
+     * @param str The input should be checked.
+     * @return true if str can be converted to a valid time format, otherwise returns false.
+     */
+    public static boolean isValidTime(String str) {
+        assert str != null;
+
+        try {
+            DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
```
kangyeelim: The method argument `str` is not being used in the method, perhaps you may have missed a line of code?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return all the matched task in String format, including the task position in txt file.
+     */
+    private String getTasks() throws TaskNotExistException {
+        int i = 0;
+        if (myList.size() != 0) {
+            Ui.printOutput(" Here are the matching tasks in your list: ");
+            StringBuilder temp = new StringBuilder();
+            for (Task tk : myList) {
+                i++;
+                temp.append(Ui.frontSpace).append(" ").append(i).append(". ")
```
kangyeelim: Consider abstracting the method that handles string building so the code has less indent levels.

kangyeelim (commented):
Good job overall! Some minor changes can be made :)

PR #100, opened 2019-08-23
[Yu Chenghui] Duke Increments

Seris370:


src/main/java/Ui.java
```diff
@@ -0,0 +1,183 @@
...
+    private String logo = " ____        _        \n"
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___|\n";
+
+    private void formatPrint(String[] lists) {
+        System.out.println(line);
+        for (String s : lists) {
+            System.out.println("     " + s);
```
shawnlsj97: you could use your format method here to return the string you want

src/main/java/Ui.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        System.out.println(line);
+    }
+
+    private void formatPrint(String s) {
+        System.out.println(line);
+        System.out.println("     " + s);
```
shawnlsj97: you could use your format method here to return the string you want

src/main/java/Command.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * Represents an abstract command.
+ */
+public abstract class Command {
+
```
shawnlsj97: empty class body, in that case its probably not necessary to create it

src/main/java/Bye.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * Represents an exit command.
+ */
+public class Bye extends Command {
+
```
shawnlsj97: empty class body, in that case its probably not necessary to create it

src/main/java/Duke.java
```diff
@@ -1,10 +1,103 @@
...
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    private void done(int num) {
```
shawnlsj97: i feel like all these methods (done, delete, add) should be moved into the respective command classes instead

src/main/java/Duke.java
```diff
@@ -1,10 +1,103 @@
...
+        try{
+            tasks.add(type, description);
+            storage.write(tasks);
+            ui.add(tasks);
+        } catch (DukeException ex) {
+            ui.showDukeException(ex);
+        }
+    }
+
+    private void run() {
```
shawnlsj97: could consider implementing the example run() given on the module website to make the run class more easy to read, especially since it contains the main logic of the entire program

src/main/java/Parser.java
```diff
@@ -0,0 +1,64 @@
...
+                        throw new DukeException("☹ OOPS!!! The description of an event cannot be empty.");
+                    case "deadline":
+                        throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
+                    default:
+                        throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                }
+            }
+            String type = input.substring(0, input.indexOf(" "));
+            String description = input.substring(input.indexOf(" ") + 1);
+            switch (type) {
```
shawnlsj97: the switch statement does not follow the proper coding style

src/main/java/PrintList.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * Represents a command to print all existing tasks.
+ */
+public class PrintList extends Command {
+
```
shawnlsj97: empty class body, in that case its probably not necessary to create it

src/main/java/Task.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param description content of a task
+     * @param isDone mark if the task is already done
+     */
+    public Task(String description, boolean isDone) {
+        this.description = description;
+        this.isDone = isDone;
+    }
+
+    private Date parseDate(String input) throws ParseException {
+        SimpleDateFormat ft = new SimpleDateFormat ("dd/MM/yyyy HHmm");
```
shawnlsj97: good idea to use this class

src/main/java/TaskList.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public void delete(int num) throws DukeException {
+        if (num >= tasks.size() || num < 1) {
+            throw new DukeException("Task number out of range.");
+        }
+        tasks.remove(num - 1);
+    }
+
+    /**
+<<<<<<< HEAD
```
shawnlsj97: merge conflict that you have not resolved

src/main/java/Todo.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public String repr() {
+        return "[T][" + super.getStatusIcon() + "] " + description;
+    }
+
+    /**
+     * Expresses that a todo task is done.
+     * @return string representation of a done task
+     */
+    public String done() {
```
shawnlsj97: repr method should already be able to distinguish whether task is done or not, can consider merging this method with repr

src/main/java/Ui.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param num number of task to be deleted
+     * @throws ParseException if description of task connot be parsed
+     */
+    public void delete(TaskList list, int num) throws ParseException {
+        ArrayList<Task> tasks = list.tasks;
+        Task t = tasks.get(num - 1);
+        System.out.println(line);
+        System.out.println(format("Noted. I've removed this task: "));
+        System.out.println("  " + format(t.repr()));
+        switch (tasks.size()) {
```
shawnlsj97: wrong coding style for switch statement

shawnlsj97 (commented):
Javadoc is missing for some classes, There are also empty classes so you could either consider putting some methods in this classes or do away with having these classes at all. Some methods are very long so do consider abstracting some of the if else cases to a new method.

PR #104, opened 2019-08-23
[Wang Qin] Duke Increments 

wangqinNick:


src/main/java/Duke.java
```diff
@@ -19,29 +19,53 @@ public static void runUI(ArrayList<Task> tasks){
...
                     displayQuit();
                     isContinue = false;
                     break;
                 case "list":
-                    list(tasks);
-                    break;
+                    try {
+                        list(tasks);
+                    } catch (RuntimeException exList) {
+                        System.out.println("    ____________________________________________________________");
```
limerencee: The seperator line `_______` is used a lot throughout the UI, don't you think? Maybe it can be declared once as a `public static final String` and be used throughout the codebase? :)

src/main/java/Duke.java
```diff
@@ -19,29 +19,53 @@ public static void runUI(ArrayList<Task> tasks){
...
+                    }
+
                 case "done":
-                    done(tasks, (int)command[1].charAt(0));
-                    break;
+                    try{
+                        done(tasks, Integer.parseInt(command[1])-1);
+                    } catch (IndexOutOfBoundsException exDone) {
+                        System.out.println("    ____________________________________________________________");
+                        System.out.println("     ☹ OOPS!!! There is no task "+Integer.parseInt(command[1])+" in the list.");
```
limerencee: Coding style: Introduce spacing when concatenating codes together. E.g.: `task " + Integer.parseInt(command[1]) + " in`

src/main/java/Duke.java
```diff
@@ -19,29 +19,53 @@ public static void runUI(ArrayList<Task> tasks){
...
+                    } finally {
+                        break;
+                    }
+
                 case "done":
-                    done(tasks, (int)command[1].charAt(0));
-                    break;
+                    try{
+                        done(tasks, Integer.parseInt(command[1])-1);
+                    } catch (IndexOutOfBoundsException exDone) {
```
limerencee: Good that you have surrounded this possibly-sensitive case in a `try-catch` block, but you should catch `NumberFormatException` as well. The user might not enter a numerical value as the argument.

src/main/java/Duke.java
```diff
@@ -19,29 +19,53 @@ public static void runUI(ArrayList<Task> tasks){
...
+                    } finally {
+                        break;
+                    }
+
                 case "delete":
-                    deleteTask(tasks, (int)command[1].charAt(0));
-                    break;
+                    try{
+                        deleteTask(tasks, Integer.parseInt(command[1])-1);
+                    } catch (IndexOutOfBoundsException exDelete) {
```
limerencee: Same thing, catch `NumberFormatException` as the user might not enter a numerical value.

src/main/java/Duke.java
```diff
@@ -52,16 +76,16 @@ public static void runUI(ArrayList<Task> tasks){
...
                         task = new Deadline (date[0],date[1].substring(3));
                         addTaskIn(task,tasks);
-                    } catch (StringIndexOutOfBoundsException ex2){
+                    } catch (StringIndexOutOfBoundsException exDeadline1){
                         System.out.println("    ____________________________________________________________");
                         System.out.println("     ☹ OOPS!!! The description of a deadline cannot be empty.");
                         System.out.println("    ____________________________________________________________");
 
-                    } catch (IndexOutOfBoundsException ex3){
+                    } catch (IndexOutOfBoundsException exDeadline2){
```
limerencee: I like how you handled the different missing arguments. You could rename the exception variable name to be more meaningful, or simply use `e`/`ex`.

src/main/java/Duke.java
```diff
@@ -131,6 +160,9 @@ public static void addTaskIn(Task task,ArrayList<Task> tasks) {
     }
 
     public static void list (ArrayList<Task> tasks){
+        if(tasks.size() == 0){
```
limerencee: Coding style: Add spacing between elements. E.g.: `if (tasks.size() == 0) {`

src/main/java/Duke.java
```diff
@@ -131,6 +160,9 @@ public static void addTaskIn(Task task,ArrayList<Task> tasks) {
...
+            throw new RuntimeException();
+        }
         System.out.println("    ____________________________________________________________");
         for (int i = 0; i < tasks.size(); i++) {
             System.out.print("     "+(i+1)+"."+"["+tasks.get(i).getType()+"]"+"["+ tasks.get(i).getStatus()+"] "+tasks.get(i).getTaskName()+" ");
```
limerencee: A very long line, consider breaking into 2 lines or more. The concatenation seem a bit excessive as well. Consider combining the hard-coded parts together instead of concatenating. E.g.: `"     " + (i + 1) + ".["`

src/main/java/Duke.java
```diff
@@ -52,16 +76,16 @@ public static void runUI(ArrayList<Task> tasks){
 
                 case "deadline":
                     try {
-                        taskName = taskName.substring(9);
+                        taskName = taskName.trim().substring(9);
                         date = taskName.split("/");
                         task = new Deadline (date[0],date[1].substring(3));
                         addTaskIn(task,tasks);
-                    } catch (StringIndexOutOfBoundsException ex2){
+                    } catch (StringIndexOutOfBoundsException exDeadline1){
```
teoha: IndexOutOfBoundsException may not necessarily mean the description is empty. It can also mean the date is empty. Maybe you can add checks for the fields to throw an appropriate error

src/main/java/Duke.java
```diff
@@ -19,29 +19,53 @@ public static void runUI(ArrayList<Task> tasks){
...
+                        System.out.println("     ☹ OOPS!!! There is no task "+Integer.parseInt(command[1])+" in the list.");
+                        System.out.println("    ____________________________________________________________");
+                    } finally {
+                        break;
+                    }
+
                 case "todo":
                     try {
-                        taskName = taskName.substring(5);
+                        taskName = taskName.trim().substring(5);         //it is very important to trim the space at the end
```
teoha: Instead of hard coding the starting index of the substring, using String.length() may be more effective and modifiable if the specs were to change.

src/main/java/Duke.java
```diff
@@ -122,6 +146,11 @@ public static void displayQuit(){
 
     //Add new task
     public static void addTaskIn(Task task,ArrayList<Task> tasks) {
+
+        if(task == null ){
```
teoha: StringIndexOutOfBoundsException() seems to be forced here even though it has never occurred. Maybe throwing another type of exception e.g. MissingTaskException() will be more appropriate

PR #106, opened 2019-08-23
[Ho Hol Yin] Duke Increments

hoholyin:


src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+    public String deleteItem(int index) throws DukeException {
+        if (index <= 0 || index > count) {
+            throw new DukeException("Invalid task number!");
+        }
+        index--;
+        Task t = taskList.get(index);
+        taskList.remove(index);
+        String result = "Noted. I've removed this task:\n";
+        result += t + "\n";
+        count -= 1;
```
MichaelaTSH: Why not change the index numbers after deleting to match their current position in the taskList?

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,22 @@
+package duke.util;
+/**
+ * Ui class to print display items
+ */
+public class Ui {
```
MichaelaTSH, line 6: Would you consider delegating more of the display items, such as adding and list messages, to the Ui?

MichaelaTSH (commented):
Overall, I found your code to be very consistent and clean, and easy to follow! Great job!

src/main/java/duke/util/TaskList.java
hoholyin: I have added a new method to re-index the task in the event of deletion. Thanks for the suggestion!

src/main/java/duke/util/Ui.java
hoholyin: I will look into this. Previously, the changes to be made to the UI class to return strings instead of printing to the console made it a little confusing. However, this is indeed a great suggestion. Thanks for the input!

src/main/java/duke/note/Note.java
```diff
@@ -0,0 +1,33 @@
+package duke.note;
+
+/**
+ * Note Class.
```
limhawjia: Can be more descriptive with your javadoc comments!

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,52 @@
...
+        return date;
+    }
+
+    /**
+     * Returns string representation of Deadline object.
+     * @return String representation of Deadline object
+     */
+    @Override
+    public String toString() {
+        String result = "[D][";
```
limhawjia, line 50: Since a large part of this code is repeated in other tasks, can consider abstracting it to the parent class.

src/main/java/duke/util/Command.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final int BYE = 0;
+    private static final int LIST = 1;
+    private static final int DONE = 2;
+    private static final int DELETE_TASK = 3;
+    private static final int FIND = 4;
+    private static final int TASK = 5;
+    private static final int NOTE = 6;
+    private static final int DELETE_NOTE = 7;
+    private static final int COMMANDS = -1;
+    private static final String list =
```
limhawjia: Purpose of this is not clear. Can rename to something more descriptive.

src/main/java/duke/util/Command.java
```diff
@@ -0,0 +1,88 @@
...
+public class Command {
+    private static final int BYE = 0;
+    private static final int LIST = 1;
+    private static final int DONE = 2;
+    private static final int DELETE_TASK = 3;
+    private static final int FIND = 4;
+    private static final int TASK = 5;
+    private static final int NOTE = 6;
+    private static final int DELETE_NOTE = 7;
+    private static final int COMMANDS = -1;
```
limhawjia, line 17: Can consider using an enum instead to reduce class bloat.

src/main/java/duke/util/Command.java
```diff
@@ -0,0 +1,88 @@
...
+        case TASK:
+            result = lists.getTasks().registerNewTask(inputParts);
+            storage.saveHistory(lists);
+            return result;
+        case NOTE:
+            result = lists.getNotes().registerNewNote(inputParts[1]);
+            return result;
+        case COMMANDS:
+            return list;
+        default:
```
limhawjia, line 86: Can consider using inheritance to represent the different commands.

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,146 @@
...
+        for (String line : lines) {
+            if (line.equals(LINE_DIVIDER)) {
+                readingTasks = false;
+                continue;
+            }
+            if (readingTasks) {
+                String[] parts = line.split("\\|");
+                String part = parts[0];
+                switch (part) {
+                case "T":
```
limhawjia, line 67: Can add comments to explain what does each part mean.

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,304 @@
...
+            result += "Now you have 1 task in the list";
+        } else {
+            result += "Now you have " + count + " tasks in the list";
+        }
+        return result;
+    }
+
+    /**
+     * Creates a new task from a given input.
+     * @param inputParts An array of <code>String</code> split into type of task, name of task and date (if required)
```
limhawjia, line 111: Cool. I didn't know you can do this

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,304 @@
...
+        return echo(t);
+    }
+
+    /**
+     * Checks if a given command is valid.
+     * @param type Type of command
+     * @throws DukeException if command is invalid
+     */
+    static void checkCommandValidity(String type) throws DukeException {
+        if (!type.equals("todo") && !type.equals("deadline") && !type.equals("event")) {
```
limhawjia, line 127: This may be hard to extend in the future.

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,304 @@
...
+     * @param type type of task
+     * @return Task object to be appended to the tasklist
+     * @throws DukeException if description of task is empty or if format is incorrect
+     */
+    public Task addToList(String s, String type) throws DukeException {
+        String trimmed = s.replaceAll("^\\s+", "");
+        if (trimmed.equals("")) {
+            throw new DukeException("Description cannot be empty!");
+        }
+        if (type.equals("todo")) {
```
limhawjia: Can consider refactoring some of the if blocks to new methods to improve readability

limhawjia (commented):
Good to see that you were able to implement many optional extensions successfully. For the most part, your code was readable, but there are segments where you can consider using SLAP.

src/main/java/duke/note/Note.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    public Note(String desc) {
+        this.desc = desc;
+    }
+
+    /**
+     * Returns the description of a note.
+     * @return description of the note
+     */
+    public String getDesc() {
```
MackyMaguire, line 23: What's the difference between the description and string representation of a note?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,57 @@
+package duke.task;
+
+/**
+ * Task class.
+ */
+public abstract class Task {
+    protected boolean completed;
```
MackyMaguire: I think isCompleted is a better name for a boolean?

src/main/java/duke/util/Duke.java
```diff
@@ -0,0 +1,68 @@
...
+        return "";
+    }
+
+    /**
+     * Returns the message to be displayed when an invalid command is received.
+     * @param errorMessage Description of error message
+     * @return String of entire error message
+     */
+    public String invalidCommandMessage(String errorMessage) {
+        return errorMessage + "\nType 'commands' to view a list of commands you can use";
```
MackyMaguire, line 68: I like this :)

src/main/java/duke/util/Command.java
```diff
@@ -0,0 +1,88 @@
...
+            return result;
+        case DELETE_NOTE:
+            result = lists.getNotes().deleteItem(Integer.parseInt(inputParts[1].split(" ")[1]));
+            storage.saveHistory(lists);
+            return result;
+        case FIND:
+            return lists.getTasks().findItem(inputParts[1]);
+        case TASK:
+            result = lists.getTasks().registerNewTask(inputParts);
+            storage.saveHistory(lists);
```
MackyMaguire, line 79: Might be better to just append the new task to the file instead of rewriting the entire tasklist to the file if you are only adding one task.

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,304 @@
...
+            } else if (parts.length != 2) {
+                String message = "Format: deadline {task_name} /by {date}";
+                throw new DukeException(message);
+            }
+            taskList.add(new Deadline(
+                    parts[0].substring(0, parts[0].length() - 1),
+                    createDateAndTime(parts[1].substring(1))
+            ));
+        } else if (type.equals("event")) {
+            String[] parts = s.split("\\/" + "at");
```
MackyMaguire: Should this job of splitting be left to the parser instead of tasklist?

src/main/java/duke/util/TaskList.java
MackyMaguire: Should all these differentiating types of tasks and splitting be left to the parser instead?

MackyMaguire (commented):
Really nice code that is easy to read and follow. Also good that you've added javadocs for all your public methods. Well Done!

src/main/java/duke/note/Note.java
hoholyin: Will do!

src/main/java/duke/task/Deadline.java
hoholyin: Great suggestion!

src/main/java/duke/util/Command.java
hoholyin: Will do!

src/main/java/duke/util/Command.java
hoholyin: I do not think this causes class bloat but it is indeed a reasonable suggestion to use enums in place of this.

src/main/java/duke/util/Storage.java
hoholyin: Will abstract for code to be more readable

src/main/java/duke/util/TaskList.java
hoholyin: May I inquire why it will be hard? Since we abstracted out the checking of command validity from the higher-level method, we can just simply modify this method should anything change in the future without changing the main code, making it open for modifications.

src/main/java/duke/util/TaskList.java
hoholyin: In progress of doing so! Thanks for spotting it.

src/main/java/duke/note/Note.java
hoholyin: `String` representation is when we are returning duke's response, while getDesc() is to get the description of the note when saving it into the storage. My intention was that although both return the same thing, but since they serve different purposes and will be called under different circumstances, it would help to have different descriptive names describing the context when these methods are being called. However, you have a good point and I might consider changing their names to make them sound suitable for all contexts.

src/main/java/duke/util/Duke.java
hoholyin: Thanks!

src/main/java/duke/util/TaskList.java
hoholyin: Considering that the `Parser` object is a higher-level object, my understanding is that it should not meddle with the lower-level details like checking if a task is valid or not. Consider that I have implemented the extension for `Notes`. I would prefer to let their respective lists verify their validity rather than having a general all-purpose Parser exposed to the lower level details.

src/main/java/duke/util/Command.java
hoholyin: Personally, I felt that this suggestion to use different `Command` objects (which is shown as a guide in the module website) is an unnecessary level of abstraction which dives too deep. Rather, in my opinion, it will be appropriate to only create command objects that inherit from a base command class if we are required to implement a few other instance methods `Command`s can use or if we want to make use of the inheritance property for perhaps storing them in a list or use some common methods. I believe that we should not blindly use inheritance just because we want to make use of it as it must add value to the code.

src/main/java/duke/task/Task.java
hoholyin: You're right! I have amended it.

src/main/java/duke/util/Command.java
hoholyin: I will consider this!

PR #107, opened 2019-08-23
[Lee Liak Ghee] Duke increments

liakify:


jietung:
I like that you extracted out all the command and makes the code looks clean. However, i think you should do some testing to avoid regression.

jonathantjendana:
I like that you make different classes for the different exceptions and having clear commit messages each time you push. However, I think you could use enumerations for your commands instead of having one class for each commands.

PR #108, opened 2019-08-23
[Wallace Lim] Duke Increments

wallacelim:
Added redundant file


src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        }
+    }
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        new Duke("C:\\Users\\Admin\\OneDrive - National University of Singapore\\School\\Y2S1\\CS2103T\\duke\\src\\main\\java\\data\\tasks.txt").run();
```
daekoon: I would recommend using a relative path for the save file path. Absolute path like this means it won't work outside of your computer :sweat_smile:

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,63 @@
+package duke.util;
+
+import duke.command.*;
```
daekoon, line 3: Try not to have wildcard imports in your program :)

daekoon (approved):
Looks good to me :)

PR #109, opened 2019-08-23
[Chen Hui] Duke Increments

chrischenhui:


src/main/java/Duke.java
```diff
@@ -1,10 +1,70 @@
...
+        }
+    }
+
+    /**
+     * Processes duke after set up.
+     */
+    public void run() {
+        ui.showWelcome(tasks.tasks);
+        /*
+        boolean isExit = false;
```
sreesubbash: Maybe delete code which would not be used or referred to in future.

src/main/java/Duke.java
```diff
@@ -1,10 +1,70 @@
...
+    /**
+     * Returns a message generated based on user input.
+     *
+     * @param input User input.
+     * @return Message in a string.
+     */
+    String getResponse(String input) {
+        String doggoResponse = "";
+        try {
+            doggoResponse = Parser.parse(input).execute(tasks, ui, storage);
```
sreesubbash: Its a very unique and funny name you are using to replace Duke :)

src/main/java/Main.java
```diff
@@ -0,0 +1,29 @@
...
+            AnchorPane ap = fxmlLoader.load();
+            Scene scene = new Scene(ap);
+            stage.setScene(scene);
+            fxmlLoader.<MainWindow>getController().setDuke(duke);
+            stage.show();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+}
```
sreesubbash, line 29: I believe reading somewhere that it is good practice to end each file with a new line. Think its useful for git.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,34 @@
...
+     * @param storage Storage to read and write files.
+     * @throws DukeException If user input a number that is not within TaskList.
+     */
+    public String execute(TaskList t, Ui ui, Storage storage) throws DukeException {
+        try {
+            Task deletedTask = t.tasks.remove(pos);
+            storage.save(t.tasks);
+            return ui.showDeletedTask(deletedTask, t.tasks.size());
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeException("Please input a number that is within the list");
```
sreesubbash, line 31: Nice that you are specifying to the user what input they are keying wrongly.

src/main/java/Duke.java
```diff
@@ -1,6 +1,9 @@
 import duke.command.Command;
 import duke.exception.DukeException;
-import duke.util.*;
+import duke.util.Parser;
+import duke.util.Storage;
+import duke.util.TaskList;
+import duke.util.Ui;
 
 public class Duke {
```
rongrongrr, line 8: Missing a header comment for Duke class? (from what I understand every public class/ method needs to have a javadoc comment).

src/main/java/duke/command/AddCommand.java
```diff
@@ -5,11 +5,21 @@
 import duke.util.Ui;
 import duke.task.Task;
 
+/**
+ * Create an AddCommand. It adds tasks to user's list.
```
rongrongrr: Comment should start with "Creates" and not "Create". The same applies to the comments for the rest of the methods.

rongrongrr (commented):
Documentation was generally well done. I like how the comments are concise and provide readers with a clear brief of the corresponding class/ method.

PR #110, opened 2019-08-23
[Yap Shi Hao] Duke Increments

shihaoyap:
Duke, Level 1-6


shihaoyap:
Duke, Week 3
Completed: Level-7, Level-8, A-MoreOOP, A-Packages, A-JUnit, A-Jar, A-JavaDoc, A-CodingStandard, Level-9, Create a PR to the upstream repo

shihaoyap:
Duke, Week 4
Completed: A-Gradle, A-CheckStyle, Level-10, A-JavaFx

cheongliyuan:
Hi Shi Hao,

I did not notice any coding standard violations. However, trying _done 1_ caused an error. 
Capture
I believe it is due to the file location being specified according to how it was placed in your pc. 
Capture2
Running your program on other computers might fail if the folder is not setup accordingly. A simple solution would be to simply not state the directory, and it will be created where the main jar is(?). Should consider changing it, as it might not work for the tester! 

Moving on, i really liked the program! Having a GUI and being able to 'talk' to the robot gives it so much creativity points haha! 
Readability of code is good too, with different levels of abstractions being observed. Perhaps considering putting the _if (inputCommand.equals("bye"))_ in _getResponse()_ to another method? 

Looking forward to seeing your program develop. 
-Li Yuan

shihaoyap:
Thank you @cheongliyuan for your comments. The points that you have made are really helpful! I have make the necessary changes and updates to the code! Glad to hear that you have enjoyed the program! :)

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Sets Duke object to allow MainWindow.java to have a duke object to use
+     *
+     * @param d takes in a duke object and pass it for MainWindow to use
+     */
+    public void setDuke(Duke d) {
+        duke = d;
+        dialogContainer.getChildren().addAll(
+                DialogBox.getDukeDialog(duke.getStartMessage(), dukeImage));
```
CarbonGrid, line 47: Good effort for displaying the Start Message when duke is being initialized.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
+    private Ui ui;
+
+    /**
+     * Constructor for Duke class
+     * instantiate a UI, Storage and TaskList object.
+     *
+     */
+    public Duke() {
+        ui = new Ui();
+        storage = new Storage("D:\\NUS Computer Science\\CS2103T\\duke\\src\\main\\java\\duke.txt");
```
CarbonGrid: Prefer relative path over full path. Full path are hardly usable in deployments.
```suggestion
        storage = new Storage("duke.txt");
```

src/main/java/Storage.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Represents a Storage Class which is used to handle all
+ * file handling task - writing, deleting, updating.
+ */
+
+public class Storage {
+
+    private File taskFile;
+    private String filePath;
```
CarbonGrid: Either taskFile or filePath should be used instead of "duke.txt" in Storage methods.

CarbonGrid (commented):
Good evening @shihaoyap ,

I think your coding standard is on the mark since checkStyle is not throwing any errors. Most expected functionalities are available but rather restrictive.
For example:
" done 1" doesn't work.
"done -1" sliently fails and doesn't clear userInput.
& Some GUI problems with restrictive parsing.
Capture

Some key pointers:
1. Ui has too many repetitions of System.out.print("_____");
2. Ui is doing both Parser's and Storage's jobs.
Try to Let Ui work only on I/O. Storage on Files. and Parser on Parsing.
3. Command is preferably helpful to tackle Pointer 2 as per iP project guide.
4. Duke doesn't load or auto-load save data as per requirement in Level-7.
5. More JUnit tests could be implemented. e.g fixedTests() & runTest()
6. (It's just my preference) As per DukeException, when you have too many "if else" chained, opt for a switch case.

Anyways, good effort for the originality.
Keep up the effort and learning experience.

shihaoyap:
Duke, Week 4
Completed: Improved OOP Structure, Added Command Classes, Added Packages, A-Gradle, A-Assertions, A-CodeQuality, A-Travis, C-DetectDuplicates, C-BetterSearch

C-DetectDuplicates 
Program can detect if the user tries to input another event with the same description. Program will prompt user to use a unique description and will not allow him to use the same description.

 C-BetterSearch
Program can now search for all task that contains certain keywords, even though users only search partial descriptions.

PR #111, opened 2019-08-23
[Dalson Tan] Duke Increments

dalsontws:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        for (String pattern : dateFormats) {
+            SimpleDateFormat dateTime = new SimpleDateFormat(pattern);
+            try {
+                this.dateTime = dateTime.parse(by);
+            } catch (ParseException e) {
+                System.out.println("Unable to convert String to SimpleDateFormat.");
```
shaoyi1997: Great that your program accounts for multiple datatime format! 
A point for improvement: Over here, if the input datetime does not match the first format in the dateFormats list, the console will log the error message. The program will then match the input datetime with the second format in the list, which hopefully matches. Notice here that even though the input datetime matches with one of your given format, the user still receives an error message that the program is "Unable to convert String to SimpleDateFormat." Perhaps find a way to only log the message once all datetime formats are checked through?

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+        Parser parser = new Parser();
+
+        // Create new user interface object
+        Ui ui = new Ui();
+
+        // Initial opening introduction and prompt for user input
+        ui.openingStatement();
+
+        // Parse Duke commands
+        parser.parseDuke(storage, tasks, input);
```
wilfredbtan: Is there a reason why you didn't create a separate Command Class? Would separating the parsing and execution of commands make more sense?

src/main/java/Duke.java
```diff
@@ -1,10 +1,235 @@
...
+            } else if (input.contains("delete")) {
+
+                try {
+                    //Mark task as done
+                    tasks.deleteTask(input);
+                } catch (NullPointerException err) {
+                    ui.invalidEntry();
+                    continue;
+                }
+            } else if (input.contains("todo") || input.contains("deadline") || input.contains("event")) {
```
shaoyi1997: Hmmm if the description of a task contains any of these keywords, wouldn't the task fall into the wrong category?
eg. deadline complete todo list /by 12/12/2020 0800; here the task will fulfill the `input.contains("todo")` condition and go through the todo handler.

dalsontws: Ah yes, thanks for the good observation! Changed it to checking the first string for the command instead!

PR #112, opened 2019-08-23
[Koh Yi Da] Duke Increments

kohyida1997:


fabbbbbbyy (commented):
Your code is clean and simple to read. One suggestion would be to abstract out long blocks of code from the public methods into private methods.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+    public void setChangedTrue() {
+        storageIsChanged = true;
+    }
+
+    /**
+     * Returns a true or false indicating whether storage has been updated.
+     *
+     * @return Status of storage.
+     */
+    public boolean storageUpdated() {
```
tysg: Will it be better to name the method to `isUpdated`?

src/main/java/duke/datetime/Timing.java
```diff
@@ -0,0 +1,52 @@
+package duke.datetime;
+
+/**
+ * Represents a timing in the 24-hour clock format.
+ */
+public class Timing {
```
tysg, line 6: Love the implementation here!

src/main/java/duke/datetime/Date.java
```diff
@@ -0,0 +1,59 @@
+package duke.datetime;
+
+/**
+ * Represents a date in the format Day/Month/Year.
+ */
+public class Date {
```
tysg, line 9: I'm not sure if you have already considered this, but would it be better to use the built-in Java Date class?

tysg (commented):
Hi Yi Da, the code LGTM. Very well composed and cleanly written!

PR #115, opened 2019-08-23
[Tuan Ding Wei] Duke Increments

tuandingwei:


src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+                    ui.showListOfTask(tasks);
+                    continue;
+                }
+
+                Parser parser = new Parser(input);
+                String userCommand = parser.getUserCommand();
+                String due = parser.getDue();
+                String taskDescription = parser.getTaskDescription();
+
+                if (userCommand.equals("done")) {
```
lumos309: It might be better to use a `switch` here instead of a series of `if-else`s. [As of JDK 7](https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html), strings can be used as the expression in a switch statement as Java compares it to the cases using the `.equals()` method instead of using identity.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+        String filePath = "/Users/TuanDingWei/Desktop/NUS_Academia" + "/CS2103/Individual_project/Duke/local/Tasks.txt";
+
+        Scanner sc = new Scanner(System.in);
+        String input;
+        String check = "dummy";
+        int taskCount;
+
+        while (check.equals("bye") == false) {
+            try {
+                input = sc.nextLine();
```
lumos309: Although it doesn't make much difference to the length of the code, for the purpose of standardisation it might be better to handle user input with the `Ui` class instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+                check = input.toLowerCase();
+                if (check.equals("list")) {
```
lumos309: "list" is a `userCommand` just like the rest of the possible commands. Why not check for it along with the other commands? It would make the logical flow a little better.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+                String due = parser.getDue();
+                String taskDescription = parser.getTaskDescription();
+
+                if (userCommand.equals("done")) {
+                    int target = Integer.valueOf(taskDescription);
+                    Task taskDone;
+                    if (tasks.size() >= target && target > 0) {
+                        taskDone = tasks.get(target - 1);
+                    } else {
+                        throw new IndexDoesNotExistException(taskDescription + " is out of the list.");
```
lumos309: Consider formatting this message as "Index position X is out of the list." instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+                Parser parser = new Parser(input);
+                String userCommand = parser.getUserCommand();
+                String due = parser.getDue();
+                String taskDescription = parser.getTaskDescription();
+
+                if (userCommand.equals("done")) {
+                    int target = Integer.valueOf(taskDescription);
+                    Task taskDone;
+                    if (tasks.size() >= target && target > 0) {
+                        taskDone = tasks.get(target - 1);
```
lumos309: `taskDone` and `taskDelete` are names that have a verb form. `doneTask` and `deletedTask` better convey the intended meaning here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+                    } else {
+                        throw new IndexDoesNotExistException(taskDescription + " is out of the list.");
+                    }
+                    taskDone.markAsDone();
+                    storage.updateLocalFile(tasks.get());
+                    ui.doneAnnouncement(taskDone);
+                } else if (userCommand.equals("find")) {
+                    tasks.keywordSearch(taskDescription);
```
lumos309: For consistency, the output here should be handled with a `Ui` announcement just like with the other commands.

src/main/java/Duke.java
```diff
@@ -1,10 +1,184 @@
...
+
+    /**
+     * Performs the operation of creating a Task (Todo, Deadline, Task).
+     * It also writes the list of tasks to the local storage.
+     *
+     * @param userCommand     Indicates the instruction (done, delete, etc) to the bot.
+     * @param due             Provides the due date/ time of the task.
+     * @param taskDescription Reflects the description of a task.
+     * @param storage         The local storage for the Tasks as a Storage object.
+     * @param ui              The User Interface object that performs all interactions with the user.
```
lumos309: It seems a bit odd to pass a `Ui` object into this method, since it is not directly involved in the creation of a task. Why not instead have the method return the created task, to be passed into `Ui.announcement`?

src/main/java/DukeException.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public DukeException(String message) {
+        super(message);
+    }
+}
+
+/**
+ * Encapsulates exception that will be thrown when Todo description is empty.
+ */
+class EmptyToDoDescriptionException extends DukeException {
```
lumos309, line 21: Why does only the `ToDo` class have its own exception, and not `Deadline` or `Event`?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,51 @@
...
+     *                The format should follow "by: Day Time".
+     *                E.g. by: Sunday 5pm / by: 11/10/2019 3pm
+     *                Alternatively, the user has
+     *                the freedom to specify their own duration of
+     *                the event such as "by no idea :-P"
+     */
+    public Deadline(String description, Date endDate) {
+        super(description);
+        this.endDate = endDate;
+        this.typeOfTask = "D";
```
lumos309: Consider using an enum for task types instead.

src/main/java/Parser.java
```diff
@@ -0,0 +1,43 @@
+/**
```
lumos309, line 1: Nice abstraction of input parsing.

src/main/java/Storage.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Constructs the Storage object.
+     * @param filePath Provides the filepath to the local text file as the local storage.
+     */
+    Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public ArrayList<Task> load() {
```
lumos309, line 44: Missing JavaDoc.

src/main/java/Storage.java
```diff
@@ -0,0 +1,184 @@
...
+                date = format.parse(input);
+                return date;
+            }
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+        return date;
+    }
+
+    public void updateLocalFile(ArrayList<Task> taskStorage) {
```
lumos309, line 131: Missing JavaDoc.

src/main/java/Task.java
```diff
@@ -0,0 +1,96 @@
...
+ * Encapsulates the Task object.
+ * The Task object includes its own description
+ * as well as if any of the object instantiated is done or uncompleted.
+ * Also, it keep tracks of a counter of all Task objects that are instantiated.
+ * The Task object provides the extension to other children classes such as Todo, Deadline and Event.
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    String typeOfTask = "";
```
lumos309: Why is this attribute given a dummy value here, while the two attributes above are not?

src/main/java/Task.java
```diff
@@ -0,0 +1,96 @@
...
+    protected static int taskCount = 0;
```
lumos309: This attribute might be more relevant to `TaskList` than `Task`. Having it there would also allow for future implementation of multiple `TaskList` objects.

src/main/java/Ui.java
```diff
@@ -0,0 +1,73 @@
+import java.util.ArrayList;
+
+/**
+ * Encapsulates the User Interface operations. It is responsible of all interaction with the user.
+ */
+
+public class Ui {
+
+    public void welcome() {
```
lumos309: Since all the methods in this class are stateless, they could all be made `static` instead. At present there is no functionality gained from instantiating a `Ui` object.

src/main/java/Duke.java
```diff
@@ -1,10 +1,358 @@
...
+        return textToAdd;
+    }
+
+    private void run() {
+        String filePath = "/Users/TuanDingWei/Desktop/NUS_Academia" + "/CS2103/Individual_project/Duke/local/Tasks.txt";
+
+        Scanner sc = new Scanner(System.in);
+        String input;
+        String check = "dummy";
+        int taskCount;
```
AhmadHatziq: Maybe initialise the variables to an invalid value like int taskCount = INVALID_NUMBER; ?

src/main/java/Duke.java
```diff
@@ -1,10 +1,358 @@
...
+     * @param taskDescription Reflects the description of a task.
+     * @param storage         The local storage for the Tasks as a Storage object.
+     * @param ui              The User Interface object that performs all interactions with the user.
+     * @param tasks           The list of Task that are temporary stored with the bot.
+     * @return Gives a feedback to the user on the operation has performed after a command is given.
+     * @throws EmptyToDoDescriptionException Indicates an empty description that should not be left empty for Todo.
+     * @throws EmptyDescriptionException     Indicates an empty description that should not be left empty.
+     * @throws UnknownCommandException       Indicates the inability of the bot to read the command that is given by the user.
+     */
+    public static String createTask(String userCommand, String due, String taskDescription, Storage storage, Ui ui, TaskList tasks) throws EmptyToDoDescriptionException, EmptyDescriptionException, UnknownCommandException {
```
AhmadHatziq: Perhaps you can perform more OOP and break this method up. 6 arguments sounds a bit like overkill

src/main/java/Duke.java
```diff
@@ -1,10 +1,358 @@
...
+            tasks.add(t);
+            storage.writeToFile(typeOfTask, "0", taskDescription, t);
+        } else {
+            throw new UnknownCommandException("I'm sorry, but I don't know what that means :-(");
+        }
+
+        int taskCount = Task.getTaskCount();
+        return ui.newTaskAdded(t, taskCount);
     }
 }
```
AhmadHatziq, line 159: Other than that, your Duke implementation looks clean, neat and good! 

I can learn some stuff from you, such as breaking down the Error classes into more OOP. All the best. Cheers!

src/main/java/Storage.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+
+                if (isDone.equals("1")) {
+                    task.markAsDone();
+                }
+
+                taskStorage.add(task);
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("The local file cannot be located. " +
```
AhmadHatziq, line 86: Maybe if local file is not found, create the local file? What if the user is running Duke for the first time? Just a suggestion

tuandingwei: Thanks for the suggestion!

src/main/java/Task.java
tuandingwei: Yes, it is redundant. I have removed it. Thanks.

src/main/java/Storage.java
tuandingwei: Thanks.

src/main/java/Storage.java
tuandingwei: Thanks.

src/main/java/Duke.java
tuandingwei: Thanks.

PR #118, opened 2019-08-24
[Teo Jun Hong] Duke Increments

JunHongT:


ChengGeng97:
Overall I see a lot of effort put into writing Javadoc comments, which makes it very easy to understand what each method does.

If I have one criticism to make, it's that more could be done to make the project more OOP (since we're working in Java, after all). Especially since you've started implementing JavaFX, you could consider moving some functionalities out of Duke to prevent it from getting too bloated. Some possible recommendations include:
- Adding a new class to handle the saving and loading of data
- Adding a new class to handle parsing inputs
- Adding a new class to handle the String-to-Date conversions

Overall, it's a good job.

tanamaroby:
Your code works well and handles a lot of common exception which is good as it prevent the program from exiting unexpectedly. 
However, in my opinion your Duke class could be shortened by implementing other classes that functions to assist Duke so as to prevent your Duke class from becoming too big. Maybe you can create a Parser class and Reader class to handle user input. Another suggestion would be to separate the different input commands into Command classes as it will enable you to implement additional features in the future more conveniently.
Despite that, this is a very well made code and has a little something for everyone. Great use of StringBuilder to lessen the usage of System.out.println()!

savedList.txt
```diff
@@ -0,0 +1,3 @@
+[T][?] a0
+[D][?] b (by: 12/12/1212 12:12)0
+[E][?] c (at: 10/10/1010 11:12 - 12:21)0
```
dingyuchen: i think you should keep data locally. add to .gitignore file

src/main/java/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param by datetime for deadline
+     */
+    public Deadline(String description, Date by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy HH:mm");
```
dingyuchen, line 22: I think the formatter should be a static variable :) since it is always the same format

src/main/java/Duke.java
```diff
@@ -1,10 +1,174 @@
...
-        System.out.println("Hello from\n" + logo);
+        Duke execute = new Duke();
+
+        try {
+            execute.run();
+        } catch (FileNotFoundException e) {
+            System.out.println("Stored file not found!!!");
+        }
+    }
+    */
```
dingyuchen, line 45: Perhaps you might want to delete the commented code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,174 @@
...
+                reply = user.emptyError();
+            }
+        } else if (input.toLowerCase().contains("find")) {
+            String searchTerm = input.substring(5);
+            reply = user.findTask(parse.getList(), searchTerm);
+        } else {
+            reply = parse.generateTask(input);
+        }
+
+        return reply;
```
dingyuchen, line 171: Perhaps you might want to consider applying more layers of abstraction, like putting this block in an Ui class or Parser class. SLAP dat.

src/main/java/Event.java
```diff
@@ -0,0 +1,49 @@
...
+        super(description);
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public String toString() {
+        SimpleDateFormat startFormat = new SimpleDateFormat("dd/MM/yyyy HH:mm");
+        SimpleDateFormat endFormat = new SimpleDateFormat(" - HH:mm");
+        return "[E]" + super.toString() + "(at: " + startFormat.format(start) + endFormat.format(end) + ")";
```
dingyuchen, line 27: Perhaps you might want to keep these formatters as a static final variable as well, to avoid reinitializing on every `toString` call

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
+
+import java.sql.SQLOutput;
```
dingyuchen, line 2: I believe this might be an unnecessary import?

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+    public ArrayList<Task> getList() {
+        return this.items;
+    }
+
+    /**
+     * Deleting of task from ArrayList.
+     * @param num index of task in list
+     * @return String delete message
+     * @throws IndexOutOfBoundsException index of task not found in list
+     *
```
dingyuchen, line 37: improper javadocs style
use `Deletes` instead of `Deleting`,
use consistent line breaks at the end of javadoc

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * Converting String to datetime for storage.
+     * @param input String from user input from date to start time for event
+     * @return Date containing dd/MM/yyy HH:mm
+     * @throws ParseException input does not matched date format
+     */
+    public static Date convertStringToEventStart(String input) throws ParseException {
+        Date result = new SimpleDateFormat("dd/MM/yyyy HH:mm").parse(input);
```
dingyuchen, line 64: keep it DRY

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+        String message = "";
+        if (input.toLowerCase().contains("todo")) {
+            //adding an item
+            try {
+                Todo newTask = new Todo(input.substring(5));
+                if (!checkDuplicate(newTask)) {
+                    items.add(newTask);
+                    message = user.generateMessage(newTask, items.size());
+                    System.out.println(message);
+                } else {
```
dingyuchen, line 95: This might be too nested.
Avoid arrowheads.
Consider SLAPing, and make the happy path obvious.

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+
+        //store current tasks into local file for future reference
+        storeCurrent(items);
+
+        //Concluding Message
+        user.conclusion();
+    }
+    */
```
dingyuchen, line 217: bruddah .

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+
+
+    public boolean checkDuplicate(Task curr) {
+        boolean result = false;
+        for (int i = 0; i < items.size(); i++) {
+            Task test = items.get(i);
+            if (test.getClass().equals(curr.getClass())) {
```
dingyuchen, line 224: is this unnecessary work? Perhaps checking of class has already been done in the `equals` method

src/main/java/Parser.java
```diff
@@ -0,0 +1,234 @@
...
+                System.out.println("same task class");
+                if(test.equals(curr)) {
+                    result = true;
+                    break;
+                }
+            }
+        }
+        return result;
```
dingyuchen, line 232: Perhaps you can consider using Streams to filter for duplicates and to keep code readable.

src/main/java/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+                    if (Integer.parseInt(input.substring(input.length() - 1)) == 1) {
+                        newTask.markAsDone();
+                    }
+                    items.add(newTask);
+                } catch (StringIndexOutOfBoundsException | ParseException e) {
+                    user.eventError();
+                }
+            }
+        }
+    }
```
dingyuchen, line 90: more SLAPing here

src/main/java/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    public static void storeCurrent(ArrayList<Task> inputs) {
+        try {
+            File file = new File("/Users/teojunhong/JavaProject/2103T/duke/savedList.txt");
+            FileWriter fw = new FileWriter(file);
+            PrintWriter pw = new PrintWriter(fw);
+            for (Task input : inputs) {
+                int status = 0;
+                if (input.isDone) {
+                    status = 1;
```
dingyuchen, line 103: perhaps use a boolean for variable with only 2 states

dingyuchen (commented):
My team believes that you have made a good effort, however we think that more abstraction would be swell.

PR #119, opened 2019-08-24
[Eugene Lim] Duke Increments

eugenood:


madanalogy:
Hi, I really like how OOP your code is, makes it easy to understand the different components and follow along. Quite cool also how your taskList just extended off ArrayList rather than containing an ArrayList, I never thought of that, haha. I don't really have any suggestions sorry, great work! :)

eugenood:
Thanks @madanalogy. Appreciate your feedback :satisfied:

danielwys:
Good job! I admire that you've already completed the JavaFX portion of Duke so much in advance, and that you've accomplished all the levels, even the optional ones. Your codestyle and formatting is also consistent throughout. It was a pleasure to review your code.

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exception;
+
+public abstract class DukeException extends Exception {
+
+    public abstract String getResponseMessage();
```
thaowei, line 5: I feel like an exception class shouldn't be responsible for what is displayed to the user.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public String getDescription() {
+        return this.description;
+    }
+
+    public abstract String toEncodedString();
```
thaowei, line 28: Your tasks are responsible for encoding themselves but there's a separate parser class.

PR #121, opened 2019-08-24
[Tan Chee Peng] Duke Increments

t-cheepeng:


src/main/java/Duke.java
```diff
@@ -1,10 +1,59 @@
...
+    /**
+     * Gets duke's response to a user's input.
+     * @param input The user's input
+     * @return Duke's response as a string
+     */
+    public String getResponse(String input) {
+        Command command = null;
+        try {
+            command = parser.parse(input);
+            if (command instanceof ByeCommand) {
```
chowyiwai, line 47: Is this if condition necessary since you are going to call command.execute() either ways? You probably don't have to use instanceof too since polymorphism will ensure that the correct execute method corresponding to the type of command is called.

src/main/java/Duke.java
```diff
@@ -1,10 +1,59 @@
...
+                command.execute();
+            }
+            command.execute();
+            io.writeTasks(userHistory.getTaskList());
+        } catch (MissingInformationException | UnknownCommandException
```
chowyiwai, line 52: I like how you have written separate Exception classes to take care of the various possible exceptions and the naming is very clear too! 👍

src/main/java/com/chee/model/Task.java
```diff
@@ -0,0 +1,57 @@
...
+ * specific types of tasks that have their own
+ * unique attributes and methods.
+ *
+ * </p>
+ *
+ * @see ToDo Event Deadline
+ */
+public abstract class Task {
+    String description;
+    boolean isDone;
```
chowyiwai: I think there should be access modifiers for these.

src/main/java/com/chee/parser/CommandParser.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param command The user inputted command
+     * @return {@code Command} objects representing a valid command
+     * @throws UnknownCommandException     when command is not known by Duke
+     * @throws MissingInformationException when command lacks information required
+     * @throws ParseException              when dates are unable to be parsed by parser
+     */
+    public Command parse(String command)
+            throws UnknownCommandException, MissingInformationException, ParseException {
+        verifier.verify(command);
+        String[] splitWhitesapce = command.split(" ");
```
chowyiwai: Typo in variable name?

chowyiwai (commented):
Hi Chee Peng!

I think you did a really good job so far 👍 I like how your code is separated into different classes so that each class has a very distinct purpose. Your Javadoc comments are also very elaborate but it would be good to add on and do them for all public methods. I also really like your variable names as they are very clear and it makes it easy for the reader to know what the variable represents. One thing I noticed though was that there were a few typos here and there but I think that can be easily fixed. 

I think I can learn from you in terms of separating my code into more classes so that each class is neater. Keep up the good work!

src/main/java/DialogBox.java
```diff
@@ -0,0 +1,58 @@
...
+            fxmlLoader.load();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        dialog.setText(text);
+        displayPicture.setImage(imageView);
+    }
+
+    private void flip() {
```
weigenie, line 36: Perhaps you might want to consider including the javadocs for these functions.

src/main/java/DialogBox.java
```diff
@@ -0,0 +1,58 @@
...
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
+        Collections.reverse(tmp);
+        getChildren().setAll(tmp);
+        setAlignment(Pos.TOP_LEFT);
+    }
+
+    public static DialogBox getUserDialog(String l, Image iv) {
```
weigenie, line 50: This too.

src/main/java/Duke.java
weigenie: I agree with yiwai that this condition may not be necessary.

src/main/java/com/chee/commands/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+import com.chee.model.TaskList;
+import com.chee.model.Deadline;
+import com.chee.model.ToDo;
+import com.chee.model.Event;
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+
+public class AddCommand implements Command {
```
weigenie: I like how you split all your Commands and Exceptions into smaller child classes to be more elaborate.

src/main/java/com/chee/commands/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+     * @throws ParseException If {@code extras} are not parceable.
+     */
+    public AddCommand(String commandType, TaskList taskList,
+                      DukePrinter dukePrinter, String... extras)
+            throws ParseException {
+        this.taskList = taskList;
+        this.dukePrinter = dukePrinter;
+        this.dateFormat = new SimpleDateFormat("dd/MM/yyyy HHmm");
+
+        switch (commandType) {
```
weigenie: Similar to how you have split the Command class into multiple child classes, perhaps you may want to split the AddCommand.

src/main/java/com/chee/commands/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+        case "event":
+            taskToAdd = new Event(extras[0], dateFormat.parse(extras[1]));
+            break;
+        default:
+            break;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
```
weigenie: I didn't know you could have done this!

src/main/java/com/chee/commands/FindCommand.java
```diff
@@ -0,0 +1,45 @@
...
+        this.dukePrinter = dukePrinter;
+    }
+
+    @Override
+    public void execute() {
+        List<Task> found = findTasks();
+        dukePrinter.printFind(found);
+    }
+
+    private List<Task> findTasks() {
```
weigenie: Could add javadocs

src/main/java/com/chee/io/DukePrinter.java
```diff
@@ -0,0 +1,158 @@
...
+    private static final String INDENTATION = "     ";
+    private static final String SEPARATOR_LINE =
+            "    ______________________________________________________";
+    private static final String WELCOME_MESSAGE =
+            SEPARATOR_LINE
+                    + "\n     Hello! I'm Duke\n"
+                    + "     What can I do for you?\n"
+                    + SEPARATOR_LINE
+                    + "\n";
+    private static final String LIST_MESSAGE =
```
weigenie: I like how you listed all these hardcoded messages as static final variables.

weigenie (commented):
I like the overall structure of your code and how object-oriented it is.

PR #123, opened 2019-08-24
[Lim Yu Rong, Samuel] Duke Increments

Sam-limyr:


wongsm7:
JavaDocs were done nicely, and code is neat. 
Can probably extract out Command classes, other than that well done!

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                            String taskDescription = Parser.parseDescription(input, true);
+                            ui.testEmptyDescription(taskDescription);
+                            taskList.createToDo(taskDescription);
+                            storage.writeSavedList(taskList.getList());
+                        } else {
+                            String taskDescription = Parser.parseDescription(input, false);
+                            ui.testTimeFormat(taskDescription);
+                            String taskContent = Parser.parseContent(taskDescription);
+                            LocalDateTime taskTime = Parser.parseTime(taskDescription);
+                            ui.testEmptyDescription(taskContent);
```
openorclose: `testEmptyDescription` doesn't really tell me what's going on, and doesn't seem like it throws an error. I had to go to the method declaration to see what it does.

maybe 

```java
if (somethingElse.isEmptyDescription(taskContent)) {
  throw new Exception(...);
}
```
would be better here as the reader can see what's going on.

If you would like to keep the exception inside the method i guess
```
throwIfEmptyDescription(...);
```
would be more descriptive as well;

Also don't think it should be part of ui.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                } else if (instruction.equals("todo") || instruction.equals("deadline")
+                        || instruction.equals("event")) {
+                    try {
+                        if (instruction.equals("todo")) {
+                            String taskDescription = Parser.parseDescription(input, true);
+                            ui.testEmptyDescription(taskDescription);
+                            taskList.createToDo(taskDescription);
+                            storage.writeSavedList(taskList.getList());
+                        } else {
+                            String taskDescription = Parser.parseDescription(input, false);
```
openorclose: Try not to use boolean arguments!

Better to have a `Parser.parseTodoDecscription` and a `Parser.parseDescriptionExcludingTime` to be more descriptive.

src/main/java/EventTask.java
```diff
@@ -0,0 +1,51 @@
...
+     * Returns a default formatted String for writing data to the user's hard drive.
+     * EventTask String includes tag identifying this as an EventTask, in addition to the Task default
+     * formatted String.
+     * Also includes the value of the deadline's time, formatted as a String.
+     *
+     * @return Returns a String.
+     */
+    @Override
+    public String formattedString() {
+        return "E | " + super.formattedString() + " | " + time.getDayOfMonth() + "/" + time.getMonthValue() + "/"
```
openorclose: You can simplify this a lot with a daete formatter!

src/main/java/Parser.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    /**
+     * Parses the time in the description of an input String.
+     * Time is parsed as a LocalDateTime, and is assumed to be formatted correctly.
+     *
+     * @param input The inputted String given by the user.
+     * @return Returns a LocalDateTime object representing the time of the Task.
+     */
+    static LocalDateTime parseTime(String input) {
+        String taskTimeBeforeParse = input.split("/by|/at", 2)[1].strip();
```
openorclose: Similarly this can be simplfied with a DateTimeFormatter

src/main/java/Ui.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+    
+    /**
+     * Tests for whether the time in a given task is correctly formatted.
+     * Valid time formats require a ' /by ' or ' /at ' to be present in the String.
+     *
+     * @param taskDescription The description of the given Task.
+     * @throws IncorrectTaskTimeFormatException A DukeException thrown to indicate an incorrect format.
+     */
+    void testTimeFormat(String taskDescription) throws IncorrectTaskTimeFormatException {
```
openorclose: think this shouldn't be part of Ui also

src/main/java/Ui.java
```diff
@@ -0,0 +1,100 @@
...
+    
+    /**
+     * Tests for whether the description of a given Task is empty.
+     * If empty, displays an exception message.
+     * A description consisting only of whitespace is considered empty.
+     *
+     * @param taskDescription The description of the given Task.
+     * @throws EmptyTaskDescriptionException A DukeException thrown to indicate an empty Task description.
+     */
+    void testEmptyDescription(String taskDescription) throws EmptyTaskDescriptionException {
```
openorclose: as mentioned think this shouldn't be part of ui

openorclose (commented):
I could understand your code quite easily. 

Some parts should be shorter though, like the gigantic if else statment for checking and executing instructions

PR #124, opened 2019-08-24
[Lee Bo Qiang] Duke Increments

lightz96:


src/main/java/duke/command/DeadlineCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * Executes Deadline command.
+     * @param taskList TaskList object for the duke program
+     * @param storage storage object for the duke program
+     * @return String to be printed
+     */
+    @Override
+    public String execute(TaskList taskList, Storage storage) {
+        try {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern(DATETIME_PATTERN);
+            String[] args = GetArgumentsUtil.getTwoCommandArgs(1, "/by", commands);
```
klementtan: Good separation of the utilities function and command.

src/main/java/duke/command/EventCommand.java
```diff
@@ -0,0 +1,46 @@
...
+     * @param storage storage object for the duke program
+     * @return String to be printed
+     */
+    @Override
+    public String execute(TaskList taskList, Storage storage) {
+        try {
+            String[] args = GetArgumentsUtil.getTwoCommandArgs(1, "/at", commands);
+            Task eventTask = new Event(args[0], args[1]);
+            taskList.addToTaskList(eventTask);
+            return String.join("\n", Messages.ADDED_TASK_MESSAGE, Messages.COMMAND_INDENTATION
```
klementtan, line 38: Can consider rendering the messages on a higher level. Ie Do not hard code the message rendered in the Command classes. If not careful, you might end up having discrepancies in the message format.

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     * Executes Exit command.
+     * @param taskList TaskList object for the duke program
+     * @param storage storage object for the duke program
+     * @return String to be printed
+     */
+    @Override
+    public String execute(TaskList taskList, Storage storage) {
+        try {
+            storage.saveData(taskList.getTaskList());
+            return Messages.BYE_MESSAGE;
```
klementtan, line 24: This is a good way of rendering messages on a higher level of abstraction

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+        return findCommand(commandArr);
+    }
+
+    /**
+     * Returns the correct command object based on user inputted command.
+     * @param commands is the last command entered by the user
+     * @throws DukeException if user inputted an invalid command
+     * @throws NumberFormatException if user inputted an invalid description for done and delete command
+     */
+    public static Command findCommand(String[] commands) throws DukeException, NumberFormatException {
```
klementtan, line 42: Can consider shortening the conditional statement with these by grouping the tasks under one condition so that you don't need to keep on throwing 
```
 else if (commands.length == 1) {
                throw new DukeException(String.format(Messages.DESCRIPTION_MISSING_EXCEPTION, "done"));
            } else {
                throw new DukeException(Messages.UNKNOWN_COMMAND_EXCEPTION);
            }
```
This will reduce the chances of human error and increase readability

src/main/java/duke/shared/GetArgumentsUtil.java
```diff
@@ -0,0 +1,48 @@
+package duke.shared;
+
+import duke.exception.DukeException;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class GetArgumentsUtil {
```
klementtan, line 11: Great idea to create a util class with static methods

klementtan:
Overall great job!!! Very impressive that you packaged the classes very nicely and great effort on writing the test cases

chensu2436 (commented):
My comments are more or less similar to what the other reviewer has pointed out. It's very impressive to see how your abstraction has been done!

PR #126, opened 2019-08-25
[Jin Shuyuan] Duke Increments

CoderStellaJ:


heze8 (commented):
Hello,
I think your code is well structured and clear. No serious problems whatsoever. Although some small issues with the Javadoc code style not aligning with CS2103's code style. Otherwise, it's logically sound and easy to follow, good job!

src/main/java/duke/Duke.java
```diff
@@ -27,7 +29,7 @@ public Duke(String filePath) {
     }
 
     /**
-     * Instantiate a Duke object and run it.
+     * Instantiate a duke.Duke object and run it.
```
heze8: To be particular, I noticed the Javadoc comments not following this excerpt from the code style page: "method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)".

It's not that serious though haha.

src/main/java/ui/Ui.java
```diff
@@ -47,7 +47,7 @@ public void showError(String message) {
     }
```
heze8: Furthermore, to be picky again, from the code style page, they state that all public methods and classes require Javadoc comments.

src/main/java/ui/Ui.java
```diff
@@ -33,7 +33,7 @@ public void showWelcome() {
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println(logo);
         showLine();
-        print("Hello! I'm Duke\nWhat can I do for you?");
+        print("Hello! I'm duke.Duke\nWhat can I do for you?");
```
heze8: Couldn't help notice that maybe when you refactored duke into a package, it also refactored all your comments and messages.

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,11 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DukeTest {
```
Krit-K: Having looking through your test function I could only find the sample test that was given. You may add more tests in specific classes. Make sure to allow access to the main/java classes that you want to refer to.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,41 @@
...
+     * @param description description of the task
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+
+    public String getStatusIcon() {
+        return (isDone ? "+" : "-"); //return tick or X symbols
```
Krit-K:
```suggestion
        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
```
I suggest using actual tick and X symbol with the specified code.

Krit-K (commented):
Your code looks great and I hope these minor touches will make your code better :)

src/main/java/tasks/Task.java
CoderStellaJ: thanks for your suggestion! I initially used ticks but due to character encoding difference, intellij console is not able to properly show the special chars, which is why I later change to +/-. But it's true that tick/x look nicer.

src/main/java/ui/Ui.java
CoderStellaJ: thanks for reminding me! I didn't notice it. Any good methods to avoid this situation when refactoring code?

src/test/java/DukeTest.java
CoderStellaJ: Thank you for pointing that out! If I have time after finishing this week's project, I will try to add more tests to make it complete. Thanks!

PR #128, opened 2019-08-25
[Chong Chee Yuan] Duke Increments

ccyccyccy:


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,59 @@
+package duke;
+
+import java.util.StringJoiner;
+
+import java.time.LocalDateTime;
```
FelixNWJ: Good use of the java Time classes to format your time neatly, without confusing any reader. I personally implemented a method just to convert the string into a time format, but it just adds unnecessary code for any reviewer.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by.format(outputFormat) + ")";
+    }
+
+    /**
+     * Used in saving of task data for storage.
+     * @return Formatted string representation of deadline.
+     */
+    @Override
+    public String getSaveString() {
```
FelixNWJ: Good use of the StringJoiner class here. Code looks neat and is good use when we need a fixed format for saving into a file.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Initialize and run program.
+     */
+    public void run() {
+        Ui ui = new Ui();
+        ui.showWelcome();
+
+        String input;
+        while (!(input = ui.readCommand()).equals("bye")) {
```
FelixNWJ: I understand that since ui.readCommand() will be lost if we do not store it, you have placed the assignment inside the while loop condition, but it may help if you add a comment here to clarify this since it looks hard to understand at first glance

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,73 @@
...
+            try {
+                Command c = Parser.parse(input);
+                c.execute(tasks);
+            } catch (DukeException e) {
+                System.out.println(e.getMessage());
+            }
```
FelixNWJ: Good work in breaking up the code specifically based on the statements that can throw a particular type of error

src/main/java/duke/ErrorCommand.java
```diff
@@ -0,0 +1,13 @@
+package duke;
```
FelixNWJ: Would it be better to make these kinds of Errors be an exception instead?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,32 @@
...
+public class Parser {
+
+    /**
+     * Returns specific Duke.Duke.Command from given user input.
+     * @param fullCommand User input string.
+     * @return Duke.Duke.Command to be executed.
+     */
+    public static Command parse(String fullCommand) {
+        String[] splitInput = fullCommand.split(" ");
+        String command = splitInput[0];
```
FelixNWJ, line 24: Code looks neat here and very organised, good as well adhering to the coding style for this particular case

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Save data to filePath specified.
+     * @param taskList Tasks to be saved.
+     * @throws IOException When cannot save task data.
+     */
+    public void save(TaskList taskList) throws IOException {
+        String directoryName = new File(filePath).getParent();
+        File directory = new File(directoryName);
+        directory.mkdir();
```
FelixNWJ, line 32: Is there a reason for creating a new directory here?

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,18 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
```
FelixNWJ, line 4: Good job implementing test methods that are easy to understand

FelixNWJ (commented):
Code looks very structured and organised, especially how there are many methods that are modularised for one task. Coding style is adhered to as well. If time is available, making sure that all public methods have javadoc comments is something that can be done to adhere to the coding standard as well.

PR #130, opened 2019-08-25
[Lin Yuting ] Duke Increments

linyutinglyt:


tswuxia (commented):
I like how the clear and detailed classification of exceptions that could appear. It is useful for identifying where it went wrong.

tswuxia (commented):
Maybe some of the code in the main Duke class can be directed into these classes for more OOP.

duke.txt
```diff
@@ -0,0 +1,4 @@
+E 0 potato 25/06/1998 1256
```
jascxx: This file can be included in .gitignore

src/main/java/Command.java
```diff
@@ -0,0 +1,170 @@
+import java.io.IOException;
+
```
jascxx, line 3: Each class should be written in a separate java file.

src/main/java/Command.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * constructor.
+     * @param tasks is the list of tasks.
+     * @param ui is the UI class.
+     * @param storage is the Storage class.
+     * @throws DukeException when input is different from required.
+     * @throws IOException when input is different from required.
+     */
+    void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException, IOException{
```
jascxx: Rather than putting a method with empty body, isn't it better to declare Command as an abstract class?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,33 @@
+import java.text.ParseException;
+import java.util.Date;
+
+public class Deadline extends Task {
+    /**
+     * a deadline task.
+     */
```
jascxx: Javadoc class description should be on top of the class declaration.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,49 @@
+public class DukeException extends Exception {
```
jascxx, line 1: Each class should be in a separate file.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,49 @@
...
+class EmptyCommandException extends DukeException {
+    private String type;
+
+    EmptyCommandException(String type) {
+        this.type = type;
+    }
+
+    @Override
+    public String toString() {
+        if (this.type.equals("event")) {
```
jascxx, line 17: Maybe it's better to check if type starts with a vowel instead of hardcoding the "event" case

src/main/java/Duke.java
```diff
@@ -1,10 +1,55 @@
...
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+    /**
+     * the main class, the Duke bot.
+     */
+    static SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
jascxx, line 26: Can be declared as final.

src/main/java/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+
+        default:
+            Task t;
+            String words = "";
+            boolean emptyCommand = !sc.hasNext();
+            if (!emptyCommand) {
+                words = sc.nextLine();
+            }
+
+            switch (command) {
```
jascxx, line 64: the nested switch blocks can be combined into 1 switch block

src/main/java/Parser.java
```diff
@@ -0,0 +1,89 @@
+import java.text.ParseException;
+import java.util.Scanner;
+
+class Parser {
+
+    /**
+     * parsing the statement by the user.
```
jascxx: Don't use continuous verb for Javadoc.

src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+     * loads the data from hard-disk file.
+     * @return an array list of task to be passed into TaskList.
+     * @throws FileNotFoundException when input format is different from required.
+     * @throws ParseException when input format is different from required.
+     */
+    ArrayList<Task> load() throws FileNotFoundException, ParseException {
+
+        ArrayList<Task> taskList = new ArrayList<>();
+        File f = new File(filepath); // create a File for the given file path
+        Scanner sc = new Scanner(f); // create a ScanneZZ:1q!r using the File as the source
```
jascxx: Accidental keyboard smashing in comment

src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+        while (sc.hasNext()) {
+            Task t = new Task(""); // dummie task
```
jascxx: Dummy task not really needed. Can just throw an error in the default switch block if there is no type.

src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * update the file when task list is changed.
+     * @throws IOException when input format is different from required.
+     */
+     void save() throws IOException {
+        FileWriter fw = new FileWriter(filepath);
+        for (Task task : taskList) {
+            String name = task.getDescription();
+            int status = task.getStatus()? 1 : 0;
+            if (task instanceof Todo) {
```
jascxx: Can use polymorphism rather than using "instance of"

src/main/java/Task.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    String getDescription() {
+        return this.description;
+    }
+
+    /**
+     * tells whether task is done or not.
+     * @return true or false of whether the task is done
+     */
+    boolean getStatus() {
```
jascxx: Should be named as "isDone()"

src/main/java/Task.java
```diff
@@ -0,0 +1,53 @@
+public class Task {
```
jascxx: Can be declared as abstract since this should not be instantiated

jascxx (commented):
Javadoc should start with capital letters, and @param argument should follow a certain format. There are also some minor styling errors (missing spaces). Otherwise, your code is easy to understand!

src/main/java/Command.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * whether is it an exit command
+     * @return true when it is an exit command, false otherwise.
+     */
+    boolean isExit() {
+        return (this instanceof ExitCommand);
+    }
+}
+
+class AddCommand extends Command {
```
Yoshi275, line 25: Please separate each class into a separate file. Each file should have one main class. If it's potentially messy, you can also package the different commands together as well.

src/main/java/Command.java
```diff
@@ -0,0 +1,170 @@
...
+     * command to add a task.
+     */
+    private Task t;
+
+    /**
+     * constructor.
+     * @param t is the task.
+     */
+    AddCommand(Task t) {
+        this.t = t;
```
Yoshi275, line 33: Potentially consider using` super(t)` instead, so you can utilise the constructor method from `Command` instead of adding a new constructor!

src/main/java/Command.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param tasks is the list of tasks.
+     * @param ui is the UI class.
+     * @param storage is the Storage class.
+     * @throws IOException when input is different from required.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws IOException {
+        tasks.addTask(t);
+        storage.save();
+        System.out.println("     Got it. I've added this task:");
```
Yoshi275: Have you considered passing such interactions to the responsibility of the `Ui` class instead? Might be better to maintain that all printing is done by the same class.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,33 @@
+import java.text.ParseException;
+import java.util.Date;
+
```
Yoshi275, line 3: Main classes require a javadoc description too!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,33 @@
...
+public class Deadline extends Task {
```
Yoshi275, line 4: Good use of the `Task` class as parent, definitely something for me to learn from.

src/main/java/Duke.java
```diff
@@ -1,10 +1,55 @@
...
+                ui.showError(e.getMessage());
+            } finally {
+                ui.showLine();
+                System.out.println();
+            }
+        }
+    }
+
+    public static void main(String[] args) throws IOException, ParseException{
+        new Duke("gradlduke.txt").run();
```
Yoshi275: I feel like there's a conventional name you have to name your file, but I'm not sure. Do check up on this if you can, just flagging it out!

src/main/java/Event.java
```diff
@@ -0,0 +1,33 @@
+import java.text.ParseException;
+import java.util.Date;
+
+public class Event extends Task {
```
Yoshi275, line 4: Again, I really like that your Deadline and Event class are inherited from the main Task class. Good choice!

Yoshi275 (commented):
Overall, good job. It's really interesting to see different implementations of the same methods, and I've definitely learnt a bit of refactoring of subclasses from you. Do try to ensure that each Class holds only one main class, and don't be afraid to start new `.java` files storing other classes. Jiayou!

PR #131, opened 2019-08-25
[Eugene Teu Chu Wei] Duke Increments

EugeneTeu:


joloong:
### Good work! :thumbsup:

Code is pretty well documented which makes it easy to read and review. This is a good takeaway and I should adopt this good habit.

One improvement I could suggest is that the commit messages could be written in imperative and first letter could be capitalized. E.g. "Clean up code for code review" instead of "cleaned up code for code review". 

P.S. Happy birthday! :cake:

PR #138, opened 2019-08-26
[Jonas Ong] Duke Increments

e0031374:


ghost (commented):
Refreshing architecture of the code, also implemented most of the extensions and optional tasks.

Very well documented with the use of interface which was not a requirement but great practice.

Code style also appears to follow style guide with rare instances of redundant blank lines.

I liked the use of streams to make the code refreshingly clean.

Test cases were quite comprehensive, task finished way beyond the requirement. Great work.

src/main/java/BasicTaskCreator.java
```diff
@@ -0,0 +1,132 @@
...
+ * 3. Make a compositeTaskCreator and accept ToDoTaskCreator et al
+ */
+
+/**
+ * Factory class which takes in string of details of what to make each class
+ * returns TaskInterface implemented class eg ToDo, Event, Deadline
+ * called by Parser addCommand.
+ */
+class BasicTaskCreator implements TaskCreator {
+    public BasicTaskCreator() {
```
chaitanyabaranwal, line 20: This might not be needed since Java creates a default constructor.

src/main/java/BasicTaskCreator.java
```diff
@@ -0,0 +1,132 @@
...
+                + task 
+                + " cannyot be empty!";
+            throw new OWOException(errMsg);
+        }            
+        return true;
+    }
+
+    /* converts a str.split(" ") back into a string */
+    public static String listToString(List<String> xs) {
+        String output = "";
```
chaitanyabaranwal, line 105: Might want to use String.join() method instead!

src/main/java/DeadlinesImplementation.java
```diff
@@ -0,0 +1,77 @@
...
+    private final Date objDate;
+    private final boolean isDone;
+    private final TaskFormatInterface formatter;
+    private final String sym = "D";
+
+    public DeadLinesImplementation(String name, String date, 
+            boolean isDone) {
+        this.name = name;
+        this.date = date;
+        this.objDate = setDate(date);
```
chaitanyabaranwal, line 17: Might want to just use one `Date` object in the instance variables, instead of using a date string and a date object. The reason is that since they are storing the same information, it might be redundant data. It's true that providing this extra variable lets the user enter the date in any format but I feel that it is good to standardise the date across different instances of the object.

src/main/java/BasicTaskCreator.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private TaskInterface createDeadLine(String command)
+        throws OWOException {
+
+        String[] subcmdList = removeFirstAndDelimit(command, 
+                " /by ");
+        boolean isValid = checkCommand(command, "deadline");
+
+        String taskName = subcmdList[0];
+        String date = subcmdList.length > 1 ? subcmdList[1] : "";
```
chaitanyabaranwal, line 43: This allows for an empty date input, which I am not sure is the ideal behaviour.

src/main/java/BasicTaskCreator.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private TaskInterface createToDo(String command) 
+        throws OWOException {
+
+        String[] cmdList = command.split(" ");
+        boolean isValid = checkCommand(command, "todo");
+        List<String> xs = 
+            new LinkedList<String>(Arrays.asList(cmdList));
+        xs.remove(0);
```
chaitanyabaranwal, line 30: Why not use `removeFirstAndDelimit()` here also?

src/main/java/DeadlinesImplementation.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public String getName() {
+        return this.name;
+    }
+
+    public boolean getIsDone() {
+        return this.isDone;
+    }
+
+    public String getSymbol() {
```
chaitanyabaranwal, line 47: Instead of using symbols to represent the task type, might want to consider using an enumeration representing the different task types.

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Returns void, call UI to generate UI and start session.
+     *  for user
+     */
+    public void start() {
+        this.display.instance();
+    }
+
+    public void stop() {
```
chaitanyabaranwal, line 47: This method does not seem to do anything, might want to consider removing.

src/main/java/BasicTaskFormatter.java
```diff
@@ -0,0 +1,26 @@
+class BasicTaskFormatter implements TaskFormatInterface {
```
chaitanyabaranwal, line 1: Instead of using a `BasicTaskFormatter`, might want to consider just putting the string format of the tasks in their respective class definitions. I personally feel that the class relationships are getting too artificial with so much inheritance, and that it is not required.

src/main/java/EventsImplementation.java
```diff
@@ -0,0 +1,77 @@
+import java.util.Date;
+import java.text.SimpleDateFormat;
+
+class EventsImplementation implements TaskInterface,
```
chaitanyabaranwal, line 4: Same thing, probably consider using only a `Date` object.

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Returns nothing, this command evaluates user command
+     *   and executes what needs to be done
+     *   individual command relies on TaskCreator implementation.
+     *
+     * @param command Input user command
+     * @return void
+     */
+    public void whatsGoingOn(String command) {
```
chaitanyabaranwal, line 77: Function names need to be more descriptive.

src/main/java/StorageInterface.java
```diff
@@ -0,0 +1,3 @@
+interface StorageInterface {
```
chaitanyabaranwal, line 1: Not sure what purpose `StorageInterface` serves, might want to add some functions to it!

chaitanyabaranwal (commented):
- I like that the tests create are very extensive, and usually cover most functions of the class.
- I like how marking a task returns a new instance, ensures a higher degree of immutability.
- Function and variable names need to better represent the intended purpose.
- Extensive documentation which makes the code easy to understand.
- Mostly followed code style.

PR #139, opened 2019-08-26
[Daryl Tan] Duke Increments

openorclose:


src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+    commandParser.add(new MarkAsDoneCommand(ui::error, this::findInListThenMarkAsDone));
+    commandParser.add(new DeleteCommand(ui::error, this::findInListThenDelete));
+    commandParser.add(new AddTodoTaskCommand(ui::error, this::addTaskToList));
+    commandParser.add(new AddEventTaskCommand(ui::error, this::addTaskToList));
+    commandParser.add(new AddDeadlineTaskCommand(ui::error, this::addTaskToList));
+    commandParser.add(new FindCommand(ui, this::findTasksMatchingQuery));
+    loadFromDiskToList();
+  }
+
+  private void printWelcome() {
```
Sam-limyr, line 87: I think this method should exist within the Ui object instead, and Duke should only call it as "ui.printWelcome( )".

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+   *
+   * @param args (unused)
+   */
+  public static void main(String[] args) {
+    Duke duke = new Duke();
+    duke.printWelcome();
+    duke.startRepl();
+  }
+
+  public Duke(Ui ui) {
```
Sam-limyr, line 48: You could consider using only one constructor for Duke, and assigning the Ui and Store classes in a run( ) method instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+   * Parses and executes command.
+   *
+   * @param command command to execute
+   */
+  public void parseAndExecuteCommand(String command) {
+    commandParser.parseAndExecuteCommand(command);
+    saveListToDisk();
+  }
+
+  private void loadFromDiskToList() {
```
Sam-limyr, line 114: This method should be abstracted out into the Storage class instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+    Scanner saveFileScanner = store.retrieveDataAsScanner();
+    while (saveFileScanner.hasNextLine()) {
+      String serialized = saveFileScanner.nextLine();
+      list.add(taskParser.fromSerial(serialized));
+    }
+  }
+
+  private void saveListToDisk() {
```
Sam-limyr, line 123: This method should also be abstracted out into Storage.java.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+    store.saveDataIntoFile(list.stream().map(Task::serialize).collect(Collectors.joining("\n")));
+  }
+
+  private void findInListThenMarkAsDone(int oneBasedIndex) {
```
Sam-limyr, line 127: Consider abstracting this method and the following five methods (all the way to the end of Duke.java) out into TaskList.java instead.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,54 @@
...
+
+  private HashMap<String, Function<String, T>> commands = new HashMap<>();
+  private Ui ui;
+
+  /**
+   * Instantiates a new Parser.
+   *
+   * @param ui the ui
+   */
+  public Parser(Ui ui) {
```
Sam-limyr, line 22: I think the Parser object and the Ui object should remain separate (i.e. both only initialized in Duke.java). Maybe you could consider using normal println( ) methods in Parser, or using static methods for the Ui class instead?

Sam-limyr (commented):
I'll be honest with you - your code is damn intimidating and I wish I could say I read it properly but it's too scary so I just gave some simpler comments for the parts that didn't scare me :p

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+    ui.println("What can I do for you?");
+
+  }
+
+  private void startRepl() {
+    while (true) {
+      parseAndExecuteCommand(scanner.nextLine().trim());
     }
+  }
+
```
wongsm7: Could probably add in JavaDocs for methods without JavaDocs, making it easier to understand

wongsm7 (commented):
Maybe the basic indentation could be changed to 4 spaces instead of 2?
Can also add in JavaDocs for some of the methods.

PR #140, opened 2019-08-26
[Chia Wei Hao, David] Duke Increments

Davidcwh:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,130 @@
...
+        super(description);
+        this.by = by;
+
+        String[] dateTime = by.split(" ");
+        String[] date = dateTime[0].split("/");
+
+        SimpleDateFormat format = new SimpleDateFormat("dd/MM/yyyy hh:mm");
+        try {
+            this.day = date[0];
+            this.month = date[1];
```
weiyang13, line 37: Good effort on using `SimpleDateFormat` to create `Date` objects! However, for the parsing, you may want to consider redefining the `SimpleDateFormat` to fit the expected input, and simply calling parse on `by`. This would also allow easy scaling in the future, in case you want to accept multiple date formats. 

I understand that you may have hard coded the parsing due to errors encountered when using `SimpleDateFormat`. If so, do feel free to ask me (or our peers on Slack) for help!

On a separate note, I am curious as to whether your code works well in distinguishing AM and PM timings, as I note that you used `"hh:mm"` instead of `"HH:mm"` or `"hh:mm aa"` to parse the time. According to the documentation of `SimpleDateFormat`, there is a subtle distinction between `"HH"` and `"hh"`, where `"HH"` is more suitable for the 24-hour clock system.

Apart from that, good job for this part of the code! I particularly appreciate your effort in printing the date in a verbose style to improve user experience.

src/main/java/Duke.java
```diff
@@ -1,4 +1,193 @@
...
+    private void displayLogo() {
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+
+        String logo2 =
+                "DDD    U     U   K     K    EEEEEE \n" +
+                "D    D   U     U    K  K     E    \n" +
```
weiyang13: I like the customized logo for Duke! It adds to the flavor of the code. 😃

src/main/java/InvalidCommandDukeException.java
```diff
@@ -0,0 +1,9 @@
+/**
+ * Represents an exception that is thrown when an invalid command is entered
```
weiyang13, line 2: I like it that you define subclasses of DukeException. This makes the logic of exception handling much clearer!

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,29 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DeadlineTest {
+    @Test
+    public void testStringConversion() throws Exception {
+        assertEquals("[D][\u2718] test deadline (by: 1st of January 2019, 06:00PM)",
```
weiyang13, line 9: Good job on developing unit tests! I appreciate that you developed a wide range of test cases (as compared to what was required). To further improve the logic of the tests, perhaps you could consider catching any exceptions thrown where it is not expected and use the `fail()` method, instead of throwing the exceptions. 

Also, I notice that you are catching and throwing the superclass `Exception` instead of the appropriate `DukeException` subclasses. Perhaps you could consider using the `DukeException` subclasses instead, as this would better capture the logic of "what could go wrong" (as expected by the method) , "what should go wrong" and "what should not go wrong" in your test cases.

src/main/java/Ui.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Returns the list of tasks in the TaskList object argument.
+     *
+     * @param tasks TaskList object with tasks to be displayed to the user in a list.
+     * @return String list of tasks in TaskList.
+     */
+    public String showListMessage(TaskList tasks) {
+        String list = line + "     Here are the tasks in your list:" + "\n";
```
weiyang13: Good job on abstracting out the UI! I appreciate the use of protected variables to store `String` objects for easy modification of the code, which I have not done much myself. You could consider extending this philosophy to indentations as well, to make it easier to modify it in the future, especially as it repeatedly occurs in the code.

weiyang13 (commented):
Good job on the code so far! The code is generally well-commented, and the separation of the code into appropriate modules is done well.

A general observation across the code I would like to highlight is the extensive presence of unused getters and setters. I understand that this is a convention you may have learnt to stick to previously. However, it is unnecessary to always define getters and setters for your variables. In fact, it is arguable that such a practice sometimes causes the abstraction barrier to be broken if you are not cautious. Perhaps you could consider only implementing these methods where necessary, and restricting access modifiers for these methods to ensure that the abstraction barrier is not broken.

Other than that, I can see that you put a lot of effort and thought in your code, and the overall structure is very good!

P.S. I did not review the JavaFX elements of the code as I have not done those myself and do not have sufficient understanding to give meaningful feedback.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public Date getByDate() {
+        return byDate;
+    }
+
+    public void setByDate(Date byDate) {
+        this.byDate = byDate;
+    }
+
+    public String getDay() {
```
WilliamRyank: Hi,

I see that you are trying to make getter and setter for the Deadline object. May i know the use of this setters and getters, as i could not really find the use of these getters and setters in your code. Maybe, i can learn from your point of view. However, just a small input, if you do not really need these methods, you can either not write them or you can have 1 method that will return the date, unless there is really a need to getHour or getMinute, i think defining these methods will make ur cleaned and organized code become a little bit messy.

But overall is good, im just curious about this approach, maybe i can learn from you as well. Cheers!!

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,48 @@
...
+
+    }
+
+    /**
+     * Indicates the exit condition of the running Duke object.
+     *
+     * @return DoneCommand is not the ExitCommand so it returns false.
+     */
+    @Override
+    public boolean isExit() {
```
WilliamRyank: Hi,

I see that you have this code repeated over the command. Do remember the DRY principle. My suggestion is that you define your isExit in the Command class as return false, and for the "bye" command, you override it with return true. I believe this will have the same effect with your current code and it will make your code shorter and do not repeatedly override the isExit method. Cheers!!

src/main/java/Event.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public Date getAtDate() {
+        return atDate;
+    }
+
+    public void setAtDate(Date atDate) {
+        this.atDate = atDate;
+    }
+
+    public String getDay() {
```
WilliamRyank: Same comment as the Deadline class.

WilliamRyank (commented):
Hi David,

Overall your code is really clean and broken down to a smaller classes. I do personally like it as it will make your Duke.java file actually very clean and concise. Great attempt on the javadoc as well as they are descriptive and cohesive.
However, i noticed that for some of your classes such as Deadline and Event class, you have a lot of setters and getters methods. While i do not know your perspective, from what i can see from your code, you never really use these methods and i doubt it will be needed in the future as well. Please do let me know your point of view as i can also learn from different perspectives. Furthermore, just remember not to repeat some of your code and use the power of polymorphism.

Overall, great attempts and really organized code. Cheers!!!

src/main/java/Deadline.java
Davidcwh: Thanks for the constructive criticism overall, I definitely learned how to make my code better from your comments especially regarding the unused getters and setters. Yes it was a convention I practiced before and IntelliJ made it easier with the "generate" function. What you said made sense, I would try to make my code as neat and concise in the future by getting rid of such unnecessary methods. 

Regarding the AM and PM timings, I do have a `dateToString` method below in this code that parses the date into string with `hh:mma` which is called in the `toString` method. Nonetheless, good eye catching this!

src/main/java/Deadline.java
Davidcwh: Thanks for the useful feedback overall, I really appreciate it! 

Regarding the unused getter and setter methods, I realized it may be a mistake on part as I implemented these methods by convention regardless of any use of them. I have learned that it would be bad practice to do so as it may sometimes cause the abstraction barrier to be broken. 

Also, good eye catching the repeated use of the `isExit` method in my `Command` classes. What you suggested made a lot of sense and was something that slipped my mind while coding the `Command` classes. 

Once again, thanks for the constructive comments, they definitely improved my current code!

PR #141, opened 2019-08-26
[Ooi Qiu Jia] Duke Increments

qiujiaaa:


src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param storage The Storage object we are currently using.
+     */
+    public abstract void execute(TaskList taskList, Ui ui, Storage storage);
+
+    /**
+     * Checks if the command is an ExitCommand.
+     *
+     * @return True if it is an ExitCommand, false if otherwise.
+     */
+    public abstract boolean isExit();
```
sianghwee: Hi, since for most of the command classes (AddCommand, DoneCommand, DeleteCommand etc) the isExit() method returns false, perhaps it would be easier to just implement the isExit() method in the abstract class and just override the isExit() method for the ExitCommand such that it returns true just for the ExitCommand. In that way, you do not need to repeat the isExit() method implementation so many times.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+import duke.ui.Ui;
+
+import duke.task.TaskList;
+
+import duke.storage.Storage;
+
+public class AddCommand extends Command {
+
+    String command;
```
sianghwee, line 17: Is there any reason why you choose to put this as default instead of private?

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,26 @@
+package duke.exception;
+
+public class DukeException extends IllegalArgumentException {
```
sianghwee, line 3: Is there a reason why you choose to extend from IllegalArgumentException instead of the general Exception class?

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,26 @@
...
+
+    String message;
+
+    /**
+     * Constructor to create a DukeException object.
+     *
+     * @param message
```
sianghwee: Even though the variable name is clear enough, perhaps you can write a few words to describe the variable so that it is even clearer to the users. For example, "a string that describes the error encountered"

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Parses the user's commands.
+     *
+     * @param command User's command.
+     * @return Command object.
+     * @throws DukeException If the input cannot be recognised.
+     */
+    public static Command parse(String command) throws DukeException {
+        if (command.equalsIgnoreCase("bye")) {
```
sianghwee, line 26: In a case where you might have multiple options, it might be better to use a switch statement instead of multiple if-else statement

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,33 @@
...
+     * Exit the chatbot. Get Ui to show the exit message and Storage to update data.
+     *
+     * @param taskList TaskList object containing the current tasks list.
+     * @param ui The Ui object we are currently using.
+     * @param storage The Storage object we are currently using.
+     */
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) {
+        ui.showExit();
+        storage.updateData(taskList.getTasks());
```
sianghwee: I noticed that the updateData() method is only run when exiting the program. However, based on my understanding of the objectives of the level, it is expected to update the file whenever the list changes. Could I have overlooked areas in the code where the save file is updated when a add or delete command is done?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+                } else {
+                    sb.append("0 ");
+                }
+                sb.append(task.getDescription());
+                fw.write(sb.toString() + "\n");
+            }
+            fw.close();
+        } catch (IOException e) {
+            ui.showError("Data file does not exist.");
+        } finally {
```
sianghwee: I think that there is no need for a _finally_ statement in this case since there is nothing to be handled.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,57 @@
+package duke.task;
+
+import duke.exception.DukeException;
+
+public class Deadline extends Task {
+
+    protected String by;
+
+    protected int date;
```
sianghwee, line 9: Is there a reason why all of these attributes are protected instead of private? Why is there also a need to store the date, month, year and time if it is not used?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+
+public class TaskList {
+
+    List<Task> tasks;
+
+    /**
+     * Constructor to create a TaskList object.
+     */
+    public TaskList() {
+        tasks = new LinkedList<>();
```
sianghwee: Is there a reason for the choice of LinkedList over ArrayList? Personally, I feel that in this project, there is a lot of referencing an object based on its index and so the use of a LinkedList will not be as efficient as that of an ArrayList. Of course, since the lists tend to be small, the cost is not significant, but just a point to consider perhaps.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+                if (type.equalsIgnoreCase("deadline")) {
+                    task = new Deadline(command.substring(9, index - 1), command.substring(index + 4));
+                } else if (type.equalsIgnoreCase("event")) {
+                    task = new Event(command.substring(6, index - 1), command.substring(index + 4));
+                } else {
+                    throw new IllegalArgumentException("     ☹ OOPS!!!No such task type.");
+                }
+            }
+            tasks.add(task);
+            ui.addedTask(task, tasks.size());
```
sianghwee: Perhaps, this line should be moved to the add command as this method seems to be doing a lot of things at the moment.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public void showWelcome(List<Task> tasks) {
+        showLine();
+        System.out.println("     Hello! I'm Duke\n" + "     What can I do for you?\n");
+
+        try {
+            if (!tasks.isEmpty()) {
+                printTasks(tasks);
+            } else {
+                System.out.println("     There are no tasks in the list right now.");
```
sianghwee: Good idea! Perhaps you could also do this for the printing of the available task too when the tasklist is empty.

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class ParserTest {
+    @Test
+    public void test_parse_bye(){
+        Command bye = Parser.parse("BYE");
+        assertEquals(true, bye instanceof ExitCommand);
+    }
+
+    @Test
+    public void test_parse_add(){
+        Command bye = Parser.parse("aDd");
```
sianghwee: Perhaps you can use a variable name that reflects the addition command since it confuses the user as they think that this is a _bye_ command. Good testing of the command though, to include words with different case.

src/test/java/duke/TaskTest.java
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    public void test_constructor(){
+        assertEquals("[T][" + "\u2718" + "] do tutorials", (new Todo("do tutorials")).toString());
+    }
+
+    @Test
+    public void test_markAsDone() {
+        Todo todo = new Todo("do work");
+        todo.markAsDone();
+        assertEquals(true, todo.getIsDone());
```
sianghwee: I think it should be isCompleted() instead.

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,45 @@
+package duke;
+
+import duke.command.*;
```
sianghwee: I think it is not advisable to use wildcard imports.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,166 @@
...
+        try {
+            keyword = keyword.substring(4).trim();
+            if (keyword.isEmpty()) {
+                throw new DukeException("     ☹ OOPS!!! Search keyword cannot be empty.");
+            }
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeException("     ☹ OOPS!!! Search keyword cannot be empty.");
+        }
+        if (tasks.isEmpty()) {
+            throw new DukeException("     ☹ OOPS!!! The list is empty.");
```
sianghwee: Having so many white space in your code make it hard to read and it is hard to debug if there is an additional space etc. Perhaps you can use a _\t_ to standardize it.

sianghwee (commented):
Hi, overall a good job. Really like the fact that you group all the related packages together and also the way you handle certain exceptions which shows that you really gave a lot of thought to the possible incorrect input that a user might input. Just have a little question regarding some of the choices that you might have made which I have linked to the code.

PR #142, opened 2019-08-26
[Seah Lynn] Duke Increments

seahlynn:


src/main/java/Duke.java
```diff
@@ -1,10 +1,155 @@
-public class Duke {
+
```
jonchan51, line 1: You should take a look at javafx tutorial 4 as a way to refactor all of these into a separate file. It cleans up the code a lot more.

src/main/java/Duke.java
```diff
@@ -1,10 +1,155 @@
...
+        Label userText = new Label(userInput.getText());
+        Label dukeText = new Label(getResponse(userInput.getText()));
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(userText, new ImageView(user)),
+                DialogBox.getDukeDialog(dukeText, new ImageView(duke))
+        );
+        userInput.clear();
+    }
+
+    private void handleExceptions(DukeException e) {
```
jonchan51, line 143: I think you could just add a try block inside `handleUserInput()` to use `e.getMessage()` instead of `getResponse()` so it is all in one method. Handling exceptions are part of dealing with the user input after all.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,75 @@
+/**
+ * Command corresponding to user input starting with either "todo", "deadline" or "event".
+ * Executes the command input by using UI and adds the corresponding tasks to the tasklist.
+ */
+public class AddCommand extends Command {
+    protected Type enumType;
```
jonchan51: Is there a reason why you chose to make it protected instead of private?

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,75 @@
...
+        switch (enumType) {
+            case TODO: {
+                Task newTask = new Todo(this.taskDesc);
+                tasks.add(newTask);
+                int numTasks = tasks.size();
+                printable = "Got it. I've added this task:" + "\n" + newTask.toString() +
+                            "\n" + "Now you have " + numTasks + " tasks in the list.";
+            }
+
+            break;
```
jonchan51, line 36: Think it would be better to just leave the `break;` inside the case block, so the case only has a block with it instead of a block and one more statement.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,75 @@
...
+
+        storage.save(tasks.getTaskList());
+        return printable;
+    }
+
+    /**
+     * Determines whether the program stops running.
+     * @return boolean value false so the program continues to run
+     */
+    public boolean isExit() {
```
jonchan51: I think it would be better to define this method in the abstract class, then override it in the ExitCommand subclass. That way you won't need to redefine this function in every command.

src/main/java/Command.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * Parent class of all the command objects that deals with user inputs.
+ */
+public abstract class Command {
+    protected String taskDesc;
```
jonchan51: I think these aren't used anywhere anymore! It's a good idea to go through your code base once in a while to clean and fix things.

src/main/java/Command.java
```diff
@@ -0,0 +1,27 @@
...
+    protected String timeDesc;
+    protected TaskList tasks;
+    protected Ui ui;
+    protected Storage storage;
+
+    /**
+     * Adds the task inputted by the user to the task list.
+     * and prints out corresponding response
+     * @param tasks holds the list of tasks currently in the program
+     * @param ui displays the output from execution
```
jonchan51: Like here too! You forgot to remove the `ui` param

src/main/java/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + date + ")";
+    }
+
+    /**
+     * Converts the Deadline object to the String form to be saved in file.
+     * @return String format of the object
+     */
+    public String toFileFormat() {
+        if (isDone) {
```
jonchan51, line 46: think it will be cleaner to have a separate method to return the tick and cross depending on `isDone`, then just have one return statement that calls this method as well to get the tick and cross.

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Adds the program task inputted by the user to the task list.
+     * and prints out corresponding response
+     * @param tasks holds the list of tasks currently in the program
+     * @param storage stores the added task to the specified file
+     * @throws DukeException if task requirements is not met
+     */
+    public String execute(TaskList tasks, Storage storage) throws DukeException {
+        int numChange = Integer.parseInt(this.num);
```
jonchan51: will be good to handle the case where a number is not fed in, otherwise it crashes the program :(

src/main/java/Events.java
```diff
@@ -0,0 +1,62 @@
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.text.ParseException;
+
+/**
+ * Represents an "Event" task inputted by the user.
+ */
+public class Events extends Task {
```
jonchan51: Just a naming thing, but I think this should be called Event instead of Events

src/main/java/KeyCommand.java
```diff
@@ -0,0 +1,64 @@
...
+ * Executes the command input by using UI and adds the corresponding tasks to the tasklist.
+ */
+public class KeyCommand extends Command {
+    protected String[] keyword;
+
+    /**
+     * Constructs a AddCommand object.
+     * @param keyword indicates the keyword the user is searching for
+     */
+    public KeyCommand(String... keyword) {
```
jonchan51, line 14: Nice idea to use a vararg here!

src/main/java/KeyCommand.java
```diff
@@ -0,0 +1,64 @@
...
+     * @return String to be displayed as Duke response in GUI
+     * @throws DukeException if task requirements is not met
+     */
+    public String execute(TaskList tasks, Storage storage) {
+        String printable = "";
+        printable = "Here are the matching tasks in your list: " + "\n";
+        ArrayList<String> resultList = new ArrayList<String>();
+
+        for (int i = 0; i < tasks.size(); i++) {
+            String[] taskDesc = tasks.get(i).getDesc().split(" ");
```
jonchan51: I don't think you need to do the `.equals`. It should be sufficient to just check whether the description `.contains` any of the keywords.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,120 @@
+import java.util.ArrayList;
+
+/**
+ * Represents the list of tasks in the program.
+ */
+public class TaskList {
+    protected ArrayList<Task> taskList;
+    protected ArrayList<String> fileList;
```
jonchan51: I don't think the `fileList` is necessary as an instance variable, you're only using it in the constructor after all.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public ArrayList<Task> getTaskList() {
+        return this.taskList;
+    }
+
+    /**
+     * Prints out all tasks currently in the list.
+     * @return String corresponding to tasks in their String format
+     */
+    public String printTaskList() {
```
jonchan51, line 79: should rename this method since it does not print anymore

src/main/java/TaskList.java
```diff
@@ -0,0 +1,120 @@
+import java.util.ArrayList;
+
+/**
+ * Represents the list of tasks in the program.
+ */
+public class TaskList {
```
jonchan51, line 6: Considering the way you're adding get, add and remove methods, you might want to consider having TaskList extend ArrayList<Task> like what I did

src/main/java/Storage.java
```diff
@@ -0,0 +1,64 @@
...
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
+     * Returns an Array:ist of String pertaining to task objects in their string format.
+     * @return an Array:ist of String
+     * @throws DukeException if the file cannot be found
+     */
+    public ArrayList<String> load() throws DukeException {
```
jonchan51, line 28: Is there a reason why you chose to parse the file data in your TaskList constructor instead of in this method?

src/main/java/Parser.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Returns differnt types of Command objects depending on user input.
+     * @param command represents the input by the user
+     * @return Command object according to what the user inputs
+     * @throws DukeException when the input is missing information or
+     *                       in the wrong format
+     */
+    @SuppressWarnings("checkstyle:FallThrough")
```
jonchan51: recommended not to do this :( Should instead address the fallthrough you put into your code

src/main/java/Todo.java
```diff
@@ -0,0 +1,47 @@
...
+            String format = "T | [✗] | " + taskDesc + "\n";
+            return format;
+        }
+    }
+
+    /**
+     * Retrieves the time description of the object.
+     * @return null because Todo tasks do not have a time description
+     */
+    public Date getDate() {
```
jonchan51, line 38: I think it's better to only declare this in the subclass. Since todo does not have a date attribute, it shouldn't have a getDate() method either

src/main/java/Type.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * Enumeration representing the type of task object.
+ */
+public enum Type {
+    TODO, DEADLINE, EVENT
```
jonchan51: oh good idea making the types into enums!

jonchan51 (commented):
Good job! It was pleasant to go through and read your code. Your variables are well named and your docs descriptive. 

Something you might want to take note of is to try to maintain some organization with your files. While it was optional, it is a good habit to keep your classes in packages so that there is a clear distinction between the general purpose of different classes.

You should also take some time every few major commits to go through your code base and ensure that there isn't any redundant code / unused things. In the case of a larger project, having things scattered around can get quite messy.

Also, good tests, might want to follow the style guide's way of naming your test methods so they are more descriptive.

src/main/java/Type.java
jonchan51: Oh one more thing I want to add, you should try to improve the way you make commit messages. Start with capital letter, using imperatives as the first word, etc.

src/main/java/Type.java
```diff
@@ -0,0 +1,3 @@
...
```
alxkohh: I think it was excellent design thinking on your part to come up with this enum, great job! My suggestion is to make the name of this enum more descriptive, eg. "TaskType" instead of "Type".

alxkohh (commented):
-

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,31 @@
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ParserTest {
+    @Test
```
alxkohh: Very comprehensive JUnit testing overall! Test methods were descriptive, hence it was easy to understand what tests you were trying to run.

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,31 @@
+import org.junit.jupiter.api.Test;
```
alxkohh, line 1: Very comprehensive JUnit testing overall! Test methods were descriptive, hence it was easy to understand what tests you were trying to run.

alxkohh: My suggestion for this commit is to follow standard convention for your commit message, eg. "Add JUnit testing"

src/main/java/AddCommand.java
```diff
@@ -1,16 +1,32 @@
...
 
+    /**
+     * Adds the program task inputted by the user to the task list.
+     * and prints out corresponding response
+     * @param tasks holds the list of tasks currently in the program
+     * @param ui displays the output from execution
+     * @param storage stores the added task to the specified file
+     * @throws DukeException if task requirements is not met
+     */
     public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
```
alxkohh: You forgot to type @Override on top of this method, but no worries, not a big issue. Good use of inheritance!

src/main/java/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+                }
+
+                return new DeleteCommand(commandSplit[1]);
+
+            case "bye":
+                return new ExitCommand();
+
+            case "todo":
+            case "deadline":
+            case "event":
```
alxkohh: Nice job using switch statement for this method. I think it does make things a little more readable as compared to if else blocks. However, I feel that your choice of using nested switch here does backfire a bit in terms of readability, maybe you can consider directly listing out the case for "todo", "deadline" and "event", instead of using a nested switch? Just my two cents!

src/main/java/KeyCommand.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        ui.showLine("Here are the matching tasks in your list: ");
+        ArrayList<String> resultList = new ArrayList<String>();
+
+        for (int i = 0; i < tasks.size(); i++) {
+            String[] taskDesc = tasks.get(i).getDesc().split(" ");
+
+            for (int j = 0; j < taskDesc.length; j++) {
```
alxkohh: I agree with Jonathan, contains captures the logic more cleanly as compared to equals. Maybe you can consider modifying!

alxkohh:
Overall, your program was well designed in my humble opinion, I feel that you have put in a lot of effort and thought to come up with this app. It was pleasant reviewing your code, very little mistakes and very readable. Keep it up!

src/main/java/Duke.java
seahlynn: ahhh thanks!! I missed out on that tutorial!! Will go work on it now!

src/main/java/Type.java
seahlynn: OKay!! I haven't really paid much attention to my commit messages so far HAHA! Thanks for pointing it out!

src/test/java/ParserTest.java
seahlynn: OKay! Will take note of my commit messages in the future HAHAHA thankyou!!

PR #144, opened 2019-08-26
[Tan Likai] Duke Increments

tanlk99:


src/main/java/Parser.java
```diff
@@ -22,14 +24,14 @@ public static Command parseInput(String rawInput) throws DukeException {
             try {
                 index = Integer.parseInt(rawInput.split(" ", 2)[1]) - 1;
                 return new DoneCommand(index);
-            } catch (NumberFormatException e) {
+            } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
```
hellodommy: Good idea to use | for catching multiple exceptions! 😊I will try that for mine.

src/main/java/duke/util/Parser.java
```diff
@@ -51,6 +53,12 @@ public Command parseInput(String rawInput) throws DukeException {
             } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
                 throw new DukeException("Please use 'delete i' to delete the i-th task in the list.");
             }
+        case "find":
+            if (!rawInput.contains(" ")) {
```
hellodommy: I did not consider the case where the search term is empty.
Good idea, I will try it.

src/main/java/duke/util/TaskList.java
```diff
@@ -66,6 +66,21 @@ public void deleteTask(int index) {
...
+    /**
+     * Checks if the description of a task given by index contains a given filter string.
+     * If the index is invalid, this method will throw an assertion error. This search is
+     * case-insensitive.
+     *
+     * @param index     1-indexed value indicating position of the task to check
+     * @param filter    the string to check
+     * @return          true if task description contains filter string
+     */
+    public boolean taskDescriptionContains(int index, String filter) {
```
hellodommy: Would it be possible to use an existing method (eg. `containsIgnoreCase`) instead of defining another method?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,45 @@
...
+    public void execute(Storage storage, Ui ui, TaskList taskList) {
+        ArrayList<Integer> matchIndices = new ArrayList<>();
+
+        for (int index = 1; index <= taskList.getSize(); index++) {
+            if (taskList.taskDescriptionContains(index, filter)) {
+                matchIndices.add(index);
+            }
+        }
+
+        ui.printHorizontalLine();
```
hellodommy: Would it be possible to print the borders once in `run()` in `Duke` so you do not have to call `printHorizontalLine()` for every `execute()`?

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,29 @@
+import java.util.ArrayList;
+
+class DeleteCommand extends Command {
+    int index;
+
+    public DeleteCommand(int index) {
+        this.index = index;
+    }
+
+    public boolean terminate() {
```
hellodommy: Instead of having `terminate()` return `false` for all commands except `ExitCommand`, I had the `ExitCommand` initialised to set `terminate` (a boolean in `Command` abstract class) to `true`. 
Hence only the `ExitCommand` class will refer to `terminate()`.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,53 @@
+package duke.task;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
```
hellodommy: Did not know this existed and went to create a `DateTime` parser on my own 😢
Will try this for mine, thanks!

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,297 @@
...
+
+            if (result.isEmpty()) {
+                throw new RuntimeException("Our parser encountered a fatal error.");
+            }
+
+            return result.get();
+        }
+    }
+
+    private boolean isBeginWithVowel(String string) {
```
hellodommy: Nice attention to detail in using this to ensure grammar is correct 👍🏻 . This idea could also be used for singular and plural words as well.

hellodommy (commented):
Your code looks very thorough and you seem to have handled many different cases of inputs. You do have many good ideas in your code that I will try to implement for mine. 😊

src/main/java/duke/util/TaskList.java
tanlk99: Did not know about that! Thank you for the suggestion.

src/main/java/DeleteCommand.java
tanlk99: I originally envisioned `Command` as an interface, so that was how it was written at the start. However, I agree that your suggestion would make my code cleaner and will implement it.

src/main/java/duke/command/FindCommand.java
tanlk99: I suppose it would be possible if I use a `finally` statement to print the closing line.

src/main/java/duke/util/TaskList.java
tanlk99: On second thought, it seems quite frivolous to include an external library for a single function. Since I want to keep this function in `TaskList` for abstraction purposes, I feel that leaving it as it is should be fine.

ManasVegi:
**General Comments**
- Insane levels of Abstraction. Like, inspiring levels of abstraction. And all the organization into the various packages. My OCD is very happy. I'm a legit fan.
- I really can't find any issues. @hellodommy already mentioned the printHorizontalLine() one, which is not a big issue anyways.
- If i have to nitpick, is that the first user may not need to see the loadingError message. Is it absolutely necessary?
- If I may ask, are you able to run tests using gradle? If yes, what is your project structure?

tanlk99:
> **General Comments**
> 
> * Insane levels of Abstraction. Like, inspiring levels of abstraction. And all the organization into the various packages. My OCD is very happy. I'm a legit fan.
> * I really can't find any issues. @hellodommy already mentioned the printHorizontalLine() one, which is not a big issue anyways.
> * If i have to nitpick, is that the first user may not need to see the loadingError message. Is it absolutely necessary?
> * If I may ask, are you able to run tests using gradle? If yes, what is your project structure?

- Thank you for the generous praise! To be honest, I did have a bit of a sneaky head-start by reading ahead on requirements, so the `MoreOOP` part was not as bad later on.
- The `loadingError` message does not actually display on first startup anymore. I'll remove this part of the message in the next commit.
- I do my testing using Gradle. The `text-ui-test` is unfortunately obselete at this point (probably should update it), so I mainly run unit tests. The project structure looks like this:

   Project Structure 5-9-2019

- To run tests, in the IntelliJ terminal (in the "Terminal" tab along the bottom of the IDE) I run the command

   ```html
   gradlew test
   ```

  This runs all the test classes inside `src/test`. If all your tests pass, you should see something like this:
  
  Gradle Test Success

  Otherwise, there will be some error messages that look like this:

  Gradle Test Fail

  You can try this with a single test file; my `DummyTest.java` file runs assertions on arithmetic.
  Hope this is useful to you!

ManasVegi:
Thanks a lot for your advice!
Can I ask how you implemented your dialog box to extend according to the amount of text in label?

tanlk99:
> Thanks a lot for your advice!
> Can I ask how you implemented your dialog box to extend according to the amount of text in label?

Do you mean the vertical extension of the dialog box? You can achieve that using the JavaFX code given in tutorial Part 3 and 4.

hellodommy:
> Thanks a lot for your advice!
> Can I ask how you implemented your dialog box to extend according to the amount of text in label?

@ManasVegi I added this `dialog.setMinSize(Label.USE_PREF_SIZE, Label.USE_PREF_SIZE);` to my code and my dialog box was able to adjust accordingly! Hope this helps you :)

PR #146, opened 2019-08-26
[Wang Shiyao] Duke Increments

shiyao821:
Underlying git commit history may not match exactly with given requirements


src/main/java/Duke.java
```diff
@@ -1,115 +1,18 @@
...
-        );
-        userInput.clear();
-    }
-
     /**
      * You should have your own function to generate a response to user input.
      * Replace this stub with your completed method.
      */
-    private String getResponse(String input) {
+    String getResponse(String input) {
```
larrylawl: Remember to implement your own function such that Duke works normally :)

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,51 @@
...
+import javafx.scene.control.Button;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+/**
+ * Controller for MainWindow. Provides the layout for the other controls.
+ *
+ * Why do I have to give a shit about JavaFX T.T
```
larrylawl, line 13: +1

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
...
+            tasks = storage.load();
+        } catch (Exception e) {
+            ui.print(e.getMessage());
+            tasks = new TaskList();
+        }
+    }
+
+    private void run() {
+        try (Scanner scanner = new Scanner(System.in);
+             PrintStream out = new PrintStream(System.out, true, StandardCharsets.UTF_8)
```
larrylawl: Oo... may i ask what's print stream used for?

src/main/java/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+
+    void checkValidSpacing(String[] inputSplit) throws InvalidInputFormatException {
+        for (String s : inputSplit) {
+            if (s.length() == 0) {
+                throw new InvalidInputFormatException();
+            }
+        }
+    }
+
+    boolean parse(String input, TaskList tasks, Ui ui, Storage storage) {
```
larrylawl: It seems that your parser handles logic of ui and storage too. May I suggest this alternative POV - 
Parser to only handle the breaking down of the string; ui and storage can be handled in the `run()` method of Duke. This lies more in line with the defn of parsing.

src/main/java/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+     */
+    private int getPrepositionPos(String[] inputSplit) {
+        for (int i = 0; i < inputSplit.length; i++) {
+            if (inputSplit[i].charAt(0) == '/') {
+                return i;
+            }
+        }
+        return 0;
+    }
+
```
larrylawl, line 18: Cool method

src/main/java/Ui.java
```diff
@@ -0,0 +1,72 @@
...
+        System.out.println("  " + t.getStatus());
+        printCount(tasks);
+    }
+
+    void printRemoved(Task removedTask, TaskList tasks) {
+        System.out.println("Noted. I've removed this task: ");
+        System.out.println("  " + removedTask.getStatus());
+        printCount(tasks);
+    }
+
```
larrylawl, line 55: Haha nice.

src/main/java/Ui.java
```diff
@@ -0,0 +1,72 @@
+class Ui {
```
larrylawl, line 1: Very neat!

src/main/java/test3.out
```diff
@@ -0,0 +1,22 @@
+Hello from
```
larrylawl, line 1: Nice test cases

src/main/java/Duke.java
shiyao821: when I used to run it on the command line, the cross and tick characters did't show. A trick I found online is to directly set the PrintStream

src/main/java/Parser.java
shiyao821: Yeah thanks I didn't think of that. Now that I saw your code, returning commands from parse makes it more intuitive. I think I'll do that for the group project.

src/main/java/Parser.java
shiyao821: Thanks! but not sure if you're being sarcastic haha

src/main/java/DialogBox.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        Circle circle = new Circle(40, 40, 40);
+        dialog.setText(text);
+        displayPicture.setImage(img);
+        displayPicture.setClip(circle);
+        dialog.setAlignment(Pos.CENTER_LEFT);
+    }
```
sandydays, line 42: Good job adding these nice Ui touches and making your Duke app really unique!

src/main/java/DukeException.java
```diff
@@ -0,0 +1,23 @@
+class DukeException extends Exception {
+}
+
+class InputUnknownException extends DukeException {
+    @Override
+    public String getMessage() {
+        return "OOPS!!! I'm sorry, but I don't know what that means :-(";
+    }
+}
```
sandydays, line 9: Nice! I like how even under the `DukeException` class you included subclasses that target more specific types of error thrown. Keeps it more clean compared to just modifying the string returned from `getMessage()` within the `DukeException` class itself!

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,70 @@
...
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+import javafx.stage.Window;
+
+/**
+ * Controller for MainWindow. Provides the layout for the other controls.
+ *
+ * Why do I have to give a shit about JavaFX T.T
```
sandydays, line 13: HAHA this is the essence of comments :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
...
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+    private TaskList tasks;
+    private Parser parser;
+    private MainWindow mw;
```
sandydays, line 4: This is a really good idea I never considered - to have the `MainWindow` as an attribute in the `Duke` class. Makes it so much more easier to call methods in `MainWindow` from `Duke`!

src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
...
+        this.mw = mw;
+        try {
+            tasks = Storage.load();
+        } catch (Exception e) {
+            mw.dukeSays(e.getMessage());
+            tasks = new TaskList();
+        }
+        mw.dukeSays(Ui.printLogo());
+        mw.dukeSays(Ui.printHello());
+        mw.dukeSays(Ui.listCommands());
```
sandydays, line 36: I like these additions! Makes it very user-friendly! 

Also, it avoids hardcoding the initial welcome message into the fxml file (something that I did but wasn't too satisfied with), instead calling `dukeSays` from Duke when the user opens the app, which is really nice and clean!

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,70 @@
...
+    void dukeSays(String s) {
+        assert(s instanceof String);
+        dialogContainer.getChildren().add(
+                DialogBox.getDukeDialog(s, dukeImage)
+        );
+    }
+
+    void closeSequence() {
+        Window stage = scrollPane.getScene().getWindow();
+        stage.hide();
```
sandydays, line 54: So this is how we hide the stage if the user sends input `bye`! I was wondering how to close the stage, but I wasn't aware of the `hide` method of `stage` as well! Good job!

Mind sharing how you picked up details on `Window stage = scrollPane.getScene().getWindow();` and `stage.hide()`?

Also, something you could consider (though it really is a stretch) is after the user inputs `bye`, you output bye from Duke, and then after a timeout hide the stage.

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
...
+    void checkValidSpacing(String[] inputSplit) throws InvalidInputFormatException {
+        for (String s : inputSplit) {
+            if (s.length() == 0) {
+                throw new InvalidInputFormatException();
+            }
+        }
+    }
+
+    boolean parse(String input, TaskList tasks, MainWindow mw) {
```
sandydays, line 89: Nice idea of returning a `boolean` in this method to indicate whether to exit the program or not!

src/main/java/Parser.java
```diff
@@ -0,0 +1,156 @@
...
+        String command = inputSplit[0];
+
+        try {
+            switch (command) {
+            case "exit":
+                return false;
+
+            case "bye":
+                Storage.save(tasks.saveFormat());
+                mw.dukeSays(Ui.printBye());
```
sandydays, line 100: I would say avoid passing `MainWindow` instance `mw` and using it in your `Parser` class. You want to keep your Duke bot logic separate from the GUI you have set up with fxml, and using calls to `dukeSays` method in `MainWindow` would only defeat that purpose. Instead, you can link your `Duke` class to `MainWindow`, but any further links that need to be made to `MainWindow` from classes like `Parser` should happen through the `Duke` class and not anywhere else. This ensures that there is only one entry point of the Ui to your duke bot, and that is the `Duke` class.

src/main/java/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+                description.append(" ").append(inputSplit[k]);
+            }
+        } else {
+            for (int k = 1; k < inputSplit.length; k++) {
+                description.append(" ").append(inputSplit[k]);
+            }
+        }
+        for (int i = prepAt + 1; i < inputSplit.length; i++) {
+            memo.append(" ").append(inputSplit[i]);
+        }
```
sandydays, line 47: The only difference between all these `for` loops is that the loop invariant is different or the loop exiting condition is different. But the body of the `for` loops are the same. Isn't there some way you could abstract it out into a `private` method that takes a string to append to, the loop invariant starting, and end, and executes a for loop based on these?

src/main/java/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+            task = new Deadline(description.deleteCharAt(0).toString(), preposition, memo.deleteCharAt(0).toString());
+            break;
+        case "event":
+            task = new Event(description.deleteCharAt(0).toString(), preposition, memo.deleteCharAt(0).toString());
+            break;
+        default:
+            task = null;
+        }
+
+        return task;
```
sandydays, line 63: Instead of assigning variable `task` to a `new ToDo`, `Deadline` or `Event` (or even `null` as is in your code) and returning that `task` at the end once it has been assigned to, couldn't you just return the `new ToDo`, `Deadline` or `Event` within each `case` statement's body? That way you can avoid the `break` statement also since returning something immediately makes the program exit this method. The `task` variable's use seems quite redundant here.

src/main/java/Task.java
```diff
@@ -0,0 +1,119 @@
...
+     * Returns a formatted String that represents the state of this Task
+     * @return String for saving
+     */
+    public String saveFormat() {
+        return (isDone ? "1" : "0") + " " + description;
+    }
+
+}
+
+class ToDo extends Task {
```
sandydays, line 40: I think you can put these classes `ToDo`, `Deadline`, and `Event` in separate java files even if they all extend from `Task`. I believe that is better coding practice (though correct me if I am wrong on this point, but this is how I usually see code arranged).

src/main/java/Task.java
```diff
@@ -0,0 +1,119 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
```
sandydays, line 3: I wasn't aware of these available Java classes for date and time! Good job using them!

src/main/java/TaskList.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return list.remove(i);
+    }
+
+    int size() {
+        return list.size();
+    }
+
+    String printFormat() {
+        StringBuilder sb = new StringBuilder();
```
sandydays, line 36: Nice idea using the `StringBuilder`. Again, I wasn't aware of this class, so used simple string concatenation to put my strings together. 

Though, I don't know, which one would be the better choice? String concatenation by `+` or using StringBuilder's `append`?

src/main/java/Ui.java
```diff
@@ -0,0 +1,79 @@
...
+                "\tlist\n" +
+                "\ttodo <description>\n" +
+                "\tdeadline <description> /by <memo>\n" +
+                "\tevent <description> /at <memo>\n" +
+                "\tdelete <task number>\n" +
+                "\tdone <task number>\n" +
+                "\tbye";
+    }
+
+    static String printFormatHelp() {
```
sandydays, line 56: This is a really nice method to have included! Makes it very easy for user!

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,62 @@
...
+    // public void dummyFailTest() {
+    //    assertEquals("lalala", "placeholder");
+    // }
+
+    private static final ByteArrayOutputStream outContent = new ByteArrayOutputStream();
+    private static final ByteArrayOutputStream errContent = new ByteArrayOutputStream();
+    private static final PrintStream originalOut = System.out;
+    private static final PrintStream originalErr = System.err;
+
+    @BeforeAll
```
sandydays, line 21: Nice job using some extra features of JUnit. I wasn't aware of `@BeforeAll` and `@AfterAll`! What do they do?

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,62 @@
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DukeTest {
```
sandydays, line 10: Nice tests, but wouldn't it be good if you split your tests into separate test classes based on which method of which class is being tested? For example, the `uiHelloTest` could go into a class called `UiTest` that only deals with testing methods in your `Ui.java`, etc.

sandydays (commented):
Overall, I think you have done a really good job on the Duke bot! I ran your code, and the interface is very presentable and easy to use! There were some things I liked about your program a lot, mainly the extra effort put into making the UI better, getting `Duke` to call methods in the `MainWindow`. I have a few suggestions, however, that you can take a look at in the detailed comments (by line). Hope this helps make your bot better!

PR #147, opened 2019-08-27
[Hema] Duke Increments

lrchema:


src/main/java/TaskList.java
```diff
@@ -0,0 +1,67 @@
...
+     * fetches the task from the list
+     * 
+     * @param i index of the task mentioned
+     * @return the task with the index i
+     */
+    public Task get(int i) {
+        return list.get(i);
+    }
+    /**
+     *
```
austinsantoso: The Javadoc in this section is missing a description of what the function does.

austinsantoso (commented):
Overall good coding practice and styling. But i do think that it would be best to separate the duke outputs, to make it more visually appealing

src/main/java/TaskList.java
lrchema: I thought the function name, param and return statement made it quite self-explanatory, but I will add that in.

lrchema:
> Overall good coding practice and styling. But i do think that it would be best to separate the duke outputs, to make it more visually appealing

Yes, I saw you had done that in your PR. You're right, I should do it as well :)

src/main/java/Verify.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * 
+     * @param command   the string command entered by user
+     * @param list      the task list to compare with for checking validity
+     * @param formatter date formatter to catch wrong date format exceptions
+     * @throws DukeException in the case of invalid input
+     */
+    public static String checkCommandValidity(String command, TaskList list, DateTimeFormatter formatter)
```
yuhongtay: Would it be easier to test if this method were broken into smaller methods (e.g.  method to test the funtionality of todo method) and this method just called them using a switch case?

yuhongtay (commented):
Code has good styling and is well documented. According to CheckStyle, documentation needs to end with a punctuation mark, which is missing here and there, but this might be excessive. (Don't kill me) Would unit testing targeting individual classes be safer in the long run?

src/main/java/Verify.java
lrchema: yes, good idea. it would probably be simpler for both testing and reading if i split up the huge if-else block

lrchema:
> Code has good styling and is well documented. According to CheckStyle, documentation needs to end with a punctuation mark, which is missing here and there, but this might be excessive. (Don't kill me) Would unit testing targeting individual classes be safer in the long run?

thanks for the feedback! yes, unit testing would probably be wise :)

PR #149, opened 2019-08-27
[Chen Kai Bin] Duke Increments

ckb055:
[Chen Kai Bin] Duke Increments in Week 3


calvincxz:
Hi, I think the level-6 code is generally well done. For the OOP part, maybe you are missing a Parser class to shorten the code in the Duke class? Also, you can have a Command parent class with different child Commands such that the Parser returns a Command object after reading the user input. :)

meiannn (commented):
Hi I think your code looks pretty good! you implemented it following exactly what was stated in the requirements so I think kudos to you. Maybe you can add in documentation to make the code clearer for others reading it :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,52 @@
+import java.text.ParseException;
+import java.util.Date;
+import java.text.SimpleDateFormat;
+
+public class Deadline extends Task {
+    private Date date;
+    private String dateToStorage;
```
dvrylc: This variable name might require further explanation, consider changing this to a more obvious one

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Parses the sentence entered by the user and returns various commands
+     * @param input by the user
+     * @return other commands
+     */
+    // parse the command string keyed in by the user
+    public static Command parse (String input) throws DukeException {
+        String[] userInput = input.split(" ", 2);
+        String firstWord = userInput[0]; //firstword of the user input
+        //try {
```
dvrylc: Redundant comment 🤔

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+                    return new DoneCommand(indexDone);
+                case "delete":
+                    int indexDelete = Integer.parseInt(userInput[1]);
+                    return new DeleteCommand(indexDelete);
+                case "massDelete":
+                    String[] indexRanges = userInput[1].split(" ", 2);
+                    int indexStart = Integer.parseInt(indexRanges[0]);
+                    int indexEnd = Integer.parseInt(indexRanges[1]);
+                    for (int i = 0; i < (indexEnd - indexStart); i++) {
+                        return new DeleteCommand(indexStart + i);
```
dvrylc: This for loop will have at most 1 iteration, is this the intended behavior? 🥇

dvrylc (commented):
- Inconsistent coding style, verify with checkstyle
- Missing exception handling (invalid ID for delete command etc)
- `*` imports violate checkstyle

src/main/java/Parser.java
ckb055: whoops, I haven't implemented that. sorry!

PR #150, opened 2019-08-27
[Lim Yan Peng Gary] Duke Increments

garylyp:


src/main/java/command/CommandCentre.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.HashMap;
+
+public class CommandCentre {
+
+    /**
+     * A collection to store and map command names to their
+     * respective Command.
+     */
+    private HashMap<String, Command> commands;
```
ShirleyWangxt, line 14: I like how you use hashmap to represent commands which improves the efficiency of access.

src/main/java/main/Duke.java
```diff
@@ -0,0 +1,183 @@
...
+import utils.Storage;
+import utils.Ui;
+
+import java.util.List;
+import java.util.Scanner;
+
+
+public class Duke {
+
+    private static final String ROOT = "D:\\Gary\\Uni\\NUS\\1920SEM1\\CS2103T\\Practices\\duke";
```
ShirleyWangxt: Nice use of final constants.

src/main/java/exception/EmptyDescriptionException.java
```diff
@@ -0,0 +1,7 @@
+package exception;
+
+public class EmptyDescriptionException extends DukeException {
```
ShirleyWangxt, line 3: It is good that you use different classes to represent specific exceptions.

src/main/java/utils/Storage.java
```diff
@@ -0,0 +1,139 @@
...
+     * contents.
+     */
+    public void updateData() {
+        try {
+            FileWriter fw = new FileWriter(file.getAbsolutePath());
+            TaskList taskList = TaskList.newInstance();
+            StringBuilder sb = new StringBuilder();
+            for (int i = 0; i < taskList.size(); i++) {
+                Task task = taskList.get(i);
+                if (task instanceof Todo) {
```
ShirleyWangxt: For each task instance, the steps to update data seems a bit long-winded and repetitive. Would it be better to construct the format in the respective subclasses instead of in the update method itself?

src/test/java/utils/ParserTest.java
```diff
@@ -0,0 +1,110 @@
+package utils;
```
ShirleyWangxt, line 1: Your test classes seem pretty complete and specific. Nice job!

src/main/java/utils/Ui.java
```diff
@@ -0,0 +1,162 @@
...
+        outputBuilder.append(DIVIDER);
+        outputBuilder.append("Hello! I'm Duke\n");
+        outputBuilder.append("What can I do for you?\n");
+        outputBuilder.append(DIVIDER);
+    }
+
+    /**
+     * Prints a list of actions that can be used.
+     */
+    public void printHelpMessage() {
```
ShirleyWangxt, line 80: I like the help message function a lot! It serves a good purpose of introducing and assisting users of the functionality.

src/main/java/utils/Storage.java
garylyp: Good idea! I can create a method in Task for the formatting of the storage string so that the process is abstracted away from the Storage method. Thank you!

ReignOfComputer:
Looks good to me.

PR #151, opened 2019-08-27
[Liang Boyuan] Duke Increments

Oscar-B-Liang:
This is the version where all requirements of the iP can be satisfied. At least I do think so.


src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,72 @@
+package duke.commands;
```
jiayushe, line 1: I think it is preferred to use singular form for package naming, i.e. `package duke.command;`.

src/main/java/duke/directprocessor/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+public class TaskList {
+
+    private ArrayList<Task> taskList;
+
+    /**
+     * The constructor of the class to generate an empty task list.
+     * Note that this constructor will be applied only when it fails to read from the file.
+     */
+    public TaskList() {
+        taskList = new ArrayList<>();
```
jiayushe, line 25: I think it is preferred to include the generic type, i.e. `new ArrayList<Task>()`.

src/main/java/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
```
jiayushe, line 1: This file is irrelevant to the project. You can add it to `.gitignore`.

jiayushe (commented):
Generally well-written code and good test coverage. Some minor issues with package naming and coding style as pointed out in my comments. Otherwise look good to me!

PR #152, opened 2019-08-27
[Lim Kang Yee] Duke Increments

kangyeelim:


src/main/java/seedu/duke/ByeCommand.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Returns true to end Duke.
+     * Prints goodbye message before returning.
+     *
+     * @return True to end Duke
+     */
+    public boolean isExit() {
+        ui.printGoodbyeMsg();
```
cheongsiuhong: Just my personal opinion, but I believe this function should purely return the boolean and not execute other irrelevant code. Perhaps a better place to call the print function is in `execute`.

src/main/java/seedu/duke/Command.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Executes command and is the parent class of all duke commands.
+ * Abstract class with no method implementations.
+ */
+public abstract class Command {
+    /**
+     * Class constructor.
+     */
+    public Command() {
```
cheongsiuhong, line 11: I believe this is not needed, but correct me if I'm wrong

src/main/java/seedu/duke/FindCommand.java
```diff
@@ -0,0 +1,85 @@
...
+                Deadline dl = (Deadline) tasks.get(i);
+                if (dl.getDescription().contains(keyword)) {
+                    matchList.add(dl);
+                } else if (outDateFormat.format(dl.getDate()).contains(keyword)) {
+                    matchList.add(dl);
+                } else if (outTimeFormat.format(dl.getTime()).contains(keyword)) {
+                    matchList.add(dl);
+                }
+            }
+        }
```
cheongsiuhong, line 46: Perhaps it may be better to store the initial string arguments in all the tasks and search that string instead? It would remove the need for type checking here.

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,287 @@
...
+            return new MarkDoneCommand(command);
+        } else if (Parser.isListCommand(command)) {
+            //listing tasks out
+            return new ListCommand(command);
+        } else if (Parser.isFindCommand(command)) {
+            return new FindCommand(command);
+        } else {
+            throw new DukeException(ui.noSuchCommand());
+        }
+    }
```
cheongsiuhong, line 285: Maybe try using switch and case? This looks a little complicated.

src/main/java/seedu/duke/Ui.java
```diff
@@ -0,0 +1,380 @@
...
+        if (tasks.size() == 0) {
+            //check if list has no task to throw exception
+            throw new DukeException("\u2639 OOPS!!! You do not have any tasks in your list.\n");
+        } else if (curr > tasks.size()) {
+            //check if index is within list size or throw exception
+            throw new DukeException("\u2639 OOPS!!! You do not have that task in your list. "
+                    + "Call 'list' to see all your tasks :-)\n");
+        }
+    }
+
```
cheongsiuhong, line 119: Shouldn't the UI file handle only input and output? I think that perhaps error checking should occur during parsing, or be in the same class of the specific command itself.

cheongsiuhong (commented):
Other than my few nitpickings, the code looks pretty well written. The JUnit tests are pretty extensive, nice!

src/main/java/seedu/duke/ByeCommand.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Returns type of command.
+     *
+     * @return String of command type.
+     */
+    @Override
+    public String toString() {
+        //for testing purposes
+        return "ByeCommand";
```
lyskevin, line 54: You can consider moving the "ByeCommand" String into a class-level constant so that it can be changed and used easily if you decide to include additional methods which use it.

src/main/java/seedu/duke/Deadline.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Returns string representation of the deadline task.
+     *
+     * @return String with the type of [task] [tick or cross] description (by:date time).
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + outDateFormat.format(date) + " "
```
lyskevin, line 57: The String.format() method might be useful here so that you don't have to concatenate so many strings together :)

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,58 @@
...
+        } catch (DukeException e) {
+            return e.toString();
+        } catch (ParseException e) {
+            return ui.showParseError();
+        } catch (ArrayIndexOutOfBoundsException e) {
+            return ui.showParseError();
+        } catch (Exception e) {
+            return ui.showExceptionMsg(e);
+        }
+    }
```
lyskevin, line 54: Nice use of abstraction here!

src/main/java/seedu/duke/DukeException.java
```diff
@@ -0,0 +1,35 @@
...
+    public DukeException(String errMsg) {
+        super(errMsg);
+        this.errMsg = errMsg;
+    }
+
+    /**
+     * Class constructor.
+     */
+    public DukeException() {
+    }
```
lyskevin, line 24: You can consider removing constructors with empty method bodies since they are already inherited from the Object class by default

src/main/java/seedu/duke/Parser.java
lyskevin: I agree here. You can consider using switch statements with additional helper methods to abstract away the nested if-else conditionals.

lyskevin (commented):
Hi Kang Yee, I think your code is generally pretty clean and definitely quite robust given the number of unit tests and exception handling methods that it contains. I just have 3 main points that you can consider:

1. You can split some of your classes into smaller sub-packages (e.g. put Parser into duke.parser)
2. Constants are helpful in cases where you want to return values that are unlikely to change at the class level (e.g. in the toString() methods of some Command subclasses)
3. I find that String.format() comes in very handy for dealing with long strings that require concatenation with a lot of variables

Other than that I think your code is well-written and I enjoyed reading through it :)

PR #153, opened 2019-08-27
[Hon Hao Chen] Duke Increments

honhaochen:


src/main/java/command/ListCommand.java
```diff
@@ -16,15 +16,17 @@
      * @param ui the User Interface which responsible for every output printing.
      * @param storage user's hard disk storage.
      */
+    @Override
     public void execute(TaskList tasks, Ui ui, Storage storage) {
-        System.out.println("     Here are the tasks in your list:");
+        ui.println("Here are the tasks in your list:");
         int i = 1;
         for (Task task : tasks.getTaskList()) {
```
tcgj: Would it be a bad idea to return the entire task list in the form of a `List<Task>`?

I don't think it's a very good idea  to expose your implementation of `TaskList` since you already encapsulated it. Personally I implemented the `Iterable` interface in my tasklist class, so I can use the enhanced for loop directly on a tasklist.

src/main/java/command/DeadlineCommand.java
```diff
@@ -43,19 +44,20 @@ public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeParseExce
                 String time = input[input.length - 2] + " " + input[input.length - 1];
                 Task newDeadlineTask = new Deadline(description, time);
                 tasks.getTaskList().add(newDeadlineTask);
-                System.out.println("     Got it. I've added this task:");
-                System.out.println("       " + newDeadlineTask);
+                ui.println("Got it. I've added this task:");
+                ui.println("  " + newDeadlineTask);
                 if (tasks.getTaskList().size() > 1) {
-                    System.out.println("     Now you have " + tasks.getTaskList().size() + " tasks in the list.");
+                    ui.println("Now you have " + tasks.getTaskList().size() + " tasks in the list.");
```
tcgj: Would it be better if you had a method inside your `TaskList` that returned the number of tasks, rather than getting the size of the `List` that is encapsulated within your `TaskList` class?

src/main/java/Parser.java
```diff
@@ -14,27 +14,28 @@
...
+     * input and react with a suitable command.
+     * 
      * @param userInput users's input as String
      * @return a Command
      * @See Command
      */
     public static Command parse(String userInput) {
         String[] input = userInput.split(" ");
-        if (input[0].equals("list")) { //list out
+        if (input[0].equals("list")) { // list out
```
tcgj: Hmm, would it be better if you wrote a switch case here instead of a set of if elses? That way you don't actually need to keep checking using `.equals(str)`.

src/main/java/task/ToDo.java
```diff
@@ -1,19 +1,31 @@
...
     public ToDo(String isDone, String description) {
         super(description);
         if (isDone.equals("\u2713")) {
             this.isDone = true;
         }
     }
 
-   public String toDataBase() {
+    /**
+     * Used to store into hard disk storage for easy reading.
```
tcgj: Will this violate coding standards?

Just an observation from most of the javadoc comments. I think you start most of them declaratively rather than imperatively. Perhaps you might want to start with "Stores into hard disk....."?

src/main/java/command/FindCommand.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * execute command for finding query in user's tasks list.
+     * @param tasks a TaskList.
+     * @param ui User Interface for output printing.
+     * @param storage hard disk storage.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        if (input.length <= 1) {
+            System.err.println("     " + new FindException());
```
tcgj: Hmm would it be better if you handled your exceptions in one place rather than just printing the exception itself in every command?

tcgj (commented):
Hey I like how you've written your code mostly! It looks pretty clean, and it looks like you've already gone through and corrected most of the coding standard violations already! Sweet.

src/main/java/Parser.java
honhaochen: Actually I think switch case is the same logic as if-else, and I personally prefer the if-else statements hm..

src/main/java/task/ToDo.java
honhaochen: Noted, thanks!

aang114 (changes requested):
Nice code! I especially liked your idea of using different types of exceptions rather than just DukeException! However, it would have been even better if all the exceptions you used inherited from DukeException.

src/main/java/Duke.java
```diff
@@ -1,10 +1,26 @@
...
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String horizontalLine = "    ____________________________________________________________";
+
+        System.out.println(horizontalLine);
+        System.out.println("     Hello from! I'm Duke\n" + "     What can I do for you?");
+        System.out.println(horizontalLine + "\n");
+
+        Scanner sc = new Scanner(System.in);
```
eizon05: You forgot to close your scanner at the end!

src/main/java/Duke.java
```diff
@@ -1,10 +1,57 @@
...
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * main run method.
```
aang114: It is recomended to start your Javadoc comment with a verb

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+                    task.markAsDone();
+                    ui.println("Nice! I've marked this task as done:");
+                    ui.println("  " + task);
+                }
+            }
+        }
+    }
+
+    @Override
+    public boolean isExit() {
```
aang114: Instead of having an abstract isExit method, you could make it return false in the Command class and only override it in the ExitCommand/ByeCommand class

src/main/java/command/ToDoCommand.java
```diff
@@ -0,0 +1,53 @@
...
+import storage.Storage;
+import task.Task;
+import task.TaskList;
+import task.ToDo;
+import ui.Ui;
+
+/**
+ * Command for creating a ToDo Task.
+ */
+public class ToDoCommand extends Command {
```
aang114, line 14: maybe you can rename this class as 'TodoCommand' to follow the naming convention?

text-ui-test/ACTUAL.TXT
```diff
@@ -0,0 +1,51 @@
+    ____________________________________________________________
```
aang114, line 1: VERY GOOD TEST FILE!!!

src/main/java/command/Command.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Abstract class for every Commands.
+ * Encapsulates user's input into a command for easy handling.
+ */
+public abstract class Command {
+
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    public abstract boolean isExit();
```
aang114: It might be more efficient to make this class non-abstract and return false. Then in your ExitCommand/ByeCommand, you can override it and make it return true.

src/main/java/command/DeadlineCommand.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param tasks the TaskList.
+     * @param ui the User Interface which responsible for every output printing.
+     * @param storage user's hard disk storage.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeParseException, DeadlineException {
+        if (input.length <= 1) {
+            throw new DeadlineException();
+        } else if (input.length <=4 ) {
+            throw new DeadlineException("OOPS!!! Did you type something wrong?\n"
```
aang114, line 38: very good that you have different error messages for the different types of errors!

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Execute 'ticking' a task in user's tasks list.
+     * Output what is needed.
+     * @param tasks the TaskList.
+     * @param ui the User Interface which responsible for every output printing.
+     * @param storage user's hard disk storage.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DoneException {
```
aang114: It seems inefficient to parse the input in the execute method for the Command class. Wouldn't it be cleaner and more efficient to parse it in entirely in the Parser first class and then have a class variable of type 'Task' in the DoneCommand? (rather than having a class variable of type String[] - which you still have to parse!)

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+import task.Task;
+import task.TaskList;
+import ui.Ui;
+
+/**
+ * Command for specifying that a task is done.
+ */
+public class DoneCommand extends Command {
+
+    protected String[] input;
```
aang114, line 15: I feel it would be better if you parsed it entirely and then have a class variable of type Task which represents the task that was done?

aang114 (commented):
Overall your code is quite solid. It is very easy to read and follows most OOP and coding standards. However, I have pointed out a few suggestions which I think could make your code a little more efficient. Kudos!

src/main/java/command/ListCommand.java
honhaochen: Alright! thanks for your suggestion!

PR #156, opened 2019-08-27
[Caleb Goh Ee Gen] Duke Increments

heze8:


.gitignore
```diff
@@ -11,4 +11,4 @@ src/main/resources/docs/
 # MacOS custom attributes files created by Finder
 .DS_Store
 *.iml
-bin/
```
CoderStellaJ: Personally, I think it would be better to keep bin/ ignored from git. as it is MacOS custom attributes files created by Finder. It probably would be better to exclude these files and focus on source code for the project so that your repository is "cleaner" and easier to be managed. Otherwise, the number of changed files is huge and harder for mainenance.

CoderStellaJ, line 14: Personally, I think it would be better to keep bin/ ignored from git. as it is MacOS custom attributes files created by Finder. It probably would be better to exclude these files and focus on source code for the project so that your repository is "cleaner" and easier to be managed. Otherwise, the number of changed files is huge and harder for mainenance.

docs/DukeTest.html
```diff
@@ -0,0 +1,305 @@
+<!DOCTYPE HTML>
```
CoderStellaJ: May I know how you generated these files inside docs folder? I also checked other people's PR but the docs folder are all almost empty.

src/main/java/duke/bin/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+import duke.bin.common.DukeException;
+import duke.bin.task.Task;
+
+import java.util.ArrayList;
+
+/**
+ * Contains the list of tasks with operations to add, delete, etc.
+ */
+public class TaskList {
+    private ArrayList<Task> textStorage = new ArrayList<>();
```
CoderStellaJ: It's better to put the initialization part into the constructor

src/main/java/duke/bin/UI.java
```diff
@@ -0,0 +1,97 @@
...
+    public void showError(DukeException exception) {
+        display(exception.getMessage());
+    }
+
+    /**
+     * Reads what the user has inputted.
+     *
+     * @return the input in String format
+     */
+    public String readCommand() {
```
CoderStellaJ: nextLine() function may throw exception when no line is found. So exception handling is recommended here.

src/main/java/duke/bin/UI.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Displays the message to the user. Overloaded for start and end phrase to be displayed.
+     *
+     * @param string the message to be displayed.
+     * @param startPhrase the start message to be displayed before the original message.
+     * @param endingPhrase the end message to be displayed after the original message.
+     */
+    public void display(String string, String startPhrase, String endingPhrase) {
```
CoderStellaJ: Lots of function overloading here. On CS2103T page, inside run function,  ui can call showline() to printout line. So you can try to extract more common parts from the display functions and create another function to reduce duplicates.

src/main/java/duke/bin/common/Constants.java
```diff
@@ -0,0 +1,9 @@
+package duke.bin.common;
+
+/**
+ * Constants used regularly by the UI.
+ */
+public class Constants {
```
CoderStellaJ: Is it better to come up with another name because Constants is pretty misleading?

src/main/java/duke/bin/common/Time.java
```diff
@@ -0,0 +1,50 @@
+package duke.bin.common;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Time class to hold value of time with correct display.
+ */
+public class Time {
```
CoderStellaJ: I like this class. It's pretty useful for Duke to understand time :)

src/main/java/duke/bin/datahandling/DataStorage.java
```diff
@@ -0,0 +1,118 @@
+package duke.bin.datahandling;
+
+import duke.bin.common.DukeException;
+import duke.bin.common.Time;
+import duke.bin.task.Deadline;
+import duke.bin.task.Event;
+import duke.bin.task.Task;
+import duke.bin.task.ToDo;
+
+import java.io.*;
```
CoderStellaJ: wildcard import violates coding standard.

CoderStellaJ (commented):
In general, I think the code quality is pretty good. And the logic is well implemented, nice try!
I have left some comments in your code, if you have any doubts or disagreements, I would like to hear your opinion as well.

docs/DukeTest.html
heze8: I used Idea Intelijj's inbuild Javadoc tools to create it. It's not updated though.

src/main/java/duke/bin/TaskList.java
heze8: Good catch

src/main/java/duke/bin/common/Constants.java
heze8: I see, okay removed the class

src/main/java/duke/bin/common/Time.java
heze8: Thanks!

PR #158, opened 2019-08-27
[Ge Shuming] Duke Increments

geshuming:


src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,7 @@
+package duke.task;
+
+import java.util.ArrayList;
+
+public class TaskList extends ArrayList<Task> {
+    private static final long serialVersionUID = 1L;
```
irene-lly, line 7: Using SERIAL_VERSION_UID here would be better because constant names should be in all uppercase letters using underscore to separate words.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,32 @@
+package duke.task;
+
+public abstract class Task {
+    public String description;
```
irene-lly: According to the java coding standards we should try not to declare class variables as public.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+     * 
+     * @param input a String object.
+     * @return a Duke Command.
+     */
+    public static Command parse(String input) {
+        String[] inputs = input.trim().split(" ", 2);
+        String command = inputs[0];
+
+        switch (command) {
+        case "todo":
```
irene-lly, line 28: Seeing as how the execution of the todo, deadline and event tasks have quite a few similarities, maybe you can try extracting it out into another method?

irene-lly (commented):
Overall I think you did a great job dividing and packaging your classes, your OOP is pretty good. I would suggest adding header comments to all your classes too especially if they are public. 

(btw I like how you completely remove all the unicode XD)

src/main/java/duke/task/TaskList.java
geshuming: [Because exception implements Serializable, the serialVersionUID should be declared as it is.](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html)

irene-lly: Oh I see! Sorry about that I did not know about the Serializable interface.

src/main/java/duke/parser/Parser.java
```diff
@@ -132,21 +134,24 @@ public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException
...
                 }
             };
         case "bye":
             return new Command() {
-                public void execute(TaskList tasks, Ui ui, Storage storage) {
+                public String execute(TaskList tasks, Storage storage) {
+                    return "Bye. Hope to see you again soon!";
                 }
 
                 public boolean isExit() {
```
mirozo, line 175: Minor change: isExit() to isExited()

src/main/java/duke/parser/Parser.java
```diff
@@ -155,16 +160,15 @@ public boolean isExit() {
...
             return new Command() {
-                public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+                public String execute(TaskList tasks, Storage storage) throws DukeException {
                     throw new DukeException("Oops! You entered an invalid command.");
                 }
             };
         }
     }
 
-    private static void echoTaskAdded(Task task, TaskList tasks, Ui ui) {
```
mirozo: Good use of functions to improve modularity and readability.

mirozo:
Impressive work. Great use of interfaces and classes for abstraction. Can still improve some of your naming conventions, just a suggestion.

PR #160, opened 2019-08-27
[Fabian Chia] Duke Increments

fabbbbbbyy:
Pull request towards upstream repo.


src/main/java/DummyCommand.java
```diff
@@ -0,0 +1,18 @@
+/**
+ * Simulates a dummy command of the Duke System.
+ * This is to be used to signify an error.
+ * @author Fabian Chia Hup Peng
+ */
+
+public class DummyCommand extends Command {
```
tysg, line 7: Could it be better to use exceptions to signify errors?

src/main/java/Parser.java
```diff
@@ -0,0 +1,206 @@
...
+public class Parser {
+
+    /**
+     * Parsers user input into appropriate commands.
+     * @param fullCommand The user input in String format.
+     * @return The command to be carried out by Duke.
+     */
+    public static Command parse(String fullCommand) {
+
+        Command returnCommand = new DummyCommand();
```
tysg, line 18: What about just declaring the `returnCommand` variable here without instantiating the `DummyCommand`? Then you can throw an exception in the `else` case if `returnCommand` is not assigned to anything. In this way, you can get rid of the dummy class.

src/main/java/Parser.java
```diff
@@ -0,0 +1,206 @@
...
+    private static boolean isBye(String str) {
+        return str.equals("bye") || str.equals("Bye");
+    }
+
+    private static boolean isList(String str) {
+        return str.equals("list");
+    }
+
+    private static boolean isDone(String str) {
+        return str.startsWith("done");
```
tysg, line 192: I feel that it might be better to check if `str.split(" ")[0].equals("done")`, as the current implementation my falsely accept commands like `doneextrawords`, which I don't feel it should.

src/main/java/Storage.java
```diff
@@ -0,0 +1,171 @@
...
+
+            int counter = 5;
+            String taskDescription = splitString[4];
+
+            for(int i = 5; i < splitString.length && !splitString[i].equals("|"); i++) {
+                taskDescription += " " + splitString[i];
+                counter++;
+            }
+
+            String dateTime = splitString[++counter];
```
tysg, line 143: Just a personal suggestion, maybe it's better to not use increment operators within the array index for better readability. But again this is just style preference. (see https://stackoverflow.com/questions/971312/why-avoid-increment-and-decrement-operators-in-javascript)

tysg (commented):
Hi Fabian, most of the code LGTM. You abstracted the Ui, Storage and Duke layers well and the code is very easy to read. I have made some minor suggestions for your consideration. Thanks!

src/main/java/DateTime.java
```diff
@@ -0,0 +1,105 @@
...
+        this.month = month;
+        this.year = year;
+        this.time = time;
+
+    }
+
+    /**
+     * Formats the user input into proper representation.
+     */
+    public void format() {
```
kohyida1997, line 36: Nice that you abstract the formatting of the DateTime!

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     * and prints the deleted message.
+     * @param taskList The task list for the task to be added to.
+     * @param ui The ui which prints the added message.
+     * @param storage The storage which deals with the hard drive.
+     * @return The string representation of a successful delete command.
+     */
+    public String execute(TaskList taskList, Ui ui, Storage storage) {
+        ArrayList<Task> arrayList = taskList.getArrayList();
+        Task task = arrayList.get(taskNumber - 1);
+        taskList.deleteTask(task);
```
kohyida1997, line 31: The implementation is solid! Perhaps you can consider abstracting the **retrieval of task to delete from the TaskList**, such that you won't need to expose the underlying ArrayList implementation.

src/main/java/DummyCommand.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Simulates a dummy command of the Duke System.
+ * This is to be used to signify an error.
```
kohyida1997, line 3: Understand where you are coming from with the use of Dummy commands, but perhaps it is more appropriate to handle the error directly using exceptions!

src/main/java/HelpCommand.java
```diff
@@ -0,0 +1,19 @@
+/**
+ * Simulates a help command of the Duke System.
+ * @author Fabian Chia Hup Peng
+ */
+
+public class HelpCommand extends Command {
+
+    /**
+     * Executes the help command; shows all possible Duke commands.
```
kohyida1997, line 9: Having this Help command feature is great- It teaches the user how to use your program!

src/main/java/Parser.java
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Parsers user input into appropriate commands.
+     * @param fullCommand The user input in String format.
+     * @return The command to be carried out by Duke.
+     */
+    public static Command parse(String fullCommand) {
+
+        Command returnCommand = new DummyCommand();
+
+        if(isBye(fullCommand)) {
```
kohyida1997, line 20: Perhaps it may be cleaner to use a switch-case instead of many if-elses ?

kohyida1997 (commented):
Your code is generally clean and well organized! Gave some minor suggestions about implementation, mainly to improve abstraction.

PR #162, opened 2019-08-27
[Bernice] Duke Increments

bernicechio:


CowSaysBaa:
The code is well written and very readable and easy to understand. The format is consistent and well implemented. I'm not sure the @Override annotation is required for the implementation of an abstract method.

briyanii:
JavaDoc comments were very clear

I like how tested for all the different inputs that should give corrects outputs, but it would be good to test for incorrect inputs as well since it is not safe to assume that the user uses your program correctly.

Nice use of existing API to parse date and time input from user.

Hard coded file path did not work when using the program on my computer. I recommend taking a look at this link:
 https://howtodoinjava.com/java/io/read-file-from-resources-folder/

CowSaysBaa (commented):
The code is well written and very readable and easy to understand. The format is consistent and well implemented.

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns true if two instances of AddCommand are equal.
+     * Otherwise, returns false.
+     *
+     * @param o  An object.
+     * @return boolean.
+     */
+    @Override
+    public boolean equals(Object o) {
```
lumos309: Argument name could be more descriptive (e.g. `object`).

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Executes the command, accesses the task list.
+     * User interface prints out message to user.
+     * Throws exception if task is not found.
+     *
+     * @param tasks Contains the task list .
+     * @param ui Deals with interactions with the user.
+     * @param storage Deals with loading tasks from the file and saving tasks in the file.
+     */
+    public abstract String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
```
lumos309, line 19: Javadoc comment should include the return type and description.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Represents a Deadline object. A <code>Deadline</code> object corresponds to
+ * an agenda to be done by.
+ */
+public class Deadline extends Task {
+    private Date date;
+
+    public Deadline(String description) {
+        super(description);
+        this.type = "D";
```
lumos309: Should be an enum instead of a "magic string"; alternatively, declare a `final static String deadlineType = "D"`.

PR #164, opened 2019-08-27
[Felix Ng] Duke Increments

FelixNWJ:


src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,234 @@
...
+    private String details;
+
+    /**
+     * Instantiates a TodoCommand Object
+     *
+     * @param details the details of the task to be stored in the TodoCommand Object
+     * @return a TodoCommand Object
+     * @throws DukeEmptyDescriptionException when details is an empty string.
+     */
+    public static Command addTodoCommand(String details) throws DukeEmptyDescriptionException {
```
ccyccyccy, line 36: There's a cyclic dependency between the `Command` class and its children class by including these static methods here. You might consider putting these methods into another class like `CommandFactory`. Same for the other static methods below.

src/main/java/duke/command/DeadlineCommand.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Adds the Deadline Object to the TaskList of the Duke Object.
+     * Stores the Deadline in the StorageData and prints out a message to confirm that it has been added.
+     * @param tasks TaskList of Duke Object
+     * @param ui DukeUI of Duke Object
+     * @param storage StorageData of Duke Object
+     */
+    public String execute(TaskList tasks, DukeUi ui, StorageData storage) {
```
ccyccyccy, line 56: Might want to add a `@Override` tag here

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,38 @@
...
+        this.taskNumber = Integer.valueOf(details);
+    }
+
+    /**
+     * Deletes the task number from the TaskList in the Duke Object, then erases the data from the file in StorageData.
+     * @param tasks TaskList of Duke Object
+     * @param ui DukeUI of Duke Object
+     * @param storage StorageData of Duke Object
+     */
+    public String execute(TaskList tasks, DukeUi ui, StorageData storage) {
```
ccyccyccy, line 34: Similarly might want to add a `@Override` tag here, and other places as well.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,234 @@
...
+     */
+    public Command() {
+        this.details = "";
+    }
+
+    public String getDetails() {
+        return this.details;
+    }
+
+    /**
```
ccyccyccy, line 148: I don't think date time utility methods belong in the `Command` class? Might want to consider creating a utility class for date time conversion, or use a library like the `java.time` package.

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,28 @@
...
+
+import duke.ui.DukeUi;
+
+import duke.storagedata.StorageData;
+
+/**
+ * Represents a Command object where the user input is "bye.
+ */
+public class ByeCommand extends Command {
+    public ByeCommand() {
```
ccyccyccy, line 13: I don't think it's necessary to include the `super()` call in the constructor, especially when there is nothing else in the constructor. The `super()` is called by [default](https://stackoverflow.com/questions/2054022/is-it-unnecessary-to-put-super-in-constructor)

src/main/java/duke/storagedata/StorageData.java
```diff
@@ -0,0 +1,214 @@
+package duke.storagedata;
```
ccyccyccy, line 1: Writing individually to the file for each command seems quite tedious and hard to maintain. Every single new command requires a new method to be defined.
You might want to implement a `save(TaskList tl)` function which can be used for all kind of commands, and reduce the complexity in this class.

src/main/java/duke/ui/DukeUi.java
```diff
@@ -0,0 +1,90 @@
...
+    public void printWelcomeMessage() {
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+    }
+
+    public String printByeMessage() {
```
ccyccyccy: Might want to change the `printByeMessage()` to `getByeMessage()` to be more clear on what the method is doing. Similarly for the methods below.

ccyccyccy (commented):
Overall good job in separating the logic into different parts.
I like the way that you packaged the different type of classes into their respective packages.
Also, the error handling was done excellently as it was clear what kind of error occured.

PR #165, opened 2019-08-27
[Jeong Yu Han] Duke Increments

ghost:
my PR request for Duke Project


src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Represents an error in the program. A <code>DukeException</code> object corresponds to
+ * an error while running the program, such as missing files and poor user input.
+ */
+public class DukeException extends Exception {
+    private String errorMsg;
+
+    public DukeException(String errorMsg) {
+        super(errorMsg);
+        this.errorMsg = errorMsg;
```
chaitanyabaranwal, line 12: I am not sure on what purpose `this.errorMsg = errorMsg` actually serve. Everywhere I saw DukeException was represented with the `getMessage()` function, which is why I think another variable called `errorMsg` and the `toString()` method is not needed. Just inheriting from `Exception` should be enough. Do let me know if I missed something!

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,66 @@
...
+     * Returns appropriate Command object as specified by the user.
+     * Flags error when user command is invalid or error occurs.
+     *
+     * @param command  The input as specified by the user.
+     * @throws DukeException  Error while executing command.
+     */
+    public static Command parse(String command) throws DukeException {
+        String[] commandArr = command.split(" ");
+        String directive = commandArr[0];
+        if (directive.equals("list")) {
```
chaitanyabaranwal, line 28: Changing a chain of `if-else` statements to `switch-case` would make the code cleaner.

src/main/java/duke/main/TaskList.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public Task deleteTaskAtNumber(int position) {
+        return tasksList.remove(position - 1);
+    }
+
+    public Task getTaskAtIndex(int position) throws DukeException {
+        try {
+            return tasksList.get(position);
+        } catch (IndexOutOfBoundsException ex) {
+            throw new DukeException(ex.getMessage());
```
chaitanyabaranwal, line 59: Might want to consider customising the error message instead of throwing the `IndexOutOfBoundsException` message because it might be less verbose. :)

src/main/java/duke/main/Ui.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Reads a user command if it exists and returns to user.
+     *
+     * @return the user command
+     */
+    public String readCommand() {
+        if (scanner.hasNextLine()) {
+            lastCommand = scanner.nextLine();
+        }
+        return lastCommand;
```
chaitanyabaranwal, line 33: Since `lastCommand` is only used in this function, might want to move the private variable and make it a local variable inside this function.

src/main/java/duke/main/Ui.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public String getNotifyTaskDeleted(Task task, int listSize) {
+        String notification = "Noted. I've removed this task:\n";
+        notification += "\t" + task + "\n";
+        notification += String.format("Now you have %d tasks in the list.\n", listSize);
+        return notification;
+    }
+
+    public void notifyMarkedAsDone(Task task) {
```
chaitanyabaranwal, line 72: The functions `notifyTaskDeleted`, `notifyTaskAdded` and `notifyMarkedAsDone` don't seem to be used anywhere, probably want to refactor and remove them? If needed, probably it is a better idea to reuse the corresponding `getNotifyTaskDeleted`, `getNotifyTaskAdded` and `getNotifyMarkedAsDone` in these functions.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,113 @@
...
+            String[] dayMonthYear = by.split(" ")[0].trim().split("/");
+            String day = formatDate(dayMonthYear[0]);
+            String month = formatMonth(dayMonthYear[1]);
+            String year = dayMonthYear[2];
+            String time = formatHour(hoursMin);
+            by = String.format("%s of %s %s, %s", day, month, year, time);
+        }
+    }
+
+    private String formatDate(String date) {
```
chaitanyabaranwal, line 41: Cleaner to use switch-case in my opinion

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,33 @@
+package duke.task;
+
+/**
+ * Represents an Task object. A <code>Task</code> object
+ * specifies whether it has been completed and its item description.
+ */
+public class Task {
```
chaitanyabaranwal, line 7: It might be a good idea to make Task an abstract class, since the concrete instances only consist of Event, Deadline and Todo.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,36 @@
...
+import duke.main.Storage;
+import duke.main.TaskList;
+import duke.main.Ui;
+
+/**
+ * Represents an executable action. A <code>Command</code> object corresponds to
+ * an action to execute as specified by Duke's user.
+ * Determines if program should exit.
+ */
+public abstract class Command {
```
chaitanyabaranwal, line 13: I like how the `Command` is created as an abstract class by itself, and its concrete instances are the different command possible. Effective OOP usage.

chaitanyabaranwal (commented):
- Documentation very extensive, easy to understand what function which part of the code serves simply by looking at the documentation,
- Extensive OOP usage, makes it easier to debug and extend the program.
- Code is clean, easy to follow, and abides by conventions.

e0031374:
Overall everything looks good and up to date,

But perhaps a suggestion would be to use some sort of a Date class to not have to go through the hassle of writing out methods to let Duke understand date in Deadlines class. However as it is right now it works so its fine too.

The implementation of the optional task of having Command classes which is good and makes Duke open to new features if required.
Effective use of abstract classes in Task to reduce duplicate code.

Code is well organized into packages and well documented, following coding style guides.
Overall it is up to date on all tasks, and a good job

PR #166, opened 2019-08-27
[Lee Yiyuan] Duke Increments

LeeYiyuan:


src/main/java/task/TodoTask.java
```diff
@@ -0,0 +1,22 @@
...
+    public TodoTask() {
+        super();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        return String.format("[T][%s] %s", this.isDone ? "✓" : "✗", this.title);
```
waynefong0401, line 20: "this."  in `this.isDone` and `this.title` should be removed according to the coding standard. Same for DeleteTaskCommand.java

src/main/java/storage/format/TaskFormatter.java
```diff
@@ -0,0 +1,28 @@
+package com.leeyiyuan.storage.format;
+
+
+import com.leeyiyuan.task.Task;
+
+/**
+ * Represents a formatter to format a task.
+ */
+public abstract class TaskFormatter {
```
waynefong0401, line 9: After reading your code, I want to ask whether this class is a TaskFormatter or a CommandFormatter. Task is like a todo, event or deadline, while Command is including add, list, find etc.

thaowei, line 9: This could be an interface.

But it's really subjective I guess...

src/main/java/command/format/AddDeadlineTaskCommandFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+ * Represents a formatter for an AddDeadlineTaskCommand. 
+ */
+public class AddDeadlineTaskCommandFormatter extends CommandFormatter {
+
+    /** 
+     * {@inheritDoc} 
+     */
+    @Override
+    public Command parse(String text) throws CommandParseException {
+        if (Pattern.matches("^deadline ?/by.*$", text)) {
```
thaowei, line 21: Now you have two problems. :(

You didn't need to use regex here, I think. Just attempt to parse and fail if something is missing.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,128 @@
...
+    protected ArrayList<TaskFormatter> taskFormatters;
+
+    /**
+     * Constructs a Storage using a file path.
+     *
+     * @param filePath Path of file used to back Storage.
+     */
+    public Storage(String filePath) {
+        this.filePath = filePath;
+        this.taskFormatters = new ArrayList<TaskFormatter>();
```
cambrian-dk, line 36: (written for tutorial activity)

there may be a better abstraction for the TaskFormatters (instead of having to add a new TaskFormatter every time there is a new type of Task)

cambrian-dk (commented):
(tutorial activity)

src/main/java/command/format/AddDeadlineTaskCommandFormatter.java
LeeYiyuan: "Just attempt to parse and fail if something is missing" - I feel that that is the point of regex, which does in a more concise manner rather the process of having a whole chunk of code to do the checking.

LeeYiyuan: Fair enough. I felt that asserting a prior with a regex makes the parsing easier since you know that the text is valid and you can skip the messy checks in between the parsing process.

src/main/java/storage/format/TaskFormatter.java
LeeYiyuan: I had used an abstract class because I had expected to expand on the shared base functionalities of `TaskFormatters` in the future, in which case an abstract class makes more sense than an interface. At the moment, you are right, it is really subjective.

src/main/java/storage/Storage.java
LeeYiyuan: Hmm... could you come up with an example of how to better abstract this? Domain knowledge has to be specified somewhere, and I thought that with this abstraction it becomes convenient to specify task specific domain knowledge in an isolated, loosely coupled way through extending `TaskFormatter`. Let me know what you think!

src/main/java/storage/format/TaskFormatter.java
LeeYiyuan: As their name suggests, `TaskFormatter` is a TaskFormatter, while the `CommandFormatter` resides in `duke/src/main/java/command/format/CommandFormatter.java`/

src/main/java/task/TodoTask.java
LeeYiyuan: Can you point me to the exact location in the coding standard? I can't seem find this rule in both the [CS2103T guidelines](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html) and the [Google guidelines](https://google.github.io/styleguide/javaguide.html) which the CS2103T guidelines default to.

src/main/java/command/AddDeadlineTaskCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * {@inheritDoc} 
+     */
+    @Override
+    public void execute(TaskList tasks, UserOutputInterface uoi, Storage storage)
+            throws CommandExecuteException, StorageException {
+        if (tasks.containsTitle(this.title)) {
+            throw new CommandExecuteException("A task with the same title already exists.");
+        }
+
+        DeadlineTask task = new DeadlineTask();
```
gabrielseow, line 47: Have no issues, but would like to ask why you wanted to have setTitle and setBy methods instead of passing the title and 'by time' in the constructor

gabrielseow (commented):
Well written code! You manage a lot of cases, e.g. trying to mark a 'done' task as done or conflicts in duplicate event names, which is impressive

src/main/java/command/TaskDoneCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /** 
+     * {@inheritDoc} 
+     */
+    @Override
+    public void execute(TaskList tasks, UserOutputInterface uoi, Storage storage)
+            throws CommandExecuteException, StorageException {
+        if (tasks.size() >= this.index) {
+            if (tasks.get(this.index - 1).getIsDone()) {
```
gabrielseow, line 34: Mentioned it in a general comment, but as mentioned, good job covering cases where task is already done

src/main/java/command/AddDeadlineTaskCommand.java
LeeYiyuan: Hi Gabriel, thanks for the kind comments. I had considered the possibility of future requirements that involve mutating the task objects, and have decided on the said design of making tasks mutable through the setter methods. Hope this clarifies!

PR #168, opened 2019-08-27
[Chai Jie Tung] Duke Increments

jietung:


liakify:
Your code looks good! I especially like that you were fairly rigorous in your unit testing. However, I do think that you could have stored the tasks in `duke.txt` in a much cleaner format as it makes parsing it much easier!

liakify (commented):
Your code looks good! I especially like that you were fairly rigorous in your unit testing. However, I do think that you could have stored the tasks in `duke.txt` in a much cleaner format as it makes parsing it much easier!

jonathantjendana:
I like that you have many unit tests to minimise any regressions from occurring. However, I think you could extract your DateTime parser in order to make it more reusable and makes your code looks much cleaner.

PR #169, opened 2019-08-27
[Ho Wei Haw] Duke Increments

wh-coder-2025:


src/main/java/duke/exception/InvalidTimeAndDateException.java
```diff
@@ -3,8 +3,17 @@
 public class InvalidTimeAndDateException extends Exception {
     private static final String MESSAGE = "☹ OOPS!!! You are breaking the space-time continuum!";
```
fatclarence: I like the consideration of making this a "private static final" constant. Very good, something I shall adapt as well! Well done!

src/main/java/duke/tasks/FormattedDate.java
```diff
@@ -51,4 +51,9 @@ private boolean isValidTime(String time) {
     public String toString() {
         return FORMAT.format(this.date);
     }
+
+    @Override
+    public int compareTo(FormattedDate anotherDate) {
+        return this.date.compareTo(anotherDate.date);
```
fatclarence: This is actually a really interesting way to do a check on whether the date is of the valid format that we want, that is "dd/mm/yyyy hhmm", however, may I ask what is the purpose of having a comparison instead of catching ParseExceptions and reporting them to the user? Still I really like this idea! But I would like to know your thought process behind this.

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,55 @@
...
+    @FXML
+    private void handleUserInput() {
+        String input = userInput.getText();
+        String response = duke.getResponse(input);
+        Group userDialog = new Group(DialogBox.getUserDialog(input, userImage));
+        Group dukeDialog = new Group(DialogBox.getDukeDialog(response, dukeImage));
+        dialogContainer.getChildren().addAll(userDialog, dukeDialog);
+        userInput.clear();
+        if (input.equals("bye")) {
+            Platform.exit();
```
bakwxh: It seems that the GUI will exit on its own when the user types "bye". Very good idea as it totally disables the user from making any additional input in the same session. I may consider doing this as well!

src/main/java/duke/tasks/FormattedDate.java
wh-coder-2025: I initially did that because it might be possible for the user to enter both the start and end dates in a valid format even though it doesn't make sense for the end date to be earlier than the start date.

However, I removed this function in my latest commit as I decided to make use of Java's LocalDateTime package which is able to throw an DateTimeParseException should the user enters an invalid date. As a result, I am still trying to figure out how to include this exception when the user enters an end date that is earlier than the start date for an event task.

PR #171, opened 2019-08-27
[Maurice Tan] Duke Increments

MauriceTXS:


data/duke.txt
```diff
@@ -0,0 +1,5 @@
+T | 0 | test
+T | 0 | test
+T | 0 | test
+T | 0 | test
+T | 0 | test
```
Yehezkiel01: I would recommend ignoring data/duke.txt file in .gitignore since this file most likely will change everytime you run the duke app. Therefore, it would be more convenient to just ignore this file.

src/main/java/UI.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Greetings when the program starts
+     */
+    public String greet() {
+        /*horizontalLine();
+        System.out.println("Hello I'm Duke");
+        System.out.println("What can I do for you?");
+        horizontalLine();
+        System.out.println();*/
```
Yehezkiel01: It might be better to remove this comment part since you can always refer back to git history if you need to compare your code.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,26 @@
...
+        super(description, isDone);
+    }
+
+    @Override
+    public String getType() {
+        return "T";
+    }
+
+    @Override
+    public String getDate() {
```
Yehezkiel01, line 18: I'm not pretty sure why a dummy getDate() method is needed in here? It might be a better idea to remove getDate from Task class and implements it only in relevant Task such as Deadline and Event.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Deletes the task from the list
+     * @param index Index of the task
+     */
+    public String deleteTask(int index) {
+        index = index - 1;
+        Task taskHolder = taskList.remove(index);
+        String holder = ui.horizontalLine() + "\nNoted. I've removed this task:\n" + "  [" + taskHolder.getType() + "]" + "[" + taskHolder.getStatusIcon() + "] " + taskHolder.getDescription() + taskHolder.getDate() + "\n" + ui.horizontalLine() + "\n";
```
Yehezkiel01: If the String becomes too complicated, you can try looking at StringBuilder. Aside from making the code looks simpler, StringBuilder is usually faster than String concatenation operation. From what I know, String is an Immutable object and therefore every concatenation operation will result in a new String Object.

Yehezkiel01 (commented):
Overall, the whole code looks good. I have added some of my comment at part which I think can be improved. Most of them are only minor issues.

One more thing, I would like to point out is to add a new line at the end of each file. Although, it is not in the coding standard, but usually git complains when there is no new line.

src/main/java/TaskList.java
MauriceTXS: Noted, thank you for the tip provided!

src/main/java/ToDo.java
MauriceTXS: Ohh, I made my abstract Task class implement an abstract getDate() method, hence, I need a dummy getDate() method for todo. Thanks for pointing it would will remove it!

src/main/java/UI.java
MauriceTXS: yup, thanks!

data/duke.txt
MauriceTXS: noted, thank you!

PR #173, opened 2019-08-27
[Zhang Dongjun] Duke Increments

Auxinnn:


src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,227 @@
+package duke.command;
```
TeoShyanJie, line 1: maybe can try to split command into different class for more OOP

TeoShyanJie (commented):
The overall is really good but the parser class maybe can try to make it more OOP

src/main/java/duke/command/Parser.java
Auxinnn: Sure! Thanks for the suggestions!

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,227 @@
...
+            System.out.println(i + 1 + "." + TaskList.getList().get(i));
+        }
+    }
+
+    /**
+     * Action taken when the command is done.
+     */
+    private void commandDone() {
+        Storage storage = new Storage();
+        int num = Integer.parseInt(action[1]);
```
zhangxuan97: This may be just me nitpicking haha but perhaps you can try to catch IndexOutOfBoundsException within this method.
After splitting the input (command) at " " in your Ui class and passing it into your Parser class as the array "action", it may not always be the case the array has a length more than 1. For instance, if I accidentally entered "done" instead of the correct "done 1" syntax, the array action would be of length 1 and calling action[1] may result in an IndexOutOfBoundsException which may unintentionally terminate the program. Hope that helps :)

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,227 @@
...
+        Task newTask = TaskList.getList().get(num - 1);
```
zhangxuan97: Maybe you can also try catching index out of bounds exception here: 
Suppose we currently have 3 items in the task list. If the user made a fat finger error and typed "done 4" this would throw the out of bounds error and unintentionally terminate the program. So maybe it would be helpful to catch the error and then print something like "Ensure task number is between 1 and " + TaskList.getList().size() to remind the user what the range of valid task numbers are :)

zhangxuan97 (commented):
Overall your code is really well organised and documented! My only suggestion is to try to catch more errors to prevent the program from accidentally terminating :)

src/main/java/duke/command/Parser.java
Auxinnn: Thank you for the suggestions! I will add them on.

PR #174, opened 2019-08-27
[Hu Yuchen] Duke Increments

brebeek:


nordic96 (commented):
5e3d371

I like how you used the switch case statement and enum to handle the commands. :)
Coding standards are well met (Parser.java:94 isFound) and neat.
One point I would like to suggest is perhaps you can consider changing the enumerations in Action (Parser.java:120) into upper case instead of lower case? to meet the convention standards. (list, bye, done, todo, ... -> LIST, BYE, DONE, TODO, ...)

nordic96 (commented):
9bd8e01

Good choice in assigning access modifiers to all your methods in the class files :)
I think it is a good idea to keep all methods either public, private or protected as software engineering highly encourage developers to follow the conventions/ guidelines.
Keep up the good work :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+import duke.actionstack.DukeActionStack;
+import duke.io.Storage;
+import duke.ui.DialogPanel;
+import duke.ui.Ui;
+import duke.tasklist.TaskList;
+
+import java.io.*;
```
halfwhole: You may want to import only the classes required! This change makes it clearer what components of `io` you're using.

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+    private TextField userInput;
+    private Button sendButton;
+    private Scene scene;
+    private Image user = new Image(this.getClass().getResourceAsStream("/images/DaUser.png"));
+    private Image bot = new Image(this.getClass().getResourceAsStream("/images/DaDuke.png"));
+
+    /**
+     *
+     *
+     */
```
halfwhole, line 42: What is with the empty Javadoc comment?

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+        boolean isExit = false;
+        Label userText = new Label(userInput.getText());
+            try {
+                if(userInput.getText().contains("bye")) {
+                    isExit = true;
+                }
+                response = ui.readUserInput(userInput.getText());
+            } catch (Exception e) {
+                response = e.getMessage();
+            }
```
halfwhole, line 136: Indentation of the `try` block, and try not to catch all exceptions (but only specific ones)

src/main/java/duke/actionstack/DukeActionStack.java
```diff
@@ -0,0 +1,22 @@
+package duke.actionstack;
+
+import java.util.LinkedList;
+
+public class DukeActionStack {
```
halfwhole, line 5: A Javadoc comment would be helpful!

src/main/java/duke/command/AddDeadline.java
```diff
@@ -0,0 +1,34 @@
...
+import duke.io.Storage;
+import duke.task.Deadline;
+import duke.task.Task;
+import duke.tasklist.TaskList;
+
+import java.io.FileNotFoundException;
+import java.text.ParseException;
+
+public class AddDeadline {
+    public static String addDeadline(String act, Storage storage) throws DukeIllegalDescriptionException,
```
halfwhole: The classes `AddDeadline`, `AddEvent`, `AddTodo` etc. should inherit from a general `Command` class, this would help to make your code more OOP.

src/main/java/duke/io/Parser.java
```diff
@@ -0,0 +1,74 @@
...
+            DukeIllegalDescriptionException, DukeIllegalActionException, DukeDuplicateTaskException {
+        String response = "";
+        if(DuplicateChecker.checkDuplication(act)) {
+            throw new DukeDuplicateTaskException();
+        }
+        try {
+            switch (Action.valueOf(act.split(" ")[0].toUpperCase())) {
+            case LIST:
+                response = List.list();
+                break;
```
halfwhole: The parsing for your commands should be shifted here, should only give the commands what they need (instead of delegating the parsing to the commands themselves)

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,38 @@
...
+
+import duke.task.Task;
+
+import java.util.LinkedList;
+
+/**
+ * Holds a universal list of tasks.
+ */
+public class TaskList extends LinkedList<Task> {
+    public static LinkedList<Task> taskList;
```
halfwhole, line 11: You probably shouldn't have a public static variable that's accessible to all other classes because it's not a good practice, would be good to be private and non-static instead, for the instance to be passed around and accessed through methods

src/main/java/duke/command/List.java
```diff
@@ -0,0 +1,13 @@
+package duke.command;
+
+import duke.tasklist.TaskList;
+
+public class List {
+    public static String list() {
```
halfwhole: Would be good to have clearer naming also, e.g. `ListCommand`

src/main/java/duke/io/Parser.java
```diff
@@ -0,0 +1,74 @@
+package duke.io;
+
+import duke.actionstack.DukeActionStack;
+import duke.command.*;
```
halfwhole: Should import the classes individually

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,48 @@
...
+public class Ui {
+
+    /**
+     * Generates and prints the greeting message of DukeBot.
+     */
+    public String initMessage() {
+          return ("Hello! I'm Duke\n" + "What can I do for you?");
+    }
+
+    private Storage storage;
```
halfwhole, line 24: Should be moved up

PR #177, opened 2019-08-27
[Wilfred Bradley Tan] Duke Increments

wilfredbtan:


src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.BufferedReader;
+import java.text.ParseException;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Locale;
+
+/**
+ * Storage class which handles the loading and saving of Task data.
+ */
+public class Storage implements StorageInterface {
```
dalsontws: Good use of the Interface class!

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public static Command parse(String userInput) throws DukeException {
+        Scanner sc = new Scanner(userInput);
+        Command newCommand = new Command();
+        String commandString = sc.next();
+        newCommand.addCommandString(commandString);
+
+        switch (commandString) {
+        case "find":
+            newCommand.addKeyword(sc.next());
```
dalsontws: Will there be a possible "NoSuchElementException" error? If command follows with a blank.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public String getStatusIcon() {
+        return (isDone ? "✓" : "✘");
+    }
+
+    /**
+     * Sets the Task as done or not depending on the argument.
+     * @param isDone boolean to determine whether Task is marked as done or not.
+     */
+    public void setDone(boolean isDone) {
```
dalsontws: Maybe the method setDone can be used without a variable? i.e. setDone()
The name of method is self-explanatory as setting the task as Done.
If the variable boolean isDone is assigned incorrectly might have errors!

dalsontws (commented):
Code written is really clean and easy to read!

src/main/java/command/Command.java
```diff
@@ -0,0 +1,219 @@
...
+ * Command class that executes parsed commands given to it.
+ */
+public class Command {
+
+    /** Command portion of the input. */
+    private String commandString;
+    /** Description portion of the input. */
+    private String desc;
+    /** Date when the task starts. */
+    private LocalDate startDate;
```
shaoyi1997: Would it be better if you have subclasses of the Command class that contain the date & time fields? Since commands like "list", "done" and "bye" do not have date and time input.

src/main/java/parser/Parser.java
wilfredbtan: Yo thanks for raising this up! Did this awhile back and forgot about it haha. Fixed it already thanks! 😀

src/main/java/command/Command.java
wilfredbtan: Thanks for the suggestion man. Updated it in my latest commit.

PR #178, opened 2019-08-27
[Russell Loh] Duke Increments

Russell-Loh-NUS:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public String getResponse(String input) {
+        Command command = Parser.parse(input); // Parse and perform logic
+        return command.execute(ui, storage, taskList);
+    }
+
+    public static void main(String[] args) {
+        Application.launch(Main.class, args);
+    }
+}
```
ROHITREDDYBALAM, line 54: Well abstracted and easy to read code! The comments made it easy to follow the code.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class Parser {
+
+    /**
+     * Returns a command object that represents the command entered.
+     *
+     * @param input Command to be parsed.
+     * @return Command object.
+     */
+    public static Command parse(String input) {
```
weiijiie, line 23: This method is rather complex with a number of nested switch and if/else statements. It may be tedious/difficult to refactor given new commands/change in requirements for existing commands/if you accidentally got one of the indices wrong. Could perhaps refactor some of the logic out into their own methods, or maybe even have each Command class have its own `isValid()` method to separate the logic of validity checking from parsing the correct command.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,72 @@
...
+            break;
+        default:
+            ;
+            break;
+        }
+
+        if (task != null) {
+            storage.save(taskList.getTaskList());
+
+            return ui.returnOutput("  " + task,
```
weiijiie, line 66: In all your commands, only static methods of the `Ui` class are accessed through this ui instance. You should instead access those static methods by using the class name instead.

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Executes the current command.
+     *
+     * @param ui       Ui object.
+     * @param storage  Storage object.
+     * @param taskList TaskList object.
+     */
+    public String execute(Ui ui, Storage storage, TaskList taskList) {
+        new Timer().schedule(new TimerTask() {
```
weiijiie, line 20: Nice use of timer to delay the app's shutdown!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+        try {
+            String path = this.filePath;
+            FileWriter fw = new FileWriter(path);
+
+            String text = "";
+            for (int i = 0; i < tasks.size(); i++) {
+                //Get tasks
+                Task task = tasks.get(i);
+
+                if (task instanceof Todo) {
```
weiijiie, line 44: It may be worth refactoring the logic for each task into an abstract method of Task, implement a concrete implementation for each Class that extends Task, and use polymorphism to get the serialized text, instead of using if/else statements, so that if there is a requirement for different types of tasks in the future, you don't need to touch the code here.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static MainWindow mw;
+
+    /**
+     * Creates a Duke Exception object.
+     *
+     * @param message Error message to print.
+     */
+    public DukeException(String message) {
+        mw.setDukeMessage("\u2639 OOPS!!! " + message);
```
weiijiie, line 16: I think having the `DukeException` class print messages to a static member in the constructor may potentially be confusing for some developers who are more used to the convention of using try/catch blocks to handle exceptions.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+                    text += "T|" + task.getIsDone() + "|" + task.getDescription();
+                } else if (task instanceof Deadline) {
+                    text += "D|" + task.getIsDone() + "|" + task.getDescription() + "|"
+                        + ((Deadline) task).getDateString();
+                } else if (task instanceof Event) {
+                    text +=
+                        "E|" + task.getIsDone() + "|" + task.getDescription() + "|" + ((Event) task)
+                            .getDateString();
+                }
+
```
ROHITREDDYBALAM, line 54: code can be abstracted further at this point

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,42 @@
...
+     * @param s           String to be modified.
+     * @param taskMessage Message returned from task.
+     * @param listSize    Size of the task list.
+     * @return Modified string.
+     */
+    public static String returnOutput(String s, String taskMessage, int listSize) {
+        return taskMessage + "\n" + s.replace("\n", "\n    ") + "\n" + "Now you have " + listSize
+            + " tasks in the list.";
+    }
+}
```
ROHITREDDYBALAM, line 44: Can the exit method be placed here?

PR #180, opened 2019-08-27
[Daryl Koh] Duke Increments

cambrian-dk:


src/main/java/Duke.java
```diff
@@ -1,23 +1,48 @@
...
+        printIndentedMsg("Hello! I'm Duke");
+        printIndentedMsg("What can I do for you?");
+        printLine();
         String msg = s.nextLine();
-        while(!msg.equals("bye")){
-            System.out.println(line);
-            System.out.println(indent + msg);
-            System.out.println(line);
+        ArrayList<String> toDoList = new ArrayList<>();
+        while(!msg.equals("bye")) {
```
AHaliq: good approach to terminate the loop. Might be better to use a boolean variable so that you can terminate in other circumstances than the command 'bye' in the future.

src/main/java/Deadline.java
```diff
@@ -1,31 +1,39 @@
 public class Deadline extends Task{
     protected String deadline;
-    
-    public Deadline(String description){
-        super(description.split("/by", 2)[0]);
-        this.deadline = description.split("/by", 2)[1];
+
+    public Deadline(String sentence) throws DukeException{
+        this(sentence.split("/by", 2)[0], (sentence.split("/by", 2).length > 1 ? sentence.split("/by", 2)[1] : "" ));
         this.taskType = "D";
```
AHaliq: could be an enum or written as a literal in toString for each task type

src/main/java/Duke.java
```diff
@@ -25,7 +25,7 @@ public static void main(String[] args) {
         ArrayList<Task> toDoList = new ArrayList<>();
         while(!(msg.equals("bye") && commands.length == 1)) {
             try{
-                switch (commands[0]) {
+                switch (commands[0].toLowerCase()) {
```
AHaliq: good solution to ignore case!

AHaliq (commented):
Good simple approach with switch statement for each command!

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,52 @@
...
+        this.searchTerm = searchTerm;
+    }
+
+    /**
+     * Executes the command to search for a keyword.
+     * It gets the {@literal arrayList<Task>} from within the TaskList to be able to search,
+     * then it loops through each task and searches if the description contains the keyword to be searched.
+     * It then creates a new TaskList from all the matches, and then creates a new ListCommand to list out
+     * all the tasks that contain the keyword.
+     *
```
tohcejasmine, line 37: A shorter JavaDoc comment is sufficient, I feel your code is readable enough to understand the mechanism behind the searching of matching keywords :-)

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,45 @@
...
+     * the TaskList class.
+     *
+     * @param tasks   The TaskList to be searched
+     * @param ui      The ui to send messages to
+     * @param storage Storage if required for the execution.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        tasks.addTaskToList(this.taskToAdd);
+        ui.messageUser("Got it. I've added this task:",
```
tohcejasmine, line 46: Maybe the job of what messages to show the user can be handled by Ui directly instead, where Ui has corresponding methods for the different executions which can be called here. This might be helpful if the format of messages in the program change, where it can edited in one file instead of across different files.

tohcejasmine (commented):
Good variable naming that makes the code easy to understand and nice appropriate abstraction of classes 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
+import duke.command.Command;
```
FelixNWJ, line 1: Different import packages should be separated by a line. Checkstyle requirement

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
...
 public class Duke {
+
+    /**
+     * The Storage used for access/writing to files.
+     */
+    private Storage storage;
+    /**
+     * The filepath to read from/write to.
+     */
+    private String filePath;
```
FelixNWJ: filePath does not change, maybe consider making it final? Applies for Ui as well

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Abstract method to execute the respective command.
+     *
+     * @param tasks   The user's current TaskList
+     * @param ui      The ui currently being used by the user
+     * @param storage The storage object being used by the user
+     * @throws DukeException An error trying to carry out the command
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
```
FelixNWJ: It would be nice if Command objects did not access Storage objects directly, but do so through TaskList objects instead. Reason being your implementation does not make use of Storage in the execute method.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,43 @@
...
+    public Command() {
+        isExit = false;
+    }
+
+    /**
+     * Check if a command is an exit command.
+     *
+     * @return whether the command is an exit command
+     */
+    public boolean isExit() {
```
FelixNWJ, line 40: Maybe consider returning this boolean at the execute method instead (execute return type changes from void to boolean), because the boolean only changes after execute().

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        ArrayList<Task> taskList = tasks.getTaskArrayList();
+        ArrayList<Task> matches = new ArrayList<>();
+        for (Task task : taskList) {
+            if (task.getTaskDescription().contains(this.searchTerm)) {
+                matches.add(task);
+            }
+        }
+        new ListCommand().execute(new TaskList(matches), ui, storage);
```
FelixNWJ: Would it be better to have a print method in TaskList itself, instead of relying on creating a new ListCommand to do so? Considering that you have to depend on ListCommand, testing may be a problem.

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Constructor for ListCommand.
+     */
+    public ListCommand() {
+        super();
+    }
+
+    /**
+     * Execute the ListCommand.
+     * It retrieves the {@literal ArrayList<Task>} from inside the TaskList,
```
FelixNWJ, line 25: Are you trying to link ArrayList here? @link may be better?

FelixNWJ (commented):
I commented because of this weeks (Week 5) tutorial requirements. Overall, great job!

PR #181, opened 2019-08-27
[Kendrick Ang] Duke Increments

KendrickAng:


Xelyion (commented):
Good use of formatting to make GUI pleasant looking. However, "list" command only displays 4 tasks after the initial display.

KendrickAng:
> 
> 
> Good use of formatting to make GUI pleasant looking. However, "list" command only displays 4 tasks after the initial display.

Hi Xelyion, thank you for your input! I'm aware of the issue where the message is truncated if there are too many tasks to display. I'm working on a solution for this, and will try to release it as soon as possible.

john0227 (commented):
Hi Kendrick! I didn't think of adding the line divider to the GUI. I thought that was cool. When you exit the program, maybe you can use Platform.exit()? That's what I used, and it seems to work fine!

PR #182, opened 2019-08-27
[Tho Wen Long, Paul] Duke Increments

dragontho:


atharvjoshi (commented):
Hello! Good job on making it to week 4 of the iP!!!

- Your documentation and commenting rigour is impressive - hope to learn how to be as concise and thorough as you. I tend to ramble sometimes when I'm writing a javadoc description for classes or methods. 

- Well encapsulated classes - you've followed all OOP principles that I'm familiar with. When it came to handling event, deadline, and todo - i used a single 'AddCommand' instead of creating separate command classes for each. This is because I realised that the three commands were kinda doing the same thing, just generating different task types. So, I separated the task generation in my parser, while keeping the task execution consistent under one 'AddCommand' for these three tasks. But hey, your approach works too, and achieves a nice one-to-one mapping between the commands understood by Duke and the command classes written in your project.

- I like how you handled a lot of the 'illegal moves' on the task list (e.g. deleting task index that does not exist in current list) within the TaskList class itself. I will probably take inspiration from this and change my code during next week's increments. Currently, I am handling these 'illegal cases' within the execute() method of the commands. 

- Great use of final string constants in UI class. Makes making changes to Duke's responses WAY easier. I am currently hard-coding the responses within each method in UI - which is laborious!

All the best for the upcoming weeks!

PR #183, opened 2019-08-27
[She Jiayu] Duke Increments

jiayushe:


src/main/java/duke/Gui.java
```diff
@@ -0,0 +1,31 @@
...
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+
+import java.io.IOException;
+
+/**
+ * A GUI for Duke using FXML.
```
Oscar-B-Liang, line 13: I like this part very much, it is very appreciable to write the start method outside the original Duke file, which minimizes the influence of building the front end to the original code.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,63 @@
...
+        return this.description;
+    }
+
+    /**
+     * Gives a string representation of the task for storage.
+     *
+     * @return a string representation of the task for storage.
+     */
+    public String toDataString() {
+        return (this.isDone ? 1 : 0) + " | " + this.description;
```
Oscar-B-Liang: I have a minor comment here, I am sure that this code here works perfectly, but this explanation might not be as clear and straightforward than if else form. I feel this way, do you agree?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,129 @@
+package duke.task;
+
+import duke.exception.DukeException;
+import duke.util.Storage;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+
+/**
+ * TaskList class, for in memory storage of tasks.
```
Oscar-B-Liang, line 10: The program structure is already very clear and well organized. but somehow, I feel that put TaskList class outside the task package can make the program structure even clearer. Do you think so?

Oscar-B-Liang (commented):
I like your code very much and you used JUnit and gradle very well to automate your program testing. I believe you must be a well trained and advanced programmer. The code in general have very high quality with just a few minor comments.

PR #185, opened 2019-08-27
[Marcus Teh] Duke Increments

marcusteh1238:
If you can find bugs please let me know! :)


czkay:
I really liked how clean your code is, especially in your Ui class. The way you ensured proper date/time formatting in your TimedTask class was also really meticulous.

I think you could have named the methods of failcases in your unit tests according to conventions (i.e. whatIsBeingTested_descriptionOfTestInputs_expectedOutcome).
You can also cut out a lot of code if you use an ObjectInput/OutputStream to directly write/load your TaskList object to your hard drive, instead of formatting your tasks with toFileString() etc.
You also create several new Ui objects across some of your classes, maybe it's better to pass one Ui object around or maybe use static methods? Might just be my preference though.
And lastly, I think there's supposed to be an empty line between your method description and your @param line in your Javadoc headers.

Overall your code was nice to read, thanks!

src/main/java/Task.java
```diff
@@ -61,21 +61,21 @@ protected static Task create(String command) throws DukeException {
...
-            return newTask;
+        String[] args = item.split(" \\| ");
+        String taskType = args[0];
+        boolean isAlreadyDone = Integer.parseInt(args[1]) == 1;
+        Task newTask;
+        if (taskType.equals("D")) {
+            newTask = new DeadlineTask(args[2], TimedTask.parseDateTime(args[3]));
+        } else if (taskType.equals("E")) {
+            newTask = new EventTask(args[2], TimedTask.parseDateTime(args[3]));
+        } else {
```
mrchensheng21, line 114: Good use of parseDateTime. Helps to clean up the code and also to fulfil its purpose of changing the date time to words.

src/main/java/Duke.java
```diff
@@ -1,104 +1,41 @@
...
-     */
-    public static boolean isValidTaskType(String taskType) {
-        for (String s : validTaskTypes) {
-            if (taskType.equals(s)) {
-                return true;
-            }
-        }
-        return false;
+    public static void main(String[] args) {
+        new Duke("D:/Marcus Folder/SCHOOL STUFF/YEAR 2/CS2103T/duke/data/duke.txt").run();
```
mrchensheng21: Duke class is very concise

PR #186, opened 2019-08-27
[Olivia Yu] Duke Increments

Alaete:


nordic96 (commented):
f49e6ad

Good work that you created extended classes for your Command class instead of handling all command cases in multiple if else statements :)

brebeek (commented):
I like your organisation of JavaFX components. Keep up the good work!

PR #187, opened 2019-08-27
[Chen Xi] Duke Increments

chen-xi-cx:
Add week 2 and week 3 iP increments


src/main/java/command/Command.java
```diff
@@ -0,0 +1,26 @@
...
+public abstract class Command {
+    /**
+     * @return false to signify the program is not terminating
+     */
+    public boolean canEnd() {
+        return false;
+    }
+
+    /**
+     * Execute a command.
```
liawsy: Minor issue, should be "Executes a command", refer to Javadocs section in [Java coding standards](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html).

src/main/java/converter/StringDateConverter.java
```diff
@@ -0,0 +1,35 @@
+package converter;
+
+import java.text.SimpleDateFormat;
```
liawsy, line 4: Cool use of Java built-in classes!

src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @return error message
+     */
+    @Override
+    public String getMessage() {
+        switch(this.type) {
+        case "invalid number":
+            return "Please enter a valid number.";
+        case "empty todo":
```
liawsy: error message type can be more concise.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Parses user input into command for execution
+     *
+     * @param userInput String of input of user
+     * @return the command based on user input
+     * @throws DukeException if user input is invalid
+     */
+    public Command parse(String userInput) throws DukeException {
+        String[] separatedInputs = userInput.trim().split("\\s+");
+        switch (separatedInputs[0].toLowerCase()) {
```
liawsy, line 30: nice consideration of how input could possibly be capitalised.

liawsy (commented):
Clean and readable code. Also, good organisation of your code using the packages. Here're some other feedback:
1) Javadocs are written in a detailed and informative manner, although be sure to note that by standards, each statement should end with a full stop.
2) Instead of using the `IncorrectCommand` class, you could make it an exception.
3) Perhaps you could create a class that handles all the messages, so that when you want to make changes to the messages in the UI, you don't have to go through every single file that contains their respective messages to make changes. 

Overall, really good job and I enjoyed reading your code.

src/main/java/converter/StringDateConverter.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Date;
+
+/**
+ * Converts String object to Date object.
+ */
+public class StringDateConverter {
```
aidilfbk, line 10: I think it would be a good idea to turn the methods in this class to be static as they do not seem to be using any instance specific properties.

src/main/java/converter/StringDateConverter.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Convert String object to Date object.
+     *
+     * @param strDate in the format "dd/MM/yyy HHmm"
+     * @return date object
+     * @throws Exception if the date is unparesable
+     */
+    public Date convertStringToDate(String strDate) throws Exception {
```
aidilfbk: Good use of checked exceptions but I think it could provide more clarity by being more specific with the type of exception that may be thrown.

src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * Returns String error message.
+     *
+     * @return error message
+     */
+    @Override
+    public String getMessage() {
+        switch (this.type) {
```
aidilfbk, line 32: This would be a good place to use enums so other developers don't have to remember the specific strings that correspond to a specific error message.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public class Parser {
+    /**
+     * Parses user input into command for execution.
+     *
+     * @param userInput String of input of user
+     * @return the command based on user input
+     * @throws DukeException if user input is invalid
+     */
+    public Command parse(String userInput) throws DukeException {
```
aidilfbk: I think it would be a good idea to make the `parse` method static as it does not seem to be using any instance specific properties.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+        case "delete":
+            if (separatedInputs.length <= 1) {
+                throw new DukeException("invalid number");
+            }
+            Scanner s = new Scanner(separatedInputs[1]);
+            if (separatedInputs.length > 2 || !s.hasNextInt()) {
+                throw new DukeException("invalid number");
+            }
+            if (separatedInputs[0].toLowerCase().equals("done")) {
+                return new DoneCommand(s.nextInt() - 1);
```
aidilfbk: I may be misunderstanding the condition that will trigger this line of code. The outer switch checks for `separatedInputs[0].toLowerCase()` being equal to `"delete"` but there is also an if-else conditional check for `separatedInputs[0].toLowerCase().equals("done")`.

I believe these two cases are mutually exclusive but is there an example input test case that explains it better?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,115 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Loads the file where task list is stored.
+     *
+     * @return list of task if the file is found
+     * @throws FileNotFoundException when no task list file is found
+     */
+    public ArrayList<Task> load() throws FileNotFoundException {
```
aidilfbk, line 41: Clear logic of how the file contents is parsed. :+1:

src/main/java/command/DeadlineCommand.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param tasks contains task list
+     * @param ui deals with interaction with users
+     * @param storage deals with loading and saving of task list
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        try {
+            StringDateConverter converter = new StringDateConverter();
+            Date by = converter.convertStringToDate(arguments[1].trim());
+            tasks.getTasks().add(new Deadline(arguments[0], by));
```
aidilfbk: I think it would provide clarity if there was a temporary variable to reference the `new Deadline` instance since it is later used, rather than requesting it again from the `tasks` variable.

A similar issue appears in [`DoneCommand`](https://github.com/nus-cs2103-AY1920S1/duke/pull/187/files#diff-a1acc0716924f84df197496c2462f84fR36).

src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Represents all exception unique to Duke.
+ */
+public class DukeException extends Exception {
+    protected String type;
+
+    /**
+     * Initializes DukeException with error type.
+     * @param input the error type
```
aidilfbk, line 11: Minor Javadoc style violation as there should be an [empty line between description and parameter section](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#comments). This style is adhered to in other Javadoc comments so I believe this style violation was unintended.

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Initializes task list with loaded tasks.
+     *
+     * @param tasks is the task list
+     */
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    public ArrayList<Task> getTasks() {
```
aidilfbk, line 27: It seems like every other class that uses a `TaskList` needs to call this method to do their operations. Consider providing those methods (e.g. `add(Task)`) on the `TaskList` instead of allowing unrestricted manipulation of the `ArrayList<Task>`.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+        String[] separatedInputs = userInput.trim().split("\\s+");
+        switch (separatedInputs[0].toLowerCase()) {
+        case "bye":
+            return new ExitCommand();
+        case "list":
+            return new ListCommand();
+        case "done":
+            //fallthrough
+        case "delete":
+            if (separatedInputs.length <= 1) {
```
aidilfbk: I like that input validation is performed as early as possible so that `Command` implementations can be simplified. 👍

aidilfbk (commented):
Hi Chen Xi,

I've made my comments on specific lines where relevant and my thoughts in general are below:
- Very good use of packaging :+1:
- Code is well organized and logically implemented, making it easy to follow.
- I'm not sure if the Javadoc style for parameters should treat the parameter name as part of the description (e.g. "[@param tasks contains task list](https://github.com/nus-cs2103-AY1920S1/duke/pull/187/files#diff-ae04d255fd2aa7bba6aee1540703424eR23)") but I'll check with my TA on this matter.
- I think it would provide more clarity if method implementations were logically chunked (see ["logical units within a block should be separated by one blank line" from the Coding Standard](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#layout)).

I hope you found my feedback useful! 😄

PR #189, opened 2019-08-28
[Low Cheng Yi] Duke Increments

Icesiolz:


src/main/java/Duke.java
```diff
@@ -1,10 +1,232 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
-    }
-}
+        new Duke("C:/Users/Low Cheng Yi/Desktop/CS2103/duke/src/data/tasks.txt").run();
```
daekoon: I would recommend using a relative path for your file so that you can use the same save files across different devices.

daekoon (commented):
1. Just a small point - There are some parts of your code that has unnecessarily large number of line breaks (3~4 lines). Can consider reducing those?

2. For retrieving descriptions from command line, you can set a limit to how many strings to split a sentence into, something like `str.split(" ", 2)`. This way you can nicely split the first word (which is the command) and the rest of the line (which is the description). 

3. Similarly, you can split the command by /by and /at using `str.split('/at')` as well. A lot easier implementation.

4. I noticed for your Event/Deadline classes, you set your variables to protected level access. I would recommend keeping those private, because I don't think (please correct me if I'm wrong) I've seen these variables being accessed directly from external classes.

src/main/java/duke/command/AddTaskCommand.java
```diff
@@ -0,0 +1,80 @@
...
+                        }
+                        break;
+                    }
+                }
+                task = new Deadline(descriptions.toString().trim(), by.toString().trim());
+                tasks.addTask(task);
+                return ui.addTask(task, tasks.getNumber());
+
+            default:
+                return "This shouldn't appear.";
```
wallacelim, line 75: Perhaps you'd want to throw an exception here instead

wallacelim (approved):
Everything looks good! OOP principles and naming conventions are well-adhered to. Project architecture seems well-structured too

src/main/java/Duke.java
Icesiolz: Thanks! May I know how to do that? >.< I tried using the ../ and it didnt work.

src/main/java/duke/command/AddTaskCommand.java
Icesiolz: Okay thanks!

PR #190, opened 2019-08-28
[Si Ning] Duke Increments

gohsnn:


src/main/java/Main.java
```diff
@@ -10,7 +10,7 @@
  */
 public class Main extends Application {
 
-    private Duke duke = new Duke();
+    private Duke duke = new Duke("../duke/data/duke.txt");
```
le0tan: Using relative address might be confusing - after gradle built the executable is inside ./build folder, however after release the .jar file could be anywhere. Probably it's better if you can move the .txt file into resource folder so that it's always stored at a fixed location.

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+import duke.managers.Ui;
+import duke.commands.Command;
+
 public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Initialises a new Duke object including new objects such as storage, ui and task list.
```
Kalsyc: Need a new line separating description of method/class and the @param @return tags for JavaDocs

Kalsyc, line 15: Need a new line separating description of method/class and the @param @return tags for JavaDocs

src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Creates a deadlineTask and adds it into memory.
+     * If command does not include a deadline, then user is prompted to enter the command again.
+     * @param b a String of information containing details for the Deadline task
+     * @exception DukeException is thrown when there is an error with the input
+     * @exception IOException is thrown when there is an error saving the data in the hard disk
+     */
+    private String deadlineTask(String b) throws DukeException, IOException {
```
Kalsyc: I feel like this should belong to some Parser.class? Or at least be abstracted to check whether it's a valid input first.

src/main/java/duke/commands/DeleteCommand.java
```diff
@@ -0,0 +1,49 @@
...
+     * @exception IOException is thrown when there is an error saving the data in the hard disk
+     */
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException, IOException {
+        int maxNum = tasks.totalNumTasks();
+        if (taskToBeDeleted > maxNum) {
+            throw new DukeException("Oops! This task number does not exist.");
+        } else {
+            Task deleted = tasks.delTask(taskToBeDeleted);
+            maxNum = tasks.totalNumTasks();
+            maxNum = tasks.totalNumTasks();
```
Kalsyc: Repeated statement?

src/main/java/duke/managers/DateTime.java
```diff
@@ -0,0 +1,154 @@
...
+        return dayWithSuffix;
+    }
+
+    /**
+     * Changes the month number to its name in the date String. It is used in the getDate method.
+     * @param month a String containing the month number
+     * @exception DateException if the month number is invalid
+     */
+    private static String wordMonth(String month) throws DateException {
+        int monthNum = Integer.parseInt(month);
```
Kalsyc, line 57: Hmm, should we be using the index of monthName you have defined earlier?

Kalsyc (commented):
Overall, your code is neat and easy to read! The abstractions/OOP structure is clear, I have no problems understanding what and how the different objects do and interact. Good job :)

PR #191, opened 2019-08-28
[Tan Wei Yang] Duke Increments

weiyang13:


src/main/java/duke/task/DatedTask.java
```diff
@@ -0,0 +1,42 @@
...
+import duke.DukeException;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * Encapsulates a Task that is associated with a date.
+ */
+public abstract class DatedTask extends Task {
```
Davidcwh, line 12: Personally did not think of creating another dated task class for event and deadline class to inherit from. Good application of OOP concepts! 👍

Davidcwh:
Hello,

Overall your code is well organised and concise. Meaningful variable and method names allowed me to follow the flow of the program without having to explicitly look up the code in them to understand what it does. Classes are well structured and adhered to OOP concepts. Furthermore, your JavaDoc comments are well written which allows readers to understand each class and method easily.

Overall good job for the effort and excellent quality in the code 👍 💯

weiyang13:
Thank you for your review, @Davidcwh! I appreciate your feedback! 😄

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
...
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Runs the chatbot.
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
```
WilliamRyank: Hi, overall ur Duke.java code looks really clean and organized. However, do you think putting the flag isExit in the Command class will make your code even cleaner. So, your while condition can be just !c.isExit(), and in the Command class if the command is actually "bye" u can call c.exit() which will mutate the isExit instance variable of the Command instance object to true.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+
+public class Parser {
+    /**
+     * Parses a user input command as a Command object.
+     *
+     * @param fullCommand String containing a command that is input by user.
+     * @return A Command object that corresponds to fullCommand.
+     * @throws DukeException If fullCommand is not properly formatted.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
```
WilliamRyank, line 25: Well done on this method. It is really well organized and its easy for the reader to see your code. Keep it up!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+                        task.setIsDone(true);
+                    }
+                    tasks.add(task);
+                    break;
+                default :
+                    break;
+                }
+
+            }
+        } catch (FileNotFoundException e) {
```
WilliamRyank, line 96: Nice approach to tackle the FileNotFoundException!!

WilliamRyank (commented):
Overall your code looks really organized and personally i learnt a few interesting approaches that i never thought before. Keep up the good work!!

src/main/java/duke/Duke.java
weiyang13: That make sense, I will see if I can integrate it after implementing JavaFX (assuming the code structure is still similar).
Thank you for the feedback!

weiyang13:
Thank you for the review, @WilliamRyank ! I appreciate the feedback and will try to implement the relevant changes where suitable.

PR #194, opened 2019-08-28
[Liew Jun Wei] Duke Increments

bigjunnn:


src/main/java/Task.java
```diff
@@ -1,15 +1,24 @@
...
+  public void setAsDone() {
+    this.isDone = true;
   }
 
   public String toString() {
-    return this.description;
+    return "[" + this.getStatusIcon() + "] " + this.description;
+  }
+
+  public String getStatusIcon() {
```
ambhinav: good practice to abstract the status icon +1

src/main/java/Parser.java
```diff
@@ -0,0 +1,238 @@
...
+      s.appendToFile(text);
+
+    } else if (command.equals("delete")) {
+      if (taskNum < 0 || taskNum >= tasks.size()) {
+        throw new DukeException("OOPS! Integer is out of range of list.");
+      }
+      Task removed = tasks.get(taskNum);
+      tasks.remove(taskNum);
+
+      System.out.println("\t____________________________________________________________");
```
ambhinav: I think it would have been better to abstract adding the borders to a separate method so as to reduce repeated code. At least that is how i did it.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,46 @@
...
+      this.s = s;
+      Scanner sc = new Scanner(s.load());
+      while (sc.hasNextLine()) {
+        String command = sc.next();
+        String description = sc.nextLine().stripLeading();
+
+        Parser p = new Parser(command, description);
+        p.executeOnly(this);
+      }
+    } catch (FileNotFoundException e) {
```
ambhinav: Perhaps you could create the file if it is not found. I thought that would be better as the user does'nt need to go create the file later

src/main/java/Task.java
```diff
@@ -9,6 +13,9 @@ public Task(String description) {
     this.isDone = false;
   }
 
+  /**
+   * Sets the task to be done.
```
ambhinav: i think u meant "Set the task as done" instead?

ambhinav:
Hey Jun Wei, I think your code was very easy to follow, especially since you documented it well. Also you commited frequently and with clear messages which is something I need to learn from u lol

- I think you made good use of all the OOP concepts. You managed to extract all the necessary classes and made the main method look clean.

- Your adherence to coding standard was great and it made it very easy to read the code

Improvments I thought could be made:

- Creating an abstract commands class and separate commands that can extend from it. (At least that's how I did it) So that later on if you want to add more commands you don't need to repeat common code

- The rest i wrote as in-line comments


Thanks

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,102 @@
...
+
+        String output = "";
+
+        switch (type) {
+        case "todo": {
+            // Create ToDo object
+            ToDo t = new ToDo(taskDescription);
+            tasks.addTask(t);
+
+            // Saving to text file
```
TanZhanMing, line 55: I really like how there are comments within the methods themselves instead of just dumping everything in the javadoc comment up top. Makes the code easier to understand and aids in debugging.

src/main/java/utils/Storage.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Deletes the task specified at index in the text file.
+     *
+     * @param index the specific index of task to be deleted.
+     */
+    public void deleteTask(int index) throws IOException {
+        int count = 0;
+        StringBuilder finalText = new StringBuilder();
```
TanZhanMing, line 110: From what I can gather (please correct me if I am wrong), you are trying to gather all the data from the existing text file and store them in the variable 'finalText' and then override the existing data when there are changes.
I feel that there might be pros and cons to this.

Pros: saves space and computational power as there is no need to create a separate temp.txt file and then replace the tasks.txt file every time there is an update
Cons: On the off chance the data set from tasks.txt is too big (2^32-1), the 'finalText' String might not be able to hold them.

TanZhanMing (commented):
Hello Jun Wei,

First of all, I would like to apologise as I had actually commented last week but at the wrong place. I commented at https://github.com/nus-cs2103-AY1920S1/duke/pull/194 instead of here and thus I just copied everything I posted there here. Once again, I am truly sorry for my late review and feedback.

Your code was very nicely written and it shows how much you value clarity in your work. There were also quite a number of methods that you used which I could definitely pick up on to improve my own code base.

The good points:

Really easy to understand code. Your code was very modular in nature which aids in comprehension as I dislike reading long blocks of code. Good usage of white-space between lines, comments within methods to keep readers up to speed etc.
Well-implemented methods. Took me a while to figure out what you were trying to do with the 'getSuffix' method but once I did, I found it to be really creative. Other methods were also clear and concise.
Points for improvement:

Maybe try and utilize switch-case instead of if-else blocks for some of the methods. It can make the code look neater.
Maybe intentionally leave a few mistakes here and there so reviewers don't have to look so hard for points for improvements? I'm just kidding.
All in all, really impressive on your part; I feel like I have learned quite alot from this.

PR #195, opened 2019-08-28
[Chow Jia Ying] Duke Increments

C-likethis123:


src/main/java/DateTime.java
```diff
@@ -0,0 +1,99 @@
+/**
+ * DateTime represents dates specified in Deadlines and Event Tasks.
+ */
+public class DateTime {
```
walnads: Just curious why you didn't use the DateTime class to implement this instead!

wpuffs: Hi, just curious why you didn't use the DateTime class to implement this instead!

src/main/java/ToDo.java
```diff
@@ -0,0 +1,9 @@
+/**
+ * This class represents ToDo tasks which is a type of Task.
+ */
+public class ToDo extends Task {
+    public ToDo(String description) {
+        super(description);
+        this.typeOfTask = "[T]";
```
wpuffs: Not really a suggestion again, just curious why there's a need for a typeOfTask variable when you already have the different types of tasks as individual classes inheriting the Task class!

wpuffs (commented):
Hey Jia Ying,

Overall a job well done! I liked how there was extensive OOP usage and your Javadocs was clear in explaining what each class and method does. Just a few comments, 
- I think you can consider adding files like graddle-wrapper.jar, graddle-wrapper.properties, gradlew, etc. to your .gitignore
-  You have a few Command class files, you could consider packaging them into duke.command so that it looks slightly neater! 😄

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Marks the current Task as done.
+     */
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    @Override
+    public String toString() {
+        String statusIcon = isDone ? "[\u2713]" : "[\u2718]"; // \u2713 represents a tick, \u2718 represents a cross
```
jxofficial, line 31: Might have exceeded character count for this line

jxofficial (commented):
Hi Jia Ying, 

I was very impressed with your code; the code in each class corresponds directly with my expectation of what that class should do. I especially like how your` UI `class handles the displaying of the errors passed, instead of leaving it to the `Exception` classes to print out the error. Also, your use of the `DateTime` class is really quite inspired, with many useful methods which cleans up the input dates well.

I just have one small concern. Your `showError` method in the `UI` class seems to be unused. Correct me if I'm wrong! 

Otherwise, I feel that your code is ready for production use. 

Looking forward to your reply.

Jun Xian

PR #196, opened 2019-08-28
[Yuan Xinran, Stanley] Duke Increments

SoilingRogue:


src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+                ui.showLine();
+            }
+        }
+    }
+
+    String getResponse(String input) {
+        return null;
+    }
+
+//    public static void main(String[] args) {
```
woon17: as we are using git, we can add commit to delete the unwanted code if u want, then easier to read and analyse code.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+                break;
+            }
+        }
+    }
+
+    public TaskList() {
+        lst = new LinkedList<>();
+    }
+
+    /**
```
woon17, line 63: javaDoc can be more detail

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,148 @@
...
+     * Prints farewell message when "bye" is entered.
+     */
+    public void showExit() {
+        printStandard("Bye. Hope to see you again soon!");
+    }
+
+    /**
+     *
+     */
+    public void showLoadingError() {
```
woon17: what about adding some information inside, like "duke.txt doesn't seem to be there or is incorrect"

woon17 (commented):
As outlined above, you can make some changes if you agree with me. Good attempt, I also learnt from your code design. Thanks!

data/tasks.txt
```diff
@@ -2,3 +2,4 @@ T | 0 | sleep
 D | 0 | 2103T ip| 28 of August 2019, 11:59 PM
 E | 1 | project meeting| Friday 2101 classroom
```
tiuweehan, line 3: As this is a local storage file, it might not be a good idea to commit this file, as it could potentially contain sensitive information. You can add the relative directory of this file to .gitignore file, so it will be ignored by git and not be committed.

src/main/java/Duke.java
```diff
@@ -1,10 +1,28 @@
 import duke.*;
```
tiuweehan: It would be a good idea to avoid using wildcard imports, as it is part of the java coding standard we are using for the module.

tiuweehan:
**Strengths**
I think it is quite helpful that you packaged your files into meaningful directories e.g. task, command, as it makes the overall application more intuitive and easily navigable.

I also like that you created more layers of abstraction for the commands inheritance hierarchy, such as TodoCommand < AddCommand < Command instead of TodoCommand inheriting directly from Command. This allows for more specific features to be added on to the AddCommand class without having to add on to the Command class, increasing reusability.

**Areas of improvement**
It would be good to provide more descriptive JavaDoc comments, including more description for tags such as @param, @return and @exception. This will improve readability and maintainability of your application.

Another point brought up by woon17 was that you should try to avoid commenting out unfinished code and committing them to master. If the feature you intend to implement is large, try to do it in a separate branch instead and merge it in later on, and add changes incrementally.

PR #198, opened 2019-08-28
[Tan De Ning, Klement] Duke Increments

klementtan:


src/main/java/AddTaskCommand.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Helper function that maps the input to the relevant task
+     * @throws DukeException
+     */
+    private void getTask() throws DukeException {
+        //Task Details is all the details for the task
+        try {
+            String taskCat = taskDetailsString.split(" ")[0];
+            ;
```
lightz96, line 24: What does semicolon use for?

src/main/java/Command.java
```diff
@@ -0,0 +1,24 @@
+public abstract class Command {
+    int actionType; // -1=>null 0=>add 1=>delete 2=>list 3=>exit 4=>done 5=>find
```
lightz96: You could use Enum to represent all these commands.

src/main/java/DeleteTaskCommand.java
```diff
@@ -0,0 +1,25 @@
+public class DeleteTaskCommand extends Command {
+    int index;
+
+    /**
+     * The delete task command
+     * @param index is the index from the list of the item to delete
+     */
+    DeleteTaskCommand(int index) {
+        super(1);
+        this.index = index-1;
```
lightz96, line 11: Missing whitespaces. It should be written as `this.index = index - 1;`

src/main/java/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * This method takes in exactly what the user types and determine if should return a command or an error
+     * @param fullCommand string of the input
+     * @return  the relevant command for that string
+     * @throws DukeException this error will be raised if the user does not enter a valid input
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        if (fullCommand.equals("bye")){
```
lightz96, line 13: Missing whitespaces between ")" and "{"

src/main/java/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+
+    }
+
+    /**
+     * Use the Buffered File provided in the constructor to read the file and populate the list with the stored
+     * tasks
+     */
+    private void populateTaskList() {
+        Stream<String> linesStream = bufferedReader.lines();
+        linesStream.forEach(line -> {
```
lightz96, line 27: I like your idea of using streams to process the input.

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,27 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DukeTest {
+    @Test
+    public void addDeadlineTaskCommandTest() throws DukeException {
```
lightz96, line 7: Perhaps you could split these three methods into 3 different classes since they are testing different classes in your Duke program.

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,27 @@
...
+        String fullCommand = "deadline return book /by 2/12/2019 1800";
+        Task actualTask = new AddTaskCommand(fullCommand).currTask;
+        Task expectedTask = new Deadline("return book ", "2/12/2019 1800");
+        assertEquals(actualTask.getTaskDetails(), expectedTask.getTaskDetails());
+    }
+
+    public void addEventTaskCommandTest() throws DukeException {
```
lightz96, line 14: You should include the `@Test` annotation for each test methods.

lightz96 (commented):
Overall, you had made good use of OOP by refactoring your code into multiple classes. However, I realized that some of your class variables are not using private access modifiers. This violates the encapsulation principle. In addition, I also noticed some coding standards violations, like missing whitespaces and incomplete JavaDoc comments.

src/main/java/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+    static Command parseDeleteCommand(String string) {
+        int index = Integer.parseInt(string.split("delete ")[1]);
+        return new DeleteTaskCommand(index);
+    }
+    /**
+     * Helper function to get the relevant command
+     * @param string input
+     * @return command
+     */
+
```
chensu2436, line 50: The comment here should be with the next method.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+            String[] lineSplit = line.split("//");
+            String taskType = lineSplit[0];
+            Task currTask = new Task("temp");
+            if (taskType.equals("T")){ // Todo
```
chensu2436, line 31: Might be better if use switch cases?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+            tasksArr.add(currTask);
+        });
+    }
+
+    /**
+     * Get the string representation of the full list to write to the file and store the data
+     * @return string
+     */
+    public String taskListString () {
```
chensu2436: Will it be better to do all the formatting in Todo/Deadline/Event classes (eg. having a storageFormat() method in these classes which return a string representation of the format in which you want the task to be stored) to achieve better abstraction?

chensu2436, line 65: Will it be better to do all the formatting in Todo/Deadline/Event classes (eg. having a storageFormat() method in these classes which return a string representation of the format in which you want the task to be stored) to achieve better abstraction?

chensu2436 (commented):
Good job! I like the way you read inputs and create several custom command classes which increases abstraction.
I feel for this project it's more suitable to use switch cases instead of if-else which can make the codes more readable.
You may also consider adding a few custom error classes which inherits DukeException class to increase abstraction.

PR #199, opened 2019-08-28
[Ang Chin Guan, Melvin] Duke Increments

krusagiz:
Duke Increments

Incrementing
Incrementing
Crematorium


jun-ha0:
<h2>Code:</h2>

I like the strong usage of OOP and software design principles (Single Responsibility Principle, Separation of Concerns Principle, etc), leading to high cohesion and low coupling. 😃 

1. public abstract boolean isExit() method in Command.java may be written to return false by default,
to shorten code (? at a cost of some readability?)

```
public boolean isExit() {
        return false;
}
```
2. It may be a good idea to consider abstracting the class AddCommand from AddEventCommand, AddDeadlineCommand, AddTodoCommand.

3. It may be a good idea to consider using Java's [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html). 

4. Lastly, good use of enums and various DukeExceptions.

<h2>Style:</h2>

Very clean. Well done! 👍 

Disclaimer: The following are optional. They are just some ideas to make the source code neater.

- It may be a good idea to follow SE-EDU [AB3](https://github.com/se-edu/addressbook-level3/blob/master/docs/SettingUp.adoc) convention for ordering of import statements.


- It may be useful to check out the codestyle import ordering with [IntelliJ.](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003038760-Codestyle-Import-ordering-with-the-new-version-of-IntelliJ-2018-3)

- Lastly, it may be a good idea to add Javadocs for all public classes and methods.

ning-sy210:
### Code
I really like the fragmentation of your code into smaller classes, such that the role of every class is very well-defined and I can see how they come together to make your program as a whole! 😄

I especially like how the different types of commands are being separated into their respective class files and the various exceptions to manage different user inputs. Good job! 👍

    public enum Checkbox {
 
        TICK("[✓]"),
        CROSS("[✗]");
 
        public final String icon;
 
        private Checkbox(String icon) {
            this.icon = icon;
        }
 
    }

I also find it interesting that you have chosen to use `Enum` to represent checkboxes. Personally, I don't find it necessary because to mark whether a task is done can simply be indicated using any form of a binary system (e.g. '0' or '1', '[x]' or '[ ]' etc) so I think it would be more convenient if they are just left alone as `String` representations instead.

Here are a few suggestions for code readability improvement:

- You can consider using switch statements for `parse(String input)` in `Parser.java` as opposed to using many `if-else` statements.
- You can consider using a line separator to separate code with different logic in the same block. An example is shown below (taken from `AddEventCommand.java`):


        public void execute(TaskList tasks, Ui ui, Storage storage) {
            try {
                Date datetimeFromDate = super.DATE_FORMAT.parse(datetimeFrom);
                Date datetimeToDate = super.DATE_FORMAT.parse(datetimeTo);
                Event newEvent = new Event(detail, datetimeFromDate, datetimeToDate);
                tasks.add(newEvent);
                duke.ui.Ui.printAddSuccess(tasks.getTasks(), newEvent);
            } catch (ParseException dfe) {
                ui.exposeError(new DukeEventException(DukeEventException.FORMAT_ERROR_MSG).getMessage());
            }
        }

Can be changed into:

        public void execute(TaskList tasks, Ui ui, Storage storage) {
            try {
                Date datetimeFromDate = super.DATE_FORMAT.parse(datetimeFrom);
                Date datetimeToDate = super.DATE_FORMAT.parse(datetimeTo);

                Event newEvent = new Event(detail, datetimeFromDate, datetimeToDate);
                tasks.add(newEvent);

                duke.ui.Ui.printAddSuccess(tasks.getTasks(), newEvent);
            } catch (ParseException dfe) {
                ui.exposeError(new DukeEventException(DukeEventException.FORMAT_ERROR_MSG).getMessage());
            }
        }

As you can see, now it is much clearer to your readers what you are trying to accomplish in this block of code!

So although they are just empty lines, they are really useful in separating different logic in your code and can improve others' ease of reading and understanding drastically! 😎

### Coding Style

1. Please do remember to add Javadoc for all public classes! This gives your readers a better idea of the significance of the respective classes that you have chosen to implement in your program.

1. Javadoc descriptions for exceptions can be more descriptive! This gives readers a better idea of the kind of error that a particular exception is handling.

        /**
         * Constructs the DukeFindException object with a custom error message.
         * @param message Custom error message
         */
        public DukeFindException(String message) {
            super(message);
        }

For example, one suggestion for the above Javadoc can be:

        /**
         * Creates a DukeFindException that is thrown to the user when no tasks can be found with input keyword.
         * @param message Custom error message
         */
        public DukeFindException(String message) {
            super(message);
        }

Other than that, your code style follows the Java Coding Standard very closely. It is also very neat and tidy so Good Job Well Done! 👍

krusagiz:
Thank You @jun-ha0 and @ning-sy210 for your _awesome_ feedback! 😄

PR #204, opened 2019-08-28
[Tan Ye Kai] Duke Increments

mrnewguy:


gachia (commented):
Great use of a custom DateTime class to generate your own specific date and time format. However, these classes require more lines of code and you may weigh the pros and cons of using it for future projects.

gachia (commented):
I noticed a lot of if statements in your Parser class. Is there a way to reduce the code length and achieve the same result?

gachia:
Great job on the code overall. Do be careful with spelling mistakes in Java Docs and don't be afraid to add more details to the descriptions.

src/main/java/Duke.java
```diff
@@ -25,44 +25,70 @@ public static void main(String[] args) {
...
-                        String stringFromUser = line.replaceAll("\\D+","");
-                        Integer indexFromUser = Integer.parseInt(stringFromUser);
-                        if(indexOfTask == indexFromUser ){
-                            Task currentTask = listOfStuff.get(i);
-                            currentTask.setStatus(true);
-                            String message = "Nice! I've marked this task as done:\n" + currentTask;
-                            System.out.println(message);
+
+                try{
+                    if((!inputnoun.equals("done")) &&
```
brianyenna: Perhaps it might be better to have some separation of concerns here. For each different value for the `inputnoun`, maybe you can consider calling a different function, so that the functions encapsulate the logic required for each `inputnoun` value. 

Also, do be careful with the spelling. For instance, your `Level-5` tag has an extra space in the tag.

brianyenna (commented):
For the first few levels, overall great code! However, perhaps there can be more separation of concerns (by separating different bits of logic into designated methods and even specific objects)

src/main/java/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+                        Integer indexFromUser = Integer.parseInt(stringFromUser);
+                        taskList.doneTask(indexFromUser - 1);
+                        storage.save();
+                            
+                    }else if (inputnoun.equals("todo")){
+                        ToDoTask newTask = new ToDoTask (false, taskName);
+                        taskList.add(newTask);
+                        storage.save();
+
+                    }else if (inputnoun.equals("deadline")){
```
brianyenna: Perhaps you can consider placing the logic of the Date object creation, along with the associated error handling, within the constructor of the DeadlinesTask object. This way, your code for Parser can be cleaner.

brianyenna (commented):
Great conversion to OOP. I think you handled very cleanly. However, perhaps a minor improvement could be to consider using the in-built Java libraries for handling date and time. They offer very similar functionality to the DateTime class you implemented.

src/main/java/FindCommand.java
```diff
@@ -1,17 +1,19 @@
...
+    }
+
+    @Override
+    public String execute(String processedCommand){
         TaskList taskListsWithKeyWords = new TaskList();
         
-        int length = taskList.size();
+        int length = super.taskList.size();
 
         for(int i = 0; i < length; i++){
```
brianyenna: Perhaps it might be possible to implement the <Iterable> interface for the TaskList object to make the iteration across all tasks easier.

src/main/java/MainWindow.java
```diff
@@ -54,7 +54,12 @@ public void setDuke(Duke duke) {
     @FXML
     private void handleUserInput() {
         String input = userInput.getText();
-        String response = duke.getParser().parse(input);
+        String response;
+        try{
+            response = duke.getParser().parse(input);
+        }catch(DukeException e){
```
brianyenna: The parentheses here should have a whitespace separator. This is a coding style issue, but one that can be easily fixed through automation. May I suggest using the in-built code style and its refactoring ability in IntelliJ? :)

brianyenna (commented):
Good changes overall! Gave me some good ideas as well. Thank you for that :)

damithc:
Please rectify the PR title to match the requirement. The PR author can edit the PR title.

PR #205, opened 2019-08-28
[Ng Choon Wah] Duke Increments

choonx99:
Duke Increments for Duke 💃


src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public DateTime(String dateTime) throws InvalidInputException {
+        String[] splitDateAndTime = dateTime.split(" ");
+        String[] splitDate = splitDateAndTime[0].split("/");
+        String day = splitDate[0];
+        String month = splitDate[1];
+        String year = splitDate[2];
+
+        if (day.equals("1") || day.equals("21") || day.equals("31")) {
```
junnbang: Great way to add the suffix of date as there's no java api that supports this!

junnbang: Great way to add the suffix of the date since there is no java api to support this.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+        this.time = time;
+    }
+
+    public DateTime(String dateTime) throws InvalidInputException {
+        String[] splitDateAndTime = dateTime.split(" ");
+        String[] splitDate = splitDateAndTime[0].split("/");
+        String day = splitDate[0];
+        String month = splitDate[1];
+        String year = splitDate[2];
+
```
junnbang: Great way to add the suffix of the date since there is no java api to support this.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+        if (day.equals("1") || day.equals("21") || day.equals("31")) {
+            day = day + "st";
+        } else if (day.equals("2") || day.equals("22")) {
+            day = day + "nd";
+        } else if (day.equals("3") || day.equals("23")) {
+            day = day + "rd";
+        } else {
+            day = day + "th";
+        }
+
```
junnbang: I would like to suggest using the Java 8 API "java.time" to parse the date string to check if the date is valid and get the date string in return. By parsing the date string using a date time format, it will check if the date string is in a valid format, otherwise, it will throw an error.
Example to check if date string is valid:
> try {
            String dateStr = "32/12/2019 1800";
            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
            LocalDateTime dateTime = LocalDateTime.parse(dateStr, formatter);
            System.out.println(dateTime);
        } catch (DateTimeParseException e) {
            System.err.println(e);
        }

Example to convert date to String format to save:

> LocalDateTime dateTime = this.getLocalDateTime();
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d MMMM yyyy, h:mma");
        String formattedDateTime = dateTime.format(formatter);

This way, you wouldn't have to check for the date time format manually. You can then add the suffix of the date string.

src/main/java/DateTime.java
junnbang: Great way to add the suffix of the date since there is no java api to support this.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+        switch (month) {
```
junnbang: I would like to suggest using the Java 8 API "java.time" to parse the date string to check if the date is valid and get the date string in return. By parsing the date string using a date time format, it will check if the date string is in a valid format, otherwise, it will throw an error.
Example to check if date string is valid:
> try {
            String dateStr = "32/12/2019 1800";
            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
            LocalDateTime dateTime = LocalDateTime.parse(dateStr, formatter);
        } catch (DateTimeParseException e) {
            System.err.println(e);
        }

Example to convert date to String format to save:

> LocalDateTime dateTime = this.getLocalDateTime();
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d MMMM yyyy, h:mma");
        String formattedDateTime = dateTime.format(formatter);

This way, you wouldn't have to check for the date time format manually. You can then add the suffix of the date string.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+        String[] splitDateAndTime = dateTime.split(" ");
+        String[] splitDate = splitDateAndTime[0].split("/");
+        String day = splitDate[0];
+        String month = splitDate[1];
+        String year = splitDate[2];
+
+        if (day.equals("1") || day.equals("21") || day.equals("31")) {
+            day = day + "st";
+        } else if (day.equals("2") || day.equals("22")) {
+            day = day + "nd";
```
junnbang, line 33: Great way to add the suffix of the date since there is no java api to support this.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,146 @@
...
+        } else if (day.equals("3") || day.equals("23")) {
+            day = day + "rd";
+        } else {
+            day = day + "th";
+        }
+
+        switch (month) {
+        case "1":
+            month = "January";
+            break;
```
junnbang, line 45: I would like to suggest using the Java 8 API "java.time" to parse the date string to check if the date is valid and get the date string in return. By parsing the date string using a date time format, it will check if the date string is in a valid format, otherwise, it will throw an error.
Example to check if date string is valid:
> try {
            String dateStr = "32/12/2019 1800";
            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
            LocalDateTime dateTime = LocalDateTime.parse(dateStr, formatter);
        } catch (DateTimeParseException e) {
            System.err.println(e);
        }

Example to convert date to String format to save:

> LocalDateTime dateTime = this.getLocalDateTime();
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d MMMM yyyy, h:mma");
        String formattedDateTime = dateTime.format(formatter);

This way, you wouldn't have to check for the date time format manually. You can then add the suffix of the date string.

src/main/java/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+            } else if (taskType == 'D') {
+                String date = "";
+                String time = "";
+                String description = line.substring(7, line.indexOf('(') - 1);
+                String by = line.substring(line.indexOf(':') + 2, line.indexOf(')')).trim();
+                String[] dateTimeSplit = by.split(" ");
+                if (dateTimeSplit.length == 3) {
+                    date = dateTimeSplit[0] + " " + dateTimeSplit[1] + " " + dateTimeSplit[2].substring(0, dateTimeSplit[2].length() - 1);
+                    time = "";
+                } else if (dateTimeSplit.length == 4) {
```
junnbang: For concatenating date and time for Event and Deadline, it may be abstracted so that it can be re-used.

src/main/java/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+     * Checks validity of users' input.
+     * @param type refers to the commands of user inputs e.g., "done", "delete", "deadline"
+     * @param userInput refers to the actual input sentence from user
+     * @throws EmptyFieldException occurs when there are missing data from user input
+     * @throws InvalidInputException occurs when user input is invalid
+     * @throws ParseException occurs when user input date is invalid
+     */
+    public static void handleInput(String type, String userInput) throws
+            EmptyFieldException, InvalidInputException, ParseException {
+        if (type.equals("find")) {
```
JermyTan, line 21: I like how you organise the checkings of the user input in a single file. :D
One suggestion would be to use a switch case here to make the code look neater.

src/main/java/Storage.java
```diff
@@ -0,0 +1,105 @@
...
+                Task deadline = new Deadline(description, new DateTime(date, time));
+                if (isDone == '0') {
+                    taskList.add(deadline);
+                } else {
+                    deadline.markAsDone();
+                    taskList.add(deadline);
+                }
+            } else if (taskType == 'E') {
+                String date = "";
+                String time = "";
```
JermyTan: Seems like there is a repeat of code for when taskType == "E" and taskType == "D".
Perhaps you could abstract out the code in a function for easier reuse.

PR #213, opened 2019-08-28
[Yeo Yuan Xin] Duke Increments

yyuanxin:


yaaanch:
Your program is really well-structured. I think the OOP breakdown is extremely well done.

I think to further the OOP, one possibility would be to have more types of exceptions of different classes that extend DukeException. Then instead of changing the message used to construct the DukeException, objects from the classes with default messages could just be thrown.

It's great that you've written most of the Javadocs comments. For Javadocs comments, you may want to look at the style guide as well. So for example, capitalisation and starting functions with singular verbs. Another addition would be using the @ keywords such as @param and @return.

Overall, it's really well done!

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,72 @@
...
+                this.task = new Deadline(arr[0].trim(), Parser.datetimeformat(arr[1].trim()));
+                this.type = "D";
+            } else if (type.equals("E")) {
+                String[] arr = command.split("/at");
+                this.task = new Event(arr[0].trim(), arr[1].trim());
+                this.type = "E";
+            } else {
+                throw new AssertionError("ERROR: Invalid Task Type!");
+            }
+        } catch (DukeException e) {
```
Yoshi275: Perhaps you don't need to `catch` the `DukeException` to throw it again? You can just `throw` the exception directly.

src/main/java/Parser.java
```diff
@@ -0,0 +1,169 @@
...
+    public static Command parse(String fullcommand) throws DukeException {
+        try {
+            String subcommand = fullcommand.split(" ")[0];
+            String command = fullcommand.replaceFirst(subcommand, "").trim();
+            String type;
+            if (Alias.aliases.containsKey(subcommand)) {
+                type = Alias.aliases.get(subcommand);
+            } else {
+                throw new DukeException("       OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
```
atharvjoshi: Hello! Our group reviewed your code during our tutorial. Your parse method seems too long - perhaps you could abstract the splitting of the command and the matching of the aliases to another method called by parse()? Cheers!

src/main/java/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+            BufferedReader bufferedReader = new BufferedReader(reader);
+
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                String[] arr = line.split(" \\| ");
+                Task task;
+
+                switch (arr[0]) {
+                case "T": //todo
+                    task = new Todo(arr[2]);
```
meiannn: Hi, you may want to consider extracting the checking of task type out, so that the switch block doesn't have to be repeated in many different classes

src/main/java/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Edits existing tasks in file through delete or update
+     *
+     * @param task Task to save in file
+     * @param command String command from user input
+     * @param position Index of task in TaskList
+     * */
+    public void updateTasks(Task task, String command, int position) {
```
madanalogy: Hi, so we were reviewing your code during our tutorial, and do you think it might be better if you abstract out certain portions of this method into separate methods? Otherwise it's really good stuff great job!

PR #214, opened 2019-08-28
[Daniel Wong Yong Shen] Duke Increments

danielwys:


madanalogy:
Hi, I really like how thorough you are with your tests! Also, your introduction of a custom DukeException was really ingenious. As a suggestion, you could consider further abstracting the different commands in TaskList.java into individual classes to provide further OOP separation.

eugenood:
Good job on your contribution.

I am concern about the `duke.command` package that is introduced in commit 165d55c. The package name seems to be inappropriate for the contents it contains. I believe the intention of the `duke.command` package is to contain all your `Command` classes, but since you did not make the `Command` abstraction in 7785512, this package is redundant.

danielwys:
> Good job on your contribution.
> 
> I am concern about the `duke.command` package that is introduced in commit [165d55c](https://github.com/nus-cs2103-AY1920S1/duke/commit/165d55c9a64ee0e6f0307a2edf15e099d928d379). The package name seems to be inappropriate for the contents it contains. I believe the intention of the `duke.command` package is to contain all your `Command` classes, but since you did not make the `Command` abstraction in [7785512](https://github.com/nus-cs2103-AY1920S1/duke/commit/77855120122fd28927527a5883ac05542e2c65d1), this package is redundant.

Hi! Thanks for your comment. Yes, you're right that I didn't end up making the Command abstraction due to the lack of time last week. I'll consider doing that! 

I had originally intended to do so, but decided instead that I'd group them into two types of classes - those that are tasks (duke.task) and those that issue commands (duke.command), hence the current class grouping. But I'll take your suggestion into account :)

danielwys:
> Hi, I really like how thorough you are with your tests! Also, your introduction of a custom DukeException was really ingenious. As a suggestion, you could consider further abstracting the different commands in TaskList.java into individual classes to provide further OOP separation.

Yes definitely, I had left that out due to a time crunch last week. Thanks for your comment!

PR #215, opened 2019-08-28
[Brian Fung] Duke Increments

briyanii:
My week 3 increments for Duke individual project.


bernicechio:
- I appreciate how your parser test case is very detailed
- Code is well written 👍🏻
- Good use of lamda expressions and enumeration

CowSaysBaa:
I am no expert at software engineering but here's my 2 cents.
Code is well formatted. The parsing operation is quite unintuitive but unique. To be honest, the code is rather complicated, which makes reading and understanding very challenging, but I can appreciate the effort that went into the abstraction of data and the preparation for changes in the future, especially for the Type enum which I guess guarantees type safety.

CowSaysBaa (commented):
I am no expert at software engineering but here's my 2 cents.
Code is well formatted. The parsing operation is quite unintuitive but unique. To be honest, the code is rather complicated, which makes reading and understanding very challenging, but I can appreciate the effort that went into the abstraction of data and the preparation for changes in the future, especially for the Type enum which I guess guarantees type safety.

PR #216, opened 2019-08-28
[Leonard Tay] Duke Increments

LeonardTay748:


src/main/java/Duke.java
```diff
@@ -1,10 +1,269 @@
...
+
 public class Duke {
-    public static void main(String[] args) {
+
+    private Storage storage;
+    //private TaskList tasks;
+    private Ui ui;
+
+    /////////////////////////// Date & Time Converter: returns the converted format as a string //////////////////////////
+    static String Convert(String timeframe){
```
shiweing: You might want to change method name to camelCase to match coding standard.

wongchuankai:
Overall, easy to follow through and readable :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,269 @@
+import org.w3c.dom.html.HTMLImageElement;
+
+import java.io.*;
```
openorclose: Try not to use wildcard imports.

src/main/java/Duke.java
```diff
@@ -1,10 +1,269 @@
...
+          int day = Integer.parseInt(timeframe.substring(0, in));    //sift out day
+          String sub1 = timeframe.substring(in+1);                   //substring: month onwards
+          int in2 = sub1.indexOf('/');
+          int month =  Integer.parseInt(sub1.substring(0, in2));     //sift out month
+          String sub2 = sub1.substring(in2+1);                       //substring: year onwards
+          String year = sub2.substring(0, 4);                        //sift out year
+          String time = sub2.substring(5);                           //sift out time (24hr clock format)
+
+          String converted_day;
+          switch (day){
```
openorclose: Try to follow the style guide for switch statements. Case statements should not be indented.

src/main/java/Duke.java
openorclose: You could also try java.time.LocalDateTime to reduce complexity of this method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,269 @@
...
+          return converted_day + " of " + converted_month + " " + year + ", " + converted_hr + converted_min + period ;
+    }
+
+    /////////////////////////////// End of DATE & TIME Converter ///////////////////////////////////////////////////
+
+
+    /////////////////////////////// Duke constructor  ///////////////////////////////////////////////////
+    public Duke(String filePath) {
+        ui = new Ui();
+        storage = new Storage("D:\\madae\\School\\cs2103T\\IdeaProjects\\DUKE\\DukesDiary.txt");
```
openorclose: Try to use relative paths as this may not work for others.

src/main/java/Duke.java
```diff
@@ -1,10 +1,269 @@
...
+        int no_of_task;
+        ArrayList<Task> taskList = new ArrayList<Task>();
+
+        taskList = storage.load();            //load file onto arraylist
+        no_of_task = storage.get_no_task();   //get number of tasks
+
+        while(true) {
+            userInput = ui.read();                 //read user input
+            //////////////////////////
+            if (userInput.equals("bye")) {
```
openorclose: You could use if else if  statements, instead of nested ifs.

Or you could do a switch statement on the first token of the string, but you may need to refactor your code.

src/main/java/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+                objReader.close();
+        } catch (IOException ex) {
+            ex.printStackTrace();
+        }
+    }
+        return taskList;
+  }
+  ///////////////////////////////////END of LOAD FILE method ////////////////////////////////////////////////
+
+  ////////////////////////////////// SAVE FILE Method ///////////////////////////////////////////////////////
```
openorclose: I like your comment style! 🚀

openorclose (commented):
Here are some overarching comments:

Try to follow the CS2103 Style guide to help other people read your ode better.

.class files should not be committed. You can change your .gitignore file to exclude it automatically.

PR #217, opened 2019-08-28
[Chong Zi Kang] Duke Increments

czkay:


src/main/java/Duke.java
```diff
@@ -1,10 +1,37 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+    private Storage storage;
+    private TaskList tasks;
+
+    public Duke() { };
```
marcusteh1238: That semicolon is redundant.

src/main/java/ExitCommand.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * A Command to exit the Duke program.
+ */
+public class ExitCommand extends Command {
+
+    /**
+     * Executes the command to exit the Duke program.
+     *
+     * @param tasks The task list.
+     * @param ui The ui that handles user output.
```
marcusteh1238: This param does not exist anymore and needs to be removed from the Javadoc comment.

src/main/java/FindTaskCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public FindTaskCommand(String keyword) {
+        this.keyword = keyword;
+    }
+
+    /**
+     * Executes the command to find tasks from the task list based on a keyword.
+     *
+     * @param tasks The task list.
+     * @param ui The ui that handles user output.
```
marcusteh1238: This param does not exist anymore and needs to be removed from the Javadoc comment.

src/main/java/PrintListCommand.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * A Command to print the task list.
+ */
+public class PrintListCommand extends Command {
+
+    /**
+     * Executes the command to print the task list.
+     *
+     * @param tasks The task list.
+     * @param ui The ui that handles user output.
```
marcusteh1238: This param does not exist anymore and needs to be removed from the Javadoc comment.

marcusteh1238 (commented):
I really like the use of `Optional`s in your Parser. It makes your `Parser` a lot less messy and it is a very elegant solution to deal with invalid input. I also feel that the test cases' method names are simple to understand.

One thing I have to comment on is that the Javadoc comments need to be updated consistently. Some of them still have `ui` listed as a parameter even though you have already removed it from the method itself. 

Additionally, in your `Duke` class, the constructor has an unnecessary semicolon which can be removed, where it should look like this: `public Duke() {  }` instead of `public Duke() {  };`. In the `getResponse()` method, I think it would be better if you could catch the `IOException` earlier and create a custom response for it, so that the user could understand what they did wrong instead of seeing a Java generated `Exception` message.

In the test case `deleteTask_wrongIndex_exceptionThrown()` for the `TaskListTest` class, it would better if you tried to delete a task in index 0 rather than in index 1, just in case there is an existing task inside the TaskList when you first create it.

Besides that, I feel that your main code is quite organized and I could follow the logic really well. Do work on the improvements I pointed out, add more test cases, and keep up the good work!

src/main/java/Parser.java
```diff
@@ -1,8 +1,88 @@
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
 /**
  * Deals with making sense of the user command.
  */
 public class Parser {
 
+    private static Optional<Command> findOneParameterCommand(String action) throws DukeException {
```
mrchensheng21: Good use of Optional to help make the code cleaner. Able to also handle null values more efficiently

src/main/java/FindTaskCommand.java
```diff
@@ -21,6 +31,11 @@ public void execute(TaskList tasks, Ui ui, Storage storage) throws IOException,
         }
     }
 
+    /**
+     * Returns a boolean value signalling whether the program should exit.
+     *
+     * @return A boolean value indicating whether the program should exit.
+     */
```
mrchensheng21: Can be clear of which boolean value you are returning, eg. (true/false) or (1/0).

damithc:
Please rectify the PR title to match the requirement. The PR author can edit the PR title.

PR #220, opened 2019-08-28
[Cheng Geng] Duke Increments

ChengGeng97:
Duke Increments for Week 3.


JunHongT:
Like your coding in general. The naming of the variables is descriptive, allowing me to understand the general flow without needing to search for each method to discover more about it. Process had also been segregated to several smaller and concise methods, allowing for minimal repeat of coding. 
However, though i was able to understand the idea of your methods, it might be a great idea to include javadoc for the methods, instead of a 1-liner comment? It might be a huge hassle but i will believe that it will make it clearer for the someone who had no idea about the project to have a better understanding.

src/main/java/ToDoTask.java
```diff
@@ -0,0 +1,48 @@
...
+    
+    @Override
+    //Returns a copy of this task but with its completion status marked as done
+    public Task getTaskMarkedAsDone() {
+        return new ToDoTask(description, true);
+    }
+
+    @Override
+    //Returns a copy of this task but with its completion status marked as undone
+    public Task getTaskMarkedUndone() {
```
JunHongT: May i ask about the situation of when we are required to use this method? Is it an addition function for the user or...
?

ChengGeng97: It's meant to be an additional feature, for cases where you accidentally marked something as done when it's not actually done. It hasn't been fully implemented though, which is a blunder on my part for having pushed it.

ChengGeng97:
> Like your coding in general. The naming of the variables is descriptive, allowing me to understand the general flow without needing to search for each method to discover more about it. Process had also been segregated to several smaller and concise methods, allowing for minimal repeat of coding.
> However, though i was able to understand the idea of your methods, it might be a great idea to include javadoc for the methods, instead of a 1-liner comment? It might be a huge hassle but i will believe that it will make it clearer for the someone who had no idea about the project to have a better understanding.

Thanks for taking the time to read through my code! Javadoc comments were kept to only the public methods for this build since only those would be generated for the Javadoc. I understand the sentiment though, and will get around to adding them for all private methods as well if time permits.

src/main/java/DukeTextFormatter.java
```diff
@@ -0,0 +1,37 @@
+/**
+ * A class that handles text formatting for Duke's responses to the user.
+ */
+public class DukeTextFormatter {
+    private static final String BAR = "____________________________________________________________\n";
+    private static final String FOUR_SPACE = "    ";
+    private static final String FIVE_SPACE = "     ";
+
```
tanamaroby: Very epic! Nice work!

src/main/java/DukeUi.java
```diff
@@ -0,0 +1,70 @@
+/**
+ * A class that stores all of Duke's mesages as Strings. These are split into Greetings, Feedback and Error Messages.
+ */
+public class DukeUi {
```
tanamaroby: This is very great! Nice one ChengGeng97!

tanamaroby (commented):
Overall your code has a little something for everyone. Great use of Object Oriented Principles which allows reviewers like me to read your code with much ease, that and the descriptive Javadoc really helped. The methods and variables are named appropriately which is descriptive enough that the code is readable on its own even without the use of Javadoc. Great use of DukeUi class in which by using all the final static attributes, you can maintain constant reply from Duke and helps with the consistency of the program. 
One criticism that I would make would be that some of the classes are too spread out and too small that most of them just feels unnecessary in my opinion. For example like the DukeTextFormatter class which I understand is used to maintain constant format of the Duke messages but I feel like this could be combined with the UI class since the logic would be similar. However, this is just my opinion and would not really affect the overall efficiency of the program. 

Conclusion : This code really makes you feel like a software engineer.

src/main/java/DukeTextFormatter.java
tanamaroby: Cool use of private static final attributes!

PR #221, opened 2019-08-28
[Tianyi] Duke Increments

tysg:


src/main/java/com/tysng/duke/exception/CommandException.java
```diff
@@ -0,0 +1,10 @@
+package com.tysng.duke.exception;
+
+/**
+ * This exception is thrown during the parsing of command line input to Command class.
+ */
+public class CommandException extends IllegalArgumentException {
```
gary-lgy, line 6: If this exception is only thrown for parsing errors, maybe it could be renamed to something like `CommandParseException`?

src/main/java/com/tysng/duke/storage/Storage.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            List<String> entries = Files.readAllLines(this.storageFilePath);
+            return entries.stream().map(this::parse).flatMap(Optional::stream).collect(Collectors.toList());
+        } catch (IOException e) {
+            e.printStackTrace();
+            return List.of();
+        }
+    }
+
+    private String toFileEntry(Task input) {
```
gary-lgy, line 121: IMO this should be implemented as a method in the `Task` classes. Any reason this approach is chosen instead?

src/main/java/com/tysng/duke/ui/Command.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Creates a new command from the user input. This static factory method parses a line of user
+     * input into a Command object.
+     *
+     * @param instruction a line of user input
+     * @return a command class with the user input wrapped inside
+     * @throws CommandException if the user input cannot be successfully parsed
+     */
+    public static Command newCommand(String instruction) throws CommandException {
+        SimpleDateFormat dateFormat = new SimpleDateFormat("d/MM/YYYY HHmm");
```
gary-lgy: Maybe this should be a constant?

src/main/java/com/tysng/duke/ui/Command.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This class parses raw user input text into a Command object that can be process by Duke.
+ */
+public class Command {
+    private CommandType type;
```
gary-lgy: IMO subclassing `Command` might be a cleaner approach than using enum and switch statements. It would avoid code duplication and cyclomatic complexity.

src/main/java/com/tysng/duke/ui/Response.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * This class formats the output of the Application. It wraps the lines of texts inside the class, and provides a
+ * convenient print() method to print to stdout.
+ */
+public class Response {
```
gary-lgy, line 14: This class is pretty neat.

src/main/java/com/tysng/duke/service/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+        this.taskList = new TaskList(this.storage.loadTasks());
+    }
+
+    /**
+     * Processes a Command object. This is the driver method of the class.
+     *
+     * @param command a successfully parsed user command
+     * @return the corresponding response for the given command
+     */
+    public Response handle(Command command) {
```
gary-lgy: It might be cleaner to subclass `Command` and implement an `execute` method for each subclass instead of using enum and switch IMO.

src/main/java/com/tysng/duke/Main.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * This class is the entry point of the Application.
+ */
+public class Main {
+    /**
+     * Starts the Application.
+     * @param args command line args
+     */
+    public static void main(String[] args) {
+        Storage storage = Storage.initialize();
```
gary-lgy: Can this be done in the `Duke` constructor?

gary-lgy (commented):
Hi Tianyi,
I think your project is very well structured and the code is very clean and self-explanatory. However, I think more OOP can reduce some code duplication and complexity.

src/main/java/com/tysng/duke/storage/Storage.java
tysg: Storage class handles the parsing and serializing of the Task class. It takes in and spits out Task objects. Serializing is done in this layer because the Task class should not be concerned with how it is represented in the Storage (or the database, in actual Java dev).

src/main/java/com/tysng/duke/ui/Command.java
tysg: Will consider. Thanks for the suggestion!

src/main/java/com/tysng/duke/domain/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package com.tysng.duke.domain;
+
+import java.util.Date;
+
+/**
+ * This class contains a Date by which the Deadline is due.
+ */
+public class Deadline extends Task {
+    private Date by;
```
kohyida1997, line 9: I understand what you're doing here with the implementation. Perhaps you could rename the variable "by" to "dateToCompleteBy" so that it is more self-explanatory!

src/main/java/com/tysng/duke/service/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+            response = this.handleEcho(command.toString());
+            break;
+        default:
+            throw new CommandException("This command is not recognized.");
+        }
+        this.storage.saveTasks(this.taskList.getTaskList());
+        return response;
+    }
+
+    private Response handleFind(String keyword) {
```
kohyida1997: This is a pretty cool way of using streams to search for a keyword!

src/main/java/com/tysng/duke/service/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+        List<Task> matchedTasks = taskList.getTaskList().stream()
+                .filter(task -> task.getTaskName().contains(keyword))
+                .collect(Collectors.toList());
+        return Response.newMatch(matchedTasks);
+    }
+
+    private Response handleDone(int targetIndex) {
```
kohyida1997: Perhaps it might be cleaner to abstract away all of the execution of logic into the Command class itself?

kohyida1997 (commented):
Code is very clean and compact! Nice :)

src/main/java/com/tysng/duke/domain/Deadline.java
tysg: Wilco, thanks!

src/main/java/com/tysng/duke/storage/Storage.java
gary-lgy: Makes sense. TIL.

src/main/java/com/tysng/duke/service/Duke.java
```diff
@@ -60,6 +66,12 @@ public Response handle(Command command) {
         return response;
     }
```
fabbbbbbyy: I haven't encountered anyone else who used streams, so it's nifty. I might consider adding streams in my current Duke as well. Switch cases are appropriate.

fabbbbbbyy:
Hi Tian Yi, I couldn't find this a few days ago. I must have searched for it incorrectly. Your code is simple to understand. I will leave the other comments on the code as I go along.

src/main/java/com/tysng/duke/ui/Response.java
```diff
@@ -72,6 +114,11 @@ public static Response newListing(List<Task> items) {
         return new Response(content);
     }
 
+    /**
+     * Generates a Echo type response.
+     *
+     * @return a response
```
fabbbbbbyy, line 157: I like how you have a Response class for hardcoding static output. I personally only had a UI class in which I hardcoded private static final Strings as the output and called it within the class itself.

PR #227, opened 2019-08-28
[William Ryan] Duke Increments

WilliamRyank:


Davidcwh:
Hello, 

I like your code in general. You used descriptive/meaningful variable and method names which allowed me to easily understand what they were for. Good application of abstraction by having processes segregated into smaller and concise methods. Logic in the code was well structured and similar to how I would have done it.

However, I noticed that the JavaDoc comments for some classes were missing (Event, Deadline, DukeException, Task etc). Although it may seem obvious to include JavaDoc descriptions in some areas that are self-explanatory, a description for the class itself would give readers who have not worked on this project a general idea of what it is at a glance, instead of having to sieve through the code.

Besides this minor issue, good job overall for having since well organised code! 💯 👍

src/main/java/Ui.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getNextLine() {
+        return sc.nextLine();
+    }
+
+    public String getNext() {
+        return sc.next();
+    }
+
+    public void printHello() {
+        System.out.println("    _____________________________________");
```
Davidcwh: Maybe could assign "    _____________________________________" to a String variable and use the same variable when needed.

src/main/java/Duke.java
```diff
@@ -1,10 +1,34 @@
...
+    public void run() {
+        ui.printHello();
+        String command;
+
+        while (!(command = ui.getNext()).equals("bye")) {
+            try {
+                Parser.parse(tasks, ui, storage, command, ui.getNextLine());
+            } catch (DukeException e) {
+                System.out.println("    _____________________________________");
+                System.out.println("     " + e.getMessage());
```
weiyang13: Good job on abstracting the major components of Duke! The logic of Duke seems to flow well. However, you could perhaps consider allowing the UI component to handle error messages as well to compartmentalize the logic more completely.

src/main/java/Storage.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Saves current list of tasks to hard drive
+     *
+     * @param listOfTasks List that will be saved to the file
+     */
+    public static void save(List<Task> listOfTasks) {
+        try {
+            FileWriter fw = new FileWriter(file);
+            for (Task task : listOfTasks) {
+                if (task instanceof Deadline) {
```
weiyang13, line 53: Great use of `instanceof`!

src/main/java/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+                ui.printAddTask(taskList, newTodo);
+            } catch (IllegalArgumentException e) {
+                throw new DukeException("OOPS!!! The description of a todo cannot be empty.");
+            } catch (Exception e) {
+                throw new DukeException("OOPS!!! Your input format is wrong. Use: todo [task description]");
+            }
+        } else if (inputType.equals("list")) {
+            try {
+                if (!userInput.isBlank()) {
+                    throw new Exception();
```
weiyang13: Perhaps you could directly throw a `DukeException` instead of throwing an `Exception`, catching it, and throwing another `DukeException`? While functionally same, this adds to the verbosity of the code. The use of the superclass `Exception` may also hinder future debugging when new sources of exceptions may arise.

weiyang13 (commented):
Good effort in writing the code! However, your code may tend on the verbose side (especially `Parser`), which makes it more difficult to read.

Regarding `Parser`, you could make the code more readable by extracting out the execution of commands into different methods (you could also create a `Command` class as suggested in the extension, but that is not absolutely necessary). I understand you might feel that this is not optimal as it requires passing around a lot of parameters (e.g. `Ui`, `TaskList`) , but this can be overcome by creating an instantiation of a `Parser` object with these components stored as attributes.

Keep up the effort!

src/main/java/Ui.java
WilliamRyank: Hi thank you for your constructive comment @Davidcwh . I did not think of doing that and i think it will really clean up my codes.

src/main/java/Duke.java
WilliamRyank: Hi thanks for the constructive comments @weiyang13 . I didnt think of moving that part in the UI, but after your review, i think this approach will make my code looks cleaner. Thanks a lot

PR #228, opened 2019-08-28
[Chiang Jiajun Jared] Duke Increments

jcjjjared:


dalisc:
Hi! Your code is really clean and readable! You followed the coding standard, the commit message format and method naming well which makes the code look neat and easy to understand. I also like that you made Duke even more OOP by creating classes for separate commands and the inheritance you used makes sense. One improvement that can be made is adding documentation at class level as well. Another thing to look into when time allows is to package your classes! Aside from that, keep up the good work!

src/main/java/Duke.java
```diff
@@ -26,95 +22,20 @@ public static void main(String[] args) throws DukeException {
...
-                }
-            } catch (DukeException ex) {
-                System.out.println(ex.getMessage());
-            } catch (StringIndexOutOfBoundsException ex) {
-                System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
-            }
-            catch (DateTimeParseException ex) {
-                System.out.println("Invalid date and time format\n"
-                        + "Please enter date and time as 'dd/MM/yyyy HHmm'");
+                String fullCommand = ui.readCommand();
```
dalisc, line 38: Your main method is very short and readable now as a result of adding command classes. Very nice and well done!

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,33 @@
...
+            String line = task.toString();
+            if (line.contains(keyWord)) {
+                matchList.add(line);
+            }
+        }
+
+        System.out.println("Here are the matching tasks in your list:");
+
+        for (String line : matchList) {
+            System.out.println((index + 1) + "." + line);
```
dalisc: It wasn't explicitly specified in the instructions whether to assign new indexes to the matching tasks (numbering them from 1 again), but it does imply that in the instructions. 

As for me, I kept the original indexes of the task so that the user can more easily refer to their index if they want to issue a Done or Delete command. 

The instructions are ambiguous and this is something I will clarify with Prof and my tutor, so just a heads up to keep a lookout for clarifications regarding this!

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public boolean isExit() {
+        return false;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        Task chosen_Task = tasks.getTask(doIndex - 1);
```
dalisc: I noticed a number of variables where you used a combination of snake case and camel case. I think you should change it to strictly camel casing for this module.

Some of the variables I saw were chosen_Task, do_Index, deleteIndex. Perhaps it is better to change them to chosenTask, doTask, and deleteTask respectively. Apologies if I did not catch all of them!

crazoter:
As dalisc has mentioned the code is very readable and effective in what it needs to do. What I am about to suggest is merely a design decision (your current implementation has its strengths). I noticed that the isExit method in your Command classes are often overridden to a method that returns false. Perhaps you may want to consider turning the Command class into an abstract class and implementing a default implementation to return false so that you don't have to keep redefining the method.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,85 @@
...
+                throw new DukeException("Please enter date and time in format: dd/MM/yyyy HHmm");
+            }
+        }
+
+        try {
+            storage.addToFile(tasks.getTask(tasks.getNumOfTasks() - 1).toSaveString());
+            return "Got it. I've added this task:\n"
+                    + "  " + tasks.getTask(tasks.getNumOfTasks() - 1) + "\n"
+                    + "Now you have " + tasks.getNumOfTasks()
+                    + (tasks.getNumOfTasks() == 1 ? " task" : " tasks") + " in the list.\n";
```
tysg, line 64: The plural handling is good.

src/main/java/Command.java
```diff
@@ -0,0 +1,8 @@
+import java.io.IOException;
+
+public abstract class Command {
```
tysg, line 3: Maybe this can be an interface? Since there is no method definition.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,87 @@
...
+        super(description);
+        this.by = by;
+    }
+
+    /**
+     * Generates a date and time string appropriate for storing in save file.
+     * @return string representing date and time in the following format: 23rd September 2019, 3.01am
+     */
+    private String generateDateAndTimeString() {
+        String dateAndTime = "" + by.getDayOfMonth();
```
tysg, line 20: Unnecessary empty string.

src/main/java/EditCommand.java
```diff
@@ -0,0 +1,61 @@
...
+                tasks.setTask(editIndex - 1, deadline);
+            }
+        }
+
+        try {
+            storage.editTaskInFile(editIndex, tasks.getTask(editIndex - 1).toSaveString());
+            reply += tasks.getTask(editIndex - 1).toString() + "\n";
+            return reply;
+        } catch (IOException ex) {
+            return "Cannot save new task in file\n";
```
tysg, line 44: Maybe there should be a way to distinguish normal responses and exceptions.

src/main/java/ExitCommand.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Prints a closing statement before the programme closes.
+     * @param tasks TaskList containing the user's saved tasks
+     * @param ui Ui object to handle the user input
+     * @param storage storage object to determine where the executed results are stored
+     */
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
+        //ui.showClosing();
```
tysg: Remove comment in commit code.

src/main/java/Parser.java
```diff
@@ -0,0 +1,59 @@
...
+        assert !fullCommand.isEmpty() : "No command entered";
+
+        if (fullCommand.equals("bye")) {
+            return new ExitCommand();
+        } else if (fullCommand.equals("list")) {
+            return new ListCommand();
+        } else {
+            String[] fields = fullCommand.split(" ", 2);
+            String command = fields[0];
+            if (command.equals("done")) {
```
tysg, line 23: Consider using a `switch` case

crazoter:
I've run through your code again and as before, the code is relatively clean.
Looking forward to the user guide and application release.

PR #229, opened 2019-08-28
[Lee Wah] Duke Increments

bakwxh:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+        super(description);
+        this.by = by;
+    }
+
+    /**
+     * Shows task.
+     */
+    @Override
+    public String showTask() {
+        return "[D]" + super.showTask() + " (by: " + by.toString() + ")";
```
wh-coder-2025: A good way of using super.showTask()! 
I was wondering how to minimise the repetitions required in displaying the done status and the task name. 
This could be one way!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Shows task as its saving format.
+     */
```
wh-coder-2025: I think the JavaDocs information is incomplete. You might need to provide a description of the object being returned by your function if there are any.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+    	if (super.getDone()) {
+    		return "donedeadline " + super.getDesc() + " /by " + byString;
+    	} else {
+    		return "deadline " + super.getDesc() + " /by " + byString;
+    	}
+    }
+
+    /**
+     * Returns by time.
+     * @return By time.
```
wh-coder-2025: I actually misread this as "returns by/ time" as opposed to "returns/ by time". I thought it will be better if a more detailed explanation could be provided.

src/main/java/Duke.java
```diff
@@ -1,10 +1,98 @@
...
+	/**
+	 * Parser object.
+	 */
+	private Parser parser;
+
+    /**
+     * Constructor.
+     * @param filePath Path to save file.
+     */
+    public Duke(final String filePath) {
```
wh-coder-2025, line 24: I am not sure if the final keyword is needed over. The final keyword is only localised to the scope of the Duke constructor. If you are sure that there won't be any way down the function where you would accidentally modify the filePath, there is no need to finalise it.

src/main/java/Duke.java
```diff
@@ -1,10 +1,98 @@
...
+        parser = new Parser();
     }
-}
+
+    /**
+     * Main run method.
+     */
+    public void run() {
+    	try {
+			storage.load();
```
wh-coder-2025: Might need to double-check on the indentations.

src/main/java/Duke.java
```diff
@@ -1,10 +1,98 @@
...
+		} catch (DukeException e) {
+			ui.showException(e);
+		}
+    	ui.logoAndIntro();
+    	Scanner sc = new Scanner(System.in);
+		while (sc.hasNext()) {
+			String rawInput = sc.nextLine();
+			String command = parser.getCommand(rawInput);
+			try {
+				if (command.equals("bye")) {
```
wh-coder-2025: Would it be better if a switch-case statement is used?

src/main/java/DukeException.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * @author bakwxh
+ * @version 0.1
+ */
+public class DukeException extends Exception {
+	/**
+	 * Constructor.
+	 * @param message Exception message.
+	 */
+	public DukeException(final String message) {
```
wh-coder-2025: Might have the same issue as that in Duke's constructor

src/main/java/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * Path to save file.
+	 */
+	private String filePath;
+
+	/**
+	 * Constructor.
+	 * @param filePath Path to save file.
+	 */
+	public Storage(final String filePath) {
```
wh-coder-2025: Instead of using the final keyword in the parameter of the Storage constructor, would it be better to use the final keyword in the filePath instance field?

src/main/java/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+	 * @throws DukeException Exceptions.
+	 */
+	public ArrayList<Task> load() throws DukeException {
+		ArrayList<Task> taskList = new ArrayList<>();
+		try {
+			FileInputStream test = new FileInputStream(filePath);
+			Scanner sc = new Scanner(test);
+			while (sc.hasNextLine()) {
+				String rawInput = sc.nextLine();
+				if (rawInput.toLowerCase().startsWith("done")) {
```
wh-coder-2025: Felt rather verbose here. Are there ways to make this more succinct and concise?

src/main/java/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    public void markAsDone() {
+    	isDone = true;
+    }
+
+    /**
+     * Shows task as its saving format.
+     * @return Task in string format for saving purposes.
+     */
+    public String toSave() {
+		return null;
```
wh-coder-2025: Why does this return a null object?

src/main/java/Task.java
```diff
@@ -0,0 +1,71 @@
+/**
+ * @author bakwxh
+ * @version 0.1
+ */
+public class Task {
+	static String line = "____________________________________________________________";
```
wh-coder-2025: What is this line for?

src/main/java/Task.java
```diff
@@ -0,0 +1,71 @@
...
+     * Description.
+     * @return Description.
+     */
+    public String getDesc() {
+    	return this.description;
+    }
+
+    /**
+     * Is done.
+     * @return Is done.
```
wh-coder-2025: Descriptions could be worded better.

src/main/java/Todo.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * @author bakwxh
+ * @version 0.1
+ */
+public class Todo extends Task {
+	/**
+	 * Constructor.
+	 * @param description Description.
+	 */
+	public Todo(final String description) {
```
wh-coder-2025: Indentation

src/main/java/Ui.java
```diff
@@ -0,0 +1,129 @@
+/**
+ * @author bakwxh
+ * @version 0.1
+ */
+public class Ui {
+	static String line = "____________________________________________________________";
```
wh-coder-2025: Might be able to use the final keyword here to show that this is a constant.

wh-coder-2025 (commented):
General comments:
1) You might need to recheck the default indentations of your IDE. There are a lot of indentation errors.
2) Your JavaDoc comments could include more details.
3) Some parts of your code are too verbose. There could be ways to shorten them while improving the readability.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public Date getBy() {
```
fatclarence: I think that by convention, it is not necessary to include javadocs for your getters.

src/main/java/Duke.java.orig
```diff
@@ -0,0 +1,331 @@
...
+				System.out.println(line);
+				System.out.println();
+				break;
+			} else if (rawInput.toLowerCase().equals("list")) {
+				System.out.println(line);
+				System.out.println(" " + "Here are the tasks in your list:");
+				list();
+				System.out.println(line);
+				System.out.println();
+			} else if (rawInput.toLowerCase().startsWith("done")) {
```
fatclarence: I like the way you use .toLowerCase here to allow for the user to type instructions that may not necessarily be all lower-case which allows user's to interact with duke with less restrictions when it comes to typing out of instructions

src/main/java/Duke.java.orig
```diff
@@ -0,0 +1,331 @@
...
+				System.out.println(line);
+				System.out.println();
+			}
+		}
+		
+		sc.close();
+	}
+	
+	private static void addEvent(String rawInput) {
+		try {
```
fatclarence: You can consider splitting the various instructions into a Command abstract class and extending the Command class for the various events that can take place within duke, for example addEvents, deleteEvent etc., then execute the Commands in the run() method. It make the code more neat and improves readability.

PR #231, opened 2019-08-28
[Ng Zhao Ming] Duke Increments

ngzhaoming:


elevatethegame:
Hello Zhao Ming! I felt that the code you did to parse the date and time input by the user and handle the exceptions, namely the checkTime, validateDate and validateTime methods, is to be commended! 
It looks like you have made use of packages as well, (though optional), to further organise your code.

I do have one suggestion, that the logic of parsing the command in the main  
```java

try {
                parser.checkCommand(commandArr);
                switch (commandArr[0]) {
                case "list":
                    taskList.printArray();
                    break;

                case "done":
                    int indexDone = Integer.parseInt(commandArr[1]) - 1;
                    Task currTask = taskList.get(indexDone);
                    currTask.markAsDone();
                    taskList.taskComplete(currTask);
                    data.replaceNthLine(filename, indexDone, currTask);
                    break;
```
...
could be moved into the Parser class. 
Additionally, if you have the time, you could consider creating a parent Command class and children classes AddCommand, DeleteCommand, DoneCommand etc as suggested (optional) at one of the levels of the iP. 
Then, you could move the logic such as "currTask.markAsDone()" for the command "done" into an "execute" method in DoneCommand class. All of the Command subclasses could override the "execute" method in Command, using polymorphism to ensure the right method gets run. This would ensure a greater level of modularity!

Overall, well done!

ngzhaoming:
Hi @bruceskellator, thank you for the prompt and well-elaborated peer review. I really appreciate it.

> I do have one suggestion, that the logic of parsing the command in the main
> 
> ```java
> try {
>                 parser.checkCommand(commandArr);
>                 switch (commandArr[0]) {
>                 case "list":
>                     taskList.printArray();
>                     break;
> 
>                 case "done":
>                     int indexDone = Integer.parseInt(commandArr[1]) - 1;
>                     Task currTask = taskList.get(indexDone);
>                     currTask.markAsDone();
>                     taskList.taskComplete(currTask);
>                     data.replaceNthLine(filename, indexDone, currTask);
>                     break;
> ```
> 
> ...
> could be moved into the Parser class.
> Additionally, if you have the time, you could consider creating a parent Command class and children classes AddCommand, DeleteCommand, DoneCommand etc as suggested (optional) at one of the levels of the iP.
> Then, you could move the logic such as "currTask.markAsDone()" for the command "done" into an "execute" method in DoneCommand class. All of the Command subclasses could override the "execute" method in Command, using polymorphism to ensure the right method gets run. This would ensure a greater level of modularity!

I have looked into the suggestions that you have recommended (quoted above) and concurred that these do help to tidy up my code and make my main method more elegant. Indeed, the switch case portion of the code can be transferred to the Parser class instead. I will do these code housekeeping when I have the time.

Once again thank you for the feedback, and I'm currently working towards given my PR to yours too!

Xuerneas:
Hi @ngzhaoming, overall I felt that your code was logical and well organised. Also, the javadocs are really understandable.

Firstly, I felt that the use of LocalDateTime with DateTimeFormatter  is a very efficient way to get the date format we want.

Secondly, I noticed that your commands are organised by classes, by extending the Command class, instead of using an enum to handle all commands like me. I think your way is better since it is clearer and easier to use.

Lastly, there's a little suggestion. I noticed that you only used the getStatusIcon() in each toString() method, while toString() method was not really used in the project, so for the code below:
`public String fileString() {
        return " | " + (this.isDone ? "1" : "0") + " | " + this.description;
    }`
Is it possible to replace the `(this.isDone ? "1" : "0")` with this.getStatusIcon() so that the code is clearer? 

In conclusion, I thought that you really did well in this individual project. Looking forward to see your final product.

PR #232, opened 2019-08-28
[Jonathan Tjendana] Duke Increments

jonathantjendana:


liakify:
I like your use of enums to handle the various types of commands! Maybe you could name it as `CommandEnum` instead to be clearer. I also like how you used a shell script to further automate testing in addition to JUnit tests.

liakify (commented):
I like your use of enums to handle the various types of commands! Maybe you could name it as `CommandEnum` instead to be clearer. I also like how you used a shell script to further automate testing in addition to JUnit tests.

jietung:
I like that you extracted out the date time handler. Your codes looks nice and neat, which makes it easy to read.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,93 @@
...
+
+import task.Task;
+import task.TaskList;
+import task.Todo;
+import task.Deadline;
+import task.Event;
+
+import Exception.DukeException;
+
+public class Command {
```
jun-ha0, line 11: You may consider using an abstract class. :)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+        ArrayList<Task> taskList = new ArrayList<Task>();
+        try {
+            Scanner s = new Scanner(file);
+            while (s.hasNext()) {
+                Task task = decodeTask(s.nextLine());
+                taskList.add(task);
+            }
+        } catch (Exception ex) {
+            throw new DukeException("I can't load your tasks.");
+        }
```
jun-ha0, line 81: You may consider closing the scanner.

src/main/java/helper/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+public class Parser {
+    /**
+     * Parses user input into enums and separate the details accordingly.
+     *
+     * @param fullCommand user input to be parsed into duke enum.
+     * @return a duke.Command object.
+     * @throws DukeException when an occurred with a specific message.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        String[] strArr = fullCommand.split(" ", 2);
```
jun-ha0, line 16: You may consider storing strArr[0] in a String variable to improve readability.

jun-ha0 (commented):
Code is good overall. Do consider more refactoring in general. 😄

PR #233, opened 2019-08-28
[Bill Chee] Duke Increments

BillChee123:


Na-Nazhou (commented):
Personally I feel all the `*.class` files should not be committed. Similarly `list.txt` should be put inside .gitignore as well. 
Reference: https://stackoverflow.com/questions/8236785/should-class-files-be-put-under-version-control#targetText=3%20Answers&targetText=Yes%2C%20.class%20files%20produced%20from,omit%20from%20your%20source%20control%3F

src/main/java/command/Command.java
```diff
@@ -0,0 +1,33 @@
...
+     * @param ui duke Ui object responsible for interface response.
+     * @param storage duke Storage object to make changes to hard drive data.
+     */
+    public abstract void execute(TaskList tasklist, Ui ui, Storage storage) throws DukeException;
+
+    public boolean isExit() {
+        return this.isExit;
+    }
+
+    public void setExit(boolean exit) {
```
Na-Nazhou: I think this boolean parameter should be named as `isExit` according to the style guide.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static java.lang.String.format;
+
+/**
+ * Ui class for user interaction and to manage interface.
+ *
+ */
+public class Ui {
+    private String line;
+    Scanner sc;
```
Na-Nazhou: Is there any reason that `sc` is not a private attribute? Do you need to access it anywhere outside the `Ui` class?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,54 @@
...
+    public String readCommand() {
+        String s = sc.nextLine();
+        return s;
+    }
+
+    /**
+     * printing the given input to the interface.
+     * @param args iteration of string inputs that each represent a line to be printed.
+     */
+    public void printStatement(String... args) {
```
Na-Nazhou: Nice usage of varargs and stream! I did not think of this :D

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Task which has a deadline.
+ *
+ */
+public class Deadline extends Task {
+    private String by;
```
Na-Nazhou: I believe `by` can be removed from this class since it is not used any where. Correct me if I am wrong :)

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+    public String toString() {
+        return "[D]" + super.toString() + " (by: "
+                + date.format(DateTimeFormatter.ofPattern("d MMM yyyy"))
+                + ", " + time.format(DateTimeFormatter.ofPattern("h:mm a")) + ")";
+    }
+
+    @Override
+    public String fileFormat() {
+        return String.format("D | %s | %s | %s %s", isDoneString(), getDescription(),
+                date.format(DateTimeFormatter.ofPattern("dd/MM/yyyy")),
```
Na-Nazhou: I think `DateTimeFormatter.ofPattern("dd/MM/yyyy")` (and other DateTime formatters as well) can be abstracted out since this formatter is used in both `fileFormat()` and the constructor. Maybe you can have another class handling parsing/formatting DateTime and it would be also easier if you wish to parse/format DateTime in other classes (such as Event) in future.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public abstract String fileFormat();
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "[done]" : "[x]"); //return tick or X symbols
```
Na-Nazhou: This comment does not reflect the code correctly. Maybe you need to update it.

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,40 @@
...
+    private int index;
+
+    /**
+     * Marks a task as done.
+     *
+     * @param input Given user input.
+     * @throws DukeException If any error is encountered.
+     */
+    public DoneCommand(String input) throws DukeException {
+        this.index = Integer.parseInt(input);
```
Na-Nazhou: `NumberFormatException` should be caught here as well I think (similar to what you did in `DeleteCommand`)?

src/main/java/command/ListCommand.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Command to List the existing tasks.
+ */
+public class ListCommand extends Command {
+
+    /**
+     * Command to List the existing tasks.
+     */
+    public ListCommand() {
+        this.setExit(false);
```
Na-Nazhou: Instead if set `isExit` to be `false` in every `Command` subclass (except for `ByeCommand`), an alternative implementation is to  default `isExit` to be false in the parent `Command` class and then you only need to set it to be true in `ExitCommand` class.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+     * duke.Storage constructor.
+     *
+     * @param filepath path of data file.
+     */
+    public Storage(String filepath) {
+        this.filepath = filepath;
+    }
+
+    /**
+     * Initially populates the Task List with the hard drive data (Input data file).
```
Na-Nazhou: There should be a blank comment line between the description and the list of tags, according to the style guide :P

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
...
+    /**
+     * Boots the Duke program.
+     */
+    private void run() {
+        boolean isExit = false;
+        ui.printStatement("Hello! I'm Duke", "What can I do for you?");
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                // !sc.hasNext("bye")
```
Na-Nazhou: This comment can be removed I think?

src/main/java/duke/Ui.java
BillChee123: Thank you for your comments! It helped me improved my code quality.

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,94 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+import static java.lang.String.format;
+
+/**
+ * Command to add task into data.
+ *
+ */
+public class AddCommand extends Command {
```
Aulud: I like how the the commands handling addition of a `Todo`, `Deadline` or `Event` has been abstracted into a more generic `AddCommand`. Nice!

src/main/java/command/ByeCommand.java
```diff
@@ -0,0 +1,22 @@
...
+import duke.Ui;
+
+/**
+ * Command to end the Duke program.
+ *
+ */
+public class ByeCommand extends Command {
+
+    public ByeCommand() {
+        this.setExit(true);
```
Aulud: I think it would be cleaner if the `ByeCommand` class simply `@Overrides` the `isExit` method (which by default would return `false`) from `Command` to return `true`, instead of maintaining and using an additional method `setExit` to set the parent's attribute.

src/main/java/command/Command.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+    /**
+     * Executes the given line instruction passed into the object.
+     *
+     * @param tasklist Tasklist object containing the data structure.
+     * @param ui duke Ui object responsible for interface response.
+     * @param storage duke Storage object to make changes to hard drive data.
+     */
+    public abstract void execute(TaskList tasklist, Ui ui, Storage storage) throws DukeException;
```
Aulud: It may be worth considering changing the `execute` method of `Command` and its child classes such that executing the command returns a result of some type, which the main `Duke` class would pass to its `Ui` instance to handle. This would de-couple commands from the class that handles UI and allow it to be more flexible.

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Command to delete a task.
+     *
+     * @param input the task number in string.
+     * @throws DukeException If the task number is of invalid format.
+     */
+    public DeleteCommand(String input) throws DukeException {
+        try {
+            this.index = Integer.parseInt(input);
+            this.setExit(false);
```
Aulud: Calling `setExit(false)` in all command constructors is a lot of work - consider having the field set in the constructor of the abstract class (which will be invoked by the child classes) instead.

src/main/java/command/FindCommand.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public FindCommand(String expr) {
+        this.expr = expr;
+    }
+
+    @Override
+    public void execute(TaskList tasklist, Ui ui, Storage storage) throws DukeException {
+        Stream<String> taskStream = tasklist
+                .getList()
+                .stream()
```
Aulud: Nice stream solution! :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param fullCommand full command input by user.
+     * @return A Command which is able to be executed.
+     * @throws DukeException If command is invalid.
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        String[] arr = fullCommand.split(" ", 2);
+        String cmd = arr[0];
+        Command command;
+        switch (cmd) {
+        case "todo":
```
Aulud, line 30: According to the Java Coding Standard,
> The explicit `//Fallthrough` comment should be included whenever there is a case statement without a break statement.

This affects line 31 - 32.

src/main/java/command/ClearCommand.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Command to clear all item in the TaskList.
+ *
+ */
+public class ClearCommand extends Command {
+
+    @Override
+    public void execute(TaskList tasklist, Ui ui, Storage storage) throws DukeException {
+        tasklist.clearAll();
```
Aulud: Nice clear command!

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,23 @@
...
+import task.Deadline;
+import task.Task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class DeadlineTest {
+
+    @Test
+    public void deadlineTest() {
+        Task deadline = new Deadline("Deadline deadlinetest", "23/05/1997", "1800");
```
Aulud, line 11: Nice! Your tests are pretty thorough.

Aulud (commented):
Nice work overall!

src/main/java/command/ByeCommand.java
BillChee123: Yes I looked through your code and agree that it isn't necessary to really use a isExit variable since it is already unique to each command class hence just need to implement each isExit() method differently. Thank you for your feedback and idea!

src/main/java/duke/Parser.java
BillChee123: These cases actually have break statement. This is a syntax indicating that they will all execute the code after the event case.

PR #234, opened 2019-08-28
[Victor Varian] Duke Increments

victorvic54:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,29 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    private LocalDateTime by;
+    private DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+
+    /**
+     * A subclass of Task that categorized itself as an important thing to-do
```
thaowei: Perhaps something like "A task with a deadline" would be better? Your Javadoc should describe what this class represents. That it is a subclass is easily seen from the class declaration.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @param description description of the Deadline (Task).
+     * @param by the date of the deadline.
+     */
+    public Deadline(String description, LocalDateTime by) {
+        super(description);
+        this.by = by;
+    }
+
+    // get the date and time with specific format
```
thaowei: This isn't proper Javadoc format.

src/main/java/Duke.java
```diff
@@ -1,10 +1,48 @@
+import java.util.Scanner;
+
 public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Setting up several classes ready to be called.
```
thaowei: The first sentence for a constructor should be something like "Constructs a ...".

In this case, maybe "Constructs a Duke instance that saves its data at the specified path."

src/main/java/Duke.java
```diff
@@ -1,10 +1,48 @@
...
+
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (Exception e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    // run the program
```
thaowei: This isn't a valid Javadoc comment.

(But I would say this doesn't really need to be documented.)

src/main/java/DukeException.java
```diff
@@ -0,0 +1,14 @@
+public class DukeException extends Exception {
+    /**
+     * To handle several logic error from the program that is thrown to this exception.
```
thaowei, line 3: This sounds like it's describing the class, not this constructor.

src/main/java/Event.java
```diff
@@ -0,0 +1,29 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task {
+    private LocalDateTime at;
+    private DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+
+    /**
+     * A subclass of Task that categorized itself as an occasion that will
+     * be held at a particular time.
```
thaowei, line 10: This sounds like it's describing the class.

src/main/java/Event.java
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @param description description of the Event (Task).
+     * @param at the date of the event.
+     */
+    public Event(String description, LocalDateTime at) {
+        super(description);
+        this.at = at;
+    }
+
+    // get the date and time with specific format
```
thaowei: Not a valid Javadoc comment.

src/main/java/Parser.java
```diff
@@ -0,0 +1,110 @@
+import java.time.LocalDateTime;
+
+public class Parser {
+    /**
+     * Convert a String into a LocalDateTime that is understood by the computer.
```
thaowei: "Converts..."

But maybe a better verb would be _parse_?

src/main/java/Parser.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param rawTimestamp time and date in string format.
+     * @return LocalDateTime with desired format.
+     * @throws DukeException if there is logic error or bad input.
+     */
+    public static LocalDateTime convertDateAndTime(String rawTimestamp) throws DukeException {
```
thaowei, line 12: Why not use [`DateTimeFormatter#parse`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html#parse(java.lang.CharSequence))?

src/main/java/Parser.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * To deals with making sense of the user command.
+     *
+     * @param tasks the TaskList from the txt file.
+     * @param ui the message for interaction.
+     * @param inputText the text that user input.
+     * @param storage the location for retrieval and write the updated file.
+     */
+    public static void parse(TaskList tasks, Ui ui, String inputText, Storage storage) {
```
thaowei: It'd be good if `parse` merely parsed the input text, and the actual execution of the command were done elsewhere. (Single responsibility principle)

src/main/java/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+     * Load the list of task from collection of words/Strings.
+     *
+     * @return LinkedList of Task.
+     * @throws FileNotFoundException if no file detected.
+     * @throws DukeException if there is logic error or bad input.
+     * @throws IOException file corrupted.
+     */
+    public LinkedList<Task> load() {
+        File file = new File(filePath);
+        LinkedList<Task> loadedTasks = new LinkedList<>();
```
thaowei, line 31: I'm curious why you chose a `LinkedList` instead of an `ArrayList`.

src/main/java/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+        File file = new File(filePath);
+        boolean isFileExists = file.exists();
+
+        try {
+            if (!isFileExists) {
+                file.createNewFile();
+            }
+
+            FileWriter fw = new FileWriter(filePath);
+            for (Task subTask: updatedTask) {
```
thaowei, line 121: We usually have one space on each side of the colon.

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    /**
+     * A class that handle the creation, deletion, and markAsDone of a task.
```
thaowei, line 6: This is more appropriate as a class Javadoc.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,33 @@
+import java.util.LinkedList;
+
+public class TaskList {
```
thaowei, line 3: I feel like this is a pretty useless abstraction... but I suppose it was asked for.

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+    @Test
+    public void testGetDateTime() {
+        String description = "exampleDescription";
+        LocalDateTime localDateTime = LocalDateTime.of(2019, 8, 12, 12, 50);
+
+        assertEquals(2, new Deadline(description, localDateTime).getBy().split(" ").length);
```
thaowei, line 12: Why not just assert equality to `localDateTime`?

src/test/java/EventTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class EventTest {
+    @Test
+    public void testGetDateTime() {
+        String description = "exampleDescription";
+        LocalDateTime localDateTime = LocalDateTime.of(2019, 8, 12, 12, 50);
+
+        assertEquals(2, new Event(description, localDateTime).getAt().split(" ").length);
```
thaowei, line 13: Same as the test for Deadline.

thaowei (commented):
I think your code is pretty clean! Most of the comments I have are about the phrasing of your Javadocs. Nice :)

src/main/java/Event.java
```diff
@@ -0,0 +1,29 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task {
```
ChangUo79, line 4: Felt that most of your description for the constructors should have been for the class.

src/main/java/Storage.java
victorvic54: I am actually curious about this. The reason why I choose LinkedList rather than ArrayList is that LinkedList can do everything that ArrayList has and I think much faster. Why I think ArrayList is better when converting ArrayList to array is faster? What do you think?

src/main/java/TaskList.java
victorvic54: Yeah lol no choice.. quite redundant I feel

PR #236, opened 2019-08-28
[Leow Wen Bin] Duke Increments

LeowWB:


src/main/java/Duke.java
```diff
@@ -1,10 +1,176 @@
...
+    }
+
+    // Processes a single line of input by identifying the command that was given, then delegating it
+    // to a subfunction to handle the command call.
+    private void processInputLine(String line) throws DukeException, ParseException {
+
+        Command command = parser.getCommandFromLine(line);
+        String commandText = command.toString();
+
+        switch (command) {
```
jon-chua: I like this switch statement, very clean 👍

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Given a line, obtains the (numerical) index present in the line.
+     * 
+     * @param line The line input by the user
+     * @return The int index present in the line
+     */
+    public int getIndexFromLine(String line) {
+        try {
+            return Integer.parseInt(line.split(" ")[1]) - 1;
+        } catch (Exception e) {
```
jon-chua, line 36: I think it would be much better to catch specific exceptions instead of catching all of them!

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+     * 
+     * @param line Line input by user
+     * @param command Command text within line
+     * @param delim Desired delimiter
+     * @return String representation of the part of the line between the command and delimiter
+     */
+    public String getBeforeDelim(String line, String command, String delim) {
+        try {
+            return line.split(command)[1].split(delim)[0].trim();
+        } catch (Exception e) {
```
jon-chua, line 61: Similar to ```getIndexFromLine``` above

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+     * 
+     * @param line Line input by user
+     * @param command Command text within line
+     * @param delim Desired delimiter
+     * @return String representation of the part of the line after the delimiter
+     */
+    public String getAfterDelim(String line, String command, String delim) {
+        try {
+            return line.split(command)[1].split(delim)[1].trim();
+        } catch (Exception e) {
```
jon-chua, line 79: Similar to ```getIndexFromLine``` above

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+     * Gets the argument from the line input by user, i.e the part of the line excluding the command
+     * 
+     * @param line Line input by user
+     * @param command String representing the command within the line
+     * @return The argument within the line
+     */
+    public String getArg(String line, String command) {
+        try {
+            return line.split(command)[1].trim();
+        } catch (Exception e) {
```
jon-chua, line 117: Similar to ```getIndexFromLine``` above

src/main/java/TaskList.java
```diff
@@ -0,0 +1,44 @@
...
+
+            if (current
+                .getDescription()
+                .toLowerCase()
+                .contains(
+                    searchStr.toLowerCase()
+                )
+            ) {
+                newList.add(current);
+            } else {
```
jon-chua, line 37: I think this else block is redundant and can be removed

jon-chua (commented):
Your code is really clean and easy to read 😄 
Furthermore, it was very well documented.
Implementing a few more tests would reduce the risk of introducing regressions in your code!

Overall, well done 👍

src/main/java/Ui.java
```diff
@@ -7,7 +7,7 @@
     // Strings that Duke will output
     private static final String PRINTED_LONG_LINE = "____________________________________________________________";
```
caesarpjz, line 8: Oh wow, this is an interesting way to print out your messages. Thanks for that! :)

src/main/java/Duke.java
```diff
@@ -26,125 +26,144 @@
...
-            );
-            break;
-        case FIND:
-            findTask(
-                parser.getArg(line, commandText)
-            );
-            break;
-        default:
-            throw new InvalidCommandException(PRINTED_OOPS + PRINTED_INVALID_COMMAND);
+    public String processInputLine(String line) {
```
caesarpjz, line 68: Very neat code!

PR #238, opened 2019-08-28
[Siang Hwee] Duke Increments

sianghwee:


src/main/java/CorruptedData.java
```diff
@@ -0,0 +1,13 @@
+package duke.exception;
+
+public class CorruptedData extends DukeException {
```
qiujiaaa, line 6: Perhaps, you can consider naming the class as "CorruptedDataException" instead because it makes it clearer that it is an exception class.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+     * @throws MissingPreposition if time does not contain a preposition by
+     */
+    public Deadline(String description, String deadline) throws MissingPreposition {
+        super(description);
+        String[] temp = deadline.split(" ", 2);
+        try {
+            deadline = temp[0] + ": " + temp[1];
+            this.deadline = deadline;
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new MissingPreposition("deadline");
```
qiujiaaa, line 28: Maybe the error message can be a bit more specific? Instead of deadline, you can mention the missing preposition in the message to make it clearer.

src/main/java/Event.java
```diff
@@ -0,0 +1,48 @@
...
+     * @throws MissingPreposition if time does not contain a preposition at
+     */
+    public Event(String description, String time) throws MissingPreposition {
+        super(description);
+        String[] temp = time.split(" ", 2);
+        try {
+            time = temp[0] + ": " + temp[1];
+            this.time = time;
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new MissingPreposition("event");
```
qiujiaaa, line 27: Similar to my previous comment, the code might be clearer if the message at this point in time shows a proper sentence of the error

src/main/java/Event.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    /**
+     * Converts the event object to a text format. Allows the object to be saved
+     * more easily in a text file.
+     * 
+     * @return a string representation of the task
+     */
+    public String toText() {
```
qiujiaaa, line 37: It might be a bit confusing because toText() and toString() sounds like they are doing the same thing, maybe it is better to emphasize that toText() is for storage purposes?

src/main/java/Parser.java
```diff
@@ -0,0 +1,91 @@
...
+        String[] desc;
+        try {
+            switch (split[0]) {
+            case "list":
+                return new ListCommand();
+            case "delete":
+                try {
+                    return new DeleteCommand(Integer.parseInt(split[1]));
+                } catch (ArrayIndexOutOfBoundsException e) {
+                    throw new InsufficientArgument();
```
qiujiaaa: It would be good to include an error message to make the code clearer!

src/main/java/Parser.java
```diff
@@ -0,0 +1,91 @@
...
+                }
+            case "done":
+                try {
+                    return new DoneCommand(Integer.parseInt(split[1]));
+                } catch (ArrayIndexOutOfBoundsException e) {
```
qiujiaaa: This 2 try catch block in "done" and "todo" are the same, it might be a bit repetitive. Maybe you can wrap the entire switch block in a try catch block instead?

src/main/java/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+        if (!dbDir.exists()) {
+            dbDir.mkdir();
+        }
+
+        // Creates a new text file within the database directory if it does not exist
+        File dbFile = new File(dbDir, filename);
+        try {
+            dbFile.createNewFile();
+        } catch (IOException e) {
+            System.out.println(e);
```
qiujiaaa, line 51: Maybe it would be good for you to write your own error message so that it can be more similar to the other error messages you have!

src/main/java/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Writes the data from a tasklist to the file.
+     * 
+     * @param db the tasklist containing all the existing tasks
+     */
+    public void write(TaskList db) {
+        try {
+            BufferedWriter writeData = new BufferedWriter(new FileWriter(dbFile));
+            for (Task t : db.getList()) {
```
qiujiaaa, line 97: If i am not wrong, there should not be another space before the ":"?

qiujiaaa (commented):
Hi! I like how your code is actually really neat and it was relatively easy to understand your code! Have added some suggestions below at some parts which i feel that it can be made clearer. Really nice job on the javadoc comments :)

PR #239, opened 2019-08-28
[Iskandar Zulkarnaien] Duke Increments

iskandarzulkarnaien:


YiJiee:
Review on: 4 Sep 2019

Code was well organised into two packages: exception and task.
Code follows coding standard and JavaDocs were written clearly.
Duke and Storage class could be packaged.
As of current, although Ui and TaskList class were not implemented, app compiles and works great until Level-9.

src/main/java/duke/Duke.java
```diff
@@ -45,7 +45,7 @@ private static void respondToInput() {
         while (!userInput.equals("bye")) {
             try {
                 if (userInput.equals("list")) {
-                    displayList();
+                    displayList(tasks);
                 } else if (checkIsInputEquals(userInput, "done ")) {
                     Task task = tasks
                             .get(Integer.parseInt(userInput.substring("done ".length())) - 1);
```
weikiat97: Perhaps you can consider creating a variable to explain what Integer.parseInt does in this case, so that it is clearer for the users.

src/main/java/duke/Duke.java
```diff
@@ -45,7 +45,7 @@ private static void respondToInput() {
         while (!userInput.equals("bye")) {
             try {
                 if (userInput.equals("list")) {
-                    displayList();
+                    displayList(tasks);
                 } else if (checkIsInputEquals(userInput, "done ")) {
```
weikiat97: Can just use userInput.substring(0, 4).contains("done") or something like that, because the current method is very long and it's a bit confusing. Could also consider changing the method to "checkIfInputEquals"

src/main/java/duke/Duke.java
```diff
@@ -60,6 +60,15 @@ private static void respondToInput() {
                 } else if (checkIsInputEquals(userInput, "delete ")) {
                     int taskIndex = Integer.parseInt(userInput.substring("delete ".length())) - 1;
                     deleteAndDisplayTask(taskIndex);
+                } else if (checkIsInputEquals(userInput, "find ")) {
+                    String keyword = userInput.substring("find ".length());
+                    ArrayList<Task> matchingTasks = new ArrayList<>(100);
```
weikiat97: No need to put 100.

src/main/java/duke/Duke.java
```diff
@@ -60,6 +60,15 @@ private static void respondToInput() {
...
+                    for (Task task : tasks) {
+                        if (task.getInfo().contains(keyword)) {
+                            matchingTasks.add(task);
+                        }
+                    }
+                    displayList(matchingTasks);
                 } else {
                     dukeReply("I don't know what that means, sorry!");
```
weikiat97: Specify where your error in this case is (invalid command at the start). Same for the rest of the exceptions in your code, so that it is clearer for the user to see what is wrong.

weikiat97 (commented):
Great effort! Well done. Perhaps you could consider making your code more OOP by having a Parser class to manage all the commands that is inputed.

Alvinnyk (commented):
Code is well written clearly and easy to understand
Maybe in the function checkIsInputEquals, u can use string.split instead
There is no need to specify size of an ArrayList in Storage.java

PR #240, opened 2019-08-28
[Lee Wei Gen] Duke Increments

weigenie:


src/main/java/duke/execution/Command.java
```diff
@@ -0,0 +1,89 @@
...
+        this.action = action;
+        this.txt = txt;
+    }
+
+    public boolean execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        switch (action) {
+        case "bye":
+            ui.exit();
+            return false;
+
```
chowyiwai: I don't think there should be empty lines in between the different cases.

src/main/java/duke/execution/Command.java
```diff
@@ -0,0 +1,89 @@
...
+    public Command(String action, String txt) {
+        this.action = action;
+        this.txt = txt;
+    }
+
+    public boolean execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        switch (action) {
+        case "bye":
+            ui.exit();
+            return false;
```
chowyiwai: Should comment // Fallthrough if you do not break for a case

src/main/java/duke/execution/Parser.java
```diff
@@ -0,0 +1,62 @@
...
+            } else if (int_day % 10 == 3) {
+                ordinalIndicator = "rd";
+            } else {
+                ordinalIndicator = "th";
+            }
+
+            String outputDate = int_day + ordinalIndicator + " of " + month + " " + year + ", " + time;
+            return outputDate;
+        } catch (ParseException e) {
+            System.out.println("isDate has bugs");
```
chowyiwai, line 74: I think it would be better to make all interactions with the user, i.e. prints to the console, go through the user interface class

src/main/java/duke/execution/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.FileWriter;
+import java.io.BufferedWriter;
+
+import java.util.ArrayList;
+
+public class Storage {
+
+    String filepath;
```
chowyiwai, line 16: Should probably include an access modifier here

src/main/java/duke/execution/Command.java
```diff
@@ -0,0 +1,89 @@
+package duke.execution;
+
+import duke.exceptions.DukeException;
+
+import duke.models.Task;
+
+public class Command {
```
chowyiwai: Might be good to split commands into different classes so that it is neater but this way works too! 👍

chowyiwai (commented):
Hi Wei Gen,

Good job on your work so far! 👍 I think that your code is neat and easy to read. I also like how your variable names are clear and this makes it easy for the reader to understand the code. Your Javadoc comments are also succinct and clear. One thing you could work on would be to try to isolate all the user interactions into one class (the Ui class). There were also some coding style issues that you might want to check again. 

But good job overall!

t-cheepeng:
Hi Wei Gen,

Super job on the iP so far. The code done was very clear and packaged into the respective packages that clearly reflect the distinctive purpose of the classes inside the package. Also, the documentation done was useful in helping to show briefly what each method does and the respective parameters and return values. That being said, there are certain portions of code that could be refactored out into their own classes to make it even neater and more friendly to read. But overall, it was a job well done!

src/main/java/duke/execution/Command.java
t-cheepeng: I agree with yi wai that the commands could have been splitted out their more specific types of command like AddCommand, FindCommand. Certainly would help in readability

src/main/java/duke/execution/Command.java
```diff
@@ -0,0 +1,89 @@
...
+            ui.listTasks(tasks.find(txt));
+            break;
+
+        default:
+            throw new DukeException(" ☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
+        return true;
+    }
+
+}
```
t-cheepeng: Maybe you could consider having your Command class hold only command related information and methods. The parsing and throwing of exception could be done by the Parser and the relevant exceptions be thrown when parsing

src/main/java/duke/execution/Parser.java
```diff
@@ -0,0 +1,62 @@
...
+            System.out.println(e.getMessage());
```
t-cheepeng, line 75: There are multiple println messages here, one with the specific error message and one of the exception message. Maybe you could consider displaying relevant information to the user

src/main/java/duke/execution/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+                        break;
+                    default:
+                        throw new DukeException("error data formatting in data.txt");
+                }
+            }
+            File file = new File(filepath);
+            bufferedReader.close();
+            return tasks;
+        } catch (IOException e) {
+            System.out.println(e.getMessage());
```
t-cheepeng: For exceptions, i think its generally a good idea to format it nicely for user or if its for developer use, you could the method
`e.printStackTrace()`. That will print the stack of the program when the program meets the exception

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,183 @@
...
+            ui.showLoadingError(e);
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Initialising of programme.
+     * @param filePath Filepath of data stored.
+     */
+    public Duke(String filePath) {
```
Kzrthikz, line 67: The logic of both constructors are similar. As a suggestion, you can consider using just one constructor to process the data.

src/main/java/duke/execution/Parser.java
```diff
@@ -0,0 +1,150 @@
...
+        String rest = getRestOfInput(split);
+        return parseForCommand(split[0], rest);
+    }
+
+    /**
+     * Checks whether a string can be formatted to be a Date.
+     * @param txt String to be checked.
+     * @return Boolean whether the string can be formatted to be a Date.
+     */
+    public static boolean isDate(String txt) {
```
Kzrthikz, line 33: We suggest that you isolate this method into an utility class to reduce dependencies between classes.

src/main/java/duke/execution/Parser.java
```diff
@@ -0,0 +1,150 @@
...
+            }
+            return new FindCommand(txt);
+
+        default:
+            System.out.println("ERROR_LOG: txt: " + txt);
+            throw new DukeException(" ☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
+    }
+
+    private static String getRestOfInput(String[] split) {
```
Kzrthikz, line 127: There is already an existing similar overloaded method eg. string.split("-", integer) that lets you limit the number of splits. 
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#split(java.lang.String,int)

src/main/java/duke/execution/Storage.java
```diff
@@ -0,0 +1,115 @@
+package duke.execution;
+
+import duke.exceptions.DukeException;
+
+import duke.models.Deadline;
+import duke.models.Event;
+import duke.models.Task;
+import duke.models.Todo;
+
+import java.io.*;
```
Kzrthikz, line 10: may refer to java coding standard, import classes should be specified

Kzrthikz (commented):
Good job so far, here are some minor comments for you to improve :)

PR #241, opened 2019-08-28
[Anisha] Duke Increments

nishcafe:


src/main/java/tool/DateTime.java
```diff
@@ -0,0 +1,38 @@
+package tool;
+
+public class DateTime {
```
iltep64, line 3: Java has a standard `LocalDateTime` class, in `java.time`. It might be easier to work with the existing libraries instead of writing your own.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,42 @@
...
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
+    }
+
+    /**
+     * Changes the status of a task from not done to done
+     */
+    public void markAsDone() {
+        this.isDone = true;
+        this.status = 1;
```
iltep64, line 34: Is `this.status` redundant here, with `this.isDone` already present?

src/main/java/tool/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+        this.ui = ui;
+    }
+
+    /**
+     * Parses the user's input to make sense of all the commands, and hands it over
+     * to tool.TaskList
+     * @param command
+     */
+    public void parse(String command) {
+            String[] inputArr = command.split(" ");
```
iltep64, line 28: This method contains a large bulk of Duke's logic. Would it be better to split each command off into a separate method? For example:
```
if(userCommand.equals("done")) {
    try {
         int index = Integer.parseInt(inputArr[1]) - 1;
         try {
             Task doneTask = this.commands.done(index);
    /* ... */
} else //...
```

becomes

```
if(userCommand.equals("done")) {
    int index = Integer.parseInt(inputArr[1]) - 1;
    this.markTaskAsDone(index);
} else
```

src/main/java/tool/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.ArrayList;
+
+
+public class Storage {
+    protected String filePath;
+    protected FileWriter fw;
```
iltep64: `this.fw` isn't used anymore, as you create a new `FileWriter` instead.

src/main/java/tool/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+            while ((x = br.readLine()) != null) {
+                if (i != j) {
+                    ff.write(x + System.lineSeparator());
+                    ff.flush();
+                }
+                j++;
+            }
+            ff.close();
+            Files.copy(Paths.get("src/main/java/data/temp.txt"), Paths.get(filePath), StandardCopyOption.REPLACE_EXISTING);
+            Files.delete(Paths.get("src/main/java/data/temp.txt"));
```
iltep64: Since all the storage-modifying logic does the same process of writing and re-saving the file, I think it would be better to pull the common logic into one single method.

src/main/java/task/Todo.java
```diff
@@ -0,0 +1,22 @@
...
+     * @return String to save in the .txt file
+     */
+    @Override
+    public String storageString() {
+        return "T/" + status + "/" + description;
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
```
iltep64, line 24: Nice reuse of the existing `Task.toString()`.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Formats the string to how it should be saved in the .txt file
+     * @return String to save in the .txt file
+     */
+    public String storageString() {
+        return "task.Task|" + status + "|" + description;
```
kelvinnharris, line 59: I think `status` can be replaced with `isDone ? "1" : "0"`, then you wouldn't have to store `status` field.

src/main/java/tool/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Loads tasks from the .txt file
+     * @param commands
+     * @return the updated array list of tasks
+     */
+    public ArrayList<Task> load(ArrayList<Task> commands) {
+        try {
+            FileReader fr = new FileReader(this.filePath);
+            BufferedReader br = new BufferedReader(fr);
```
kelvinnharris, line 43: `FileReader` and `BufferedReader` can be stored in `Storage` as a field since they are used in most methods.

src/main/java/tool/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+    }
+
+    /**
+     * Loads tasks from the .txt file
+     * @param commands
+     * @return the updated array list of tasks
+     */
+    public ArrayList<Task> load(ArrayList<Task> commands) {
```
kelvinnharris: `commands` can be renamed to avoid confusing names.

src/main/java/tool/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+                        try {
+                            String[] taskDeadLine = dL.split(" /by ");
+                            String taskD = taskDeadLine[0];
+                            String by = taskDeadLine[1];
+                            Task tt = new Deadline(taskD, new DateTime(by));
+                            this.commands.add(tt);
+                            this.ui.addTask(tt, this.commands.size());
+                            storage.save(tt);
+                        } catch (ArrayIndexOutOfBoundsException e) {
+                            throw new DukeException("OOPS!!! The format for deadline is wrong. Please follow: <description> /by <time>");
```
kelvinnharris: It is nice that you give suggestions when an invalid command is entered.

PR #243, opened 2019-08-28
[Chua Sim Nee] Duke Increments

SimNee:


src/main/java/Command.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param storage Storing the text files and updating it
+     * @throws DukeException Exception cases
+     * @throws IOException Unexpected exception cases
+     */
+    public void execute(TaskList taskList, UI ui,Storage storage) throws DukeException, IOException {
+        storage.LoadFile();
+        ui.printGreeting();
+        String TaskLine = ui.readCommand();
+        Parser parser = new Parser();
+        outLoop:
```
halfwhole, line 16: This `execute` method here seems to be the method that drives the entire program -- it contains the while loop that reads the input command and then does the appropriate things with `ui` and `taskList`, depending on what input it is. Might it be better to place this driver loop in the `Duke` class instead?

Also considering that there are really different types of commands (bye, list, done, todo, etc.), perhaps it could be helpful to create subclasses (e.g. `ListCommand`) that inherit from the `Command` class, and implement their own `execute` method (that does the appropriate thing with `ui` and `taskList` depending on what command it is).

halfwhole (commented):
There seems to be a lot of javadoc and other miscellaneous files here! I'm not 100% sure of the requirements, but I don't think they need to be included in the PR?

src/test/TaskListTest.java
```diff
@@ -0,0 +1,56 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
ernestyyh, line 3: Perhaps you could consider explicitly listing out the imported methods from the Assertions class. This would make it clearer.

src/test/TaskListTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+class TaskListTest {
+    TaskList taskListTest = new TaskList();
+    @Test
+    void add() {
```
ernestyyh, line 8: Perhaps the test methods could be more descriptive? For example taskList_add()?

src/main/java/UI.java
```diff
@@ -0,0 +1,128 @@
...
+        this.sc = sc;
+    }
+
+    public String readCommand() {
+        sc = new Scanner(System.in);
+        return sc.nextLine();
+    }
+
+    public void printGreeting() {
+        String logo = "   _     _      _     _      _     _       _     _      _     _      _     _      _     _   \n" +
```
ernestyyh: Wow. That's an interesting and very detailed logo for your chatbot.

src/main/java/UI.java
```diff
@@ -0,0 +1,128 @@
...
+        int count = 1;
+        for (int i = 0; i < findList.size(); i++) {
+            Task t = findList.get(i);
+            System.out.println(count + "." + t.toString());
+            count++;
+        }
+        System.out.println("____________________________________________________________\n");
+    }
+
+    public void throwInputError(String taskType) throws DukeException {
```
ernestyyh: Perhaps you could store pretty printing elements (e.g. the line segmentation) as a constant in the class?

PR #244, opened 2019-08-28
[Mo Zongran] Duke Increments

moziliar:


src/main/java/duke/command/EventCommand.java
```diff
@@ -0,0 +1,54 @@
...
+    private String eventDate;
+
+    /**
+     * Constructs an event command based on the input string,
+     * and store it as task description and event date.
+     *
+     * @param input Input string.
+     */
+    public EventCommand(String input) {
+        String[] processedInput = input.split("/by");
```
blimyj: Perhaps you might want to consider using a try block here to verify if the processedInput array is indeed length 2.
In cases where "/by" is not within the string, an error might occur.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,116 @@
...
+     * Prints formatted report on the task removed.
+     *
+     * @param task Task removed.
+     * @param numOfTasks Number of tasks left in the task list.
+     */
+    public void reportRemove(Task task, int numOfTasks) {
+        printFormattedText("    Noted. I've removed this task:\n"
+                + "       " + task
+                + "\n    Now you have " + numOfTasks
+                + " task" + (numOfTasks > 1 ? "s" : "") + " in the list.");
```
blimyj: Wish I had thought of this solution of deciding whether to put an "s"!

blimyj (commented):
Nice work! I like the segregation of commands into individual classes! I think I should learn from that as well. I think it might be possible for a little fine-tuning on the exception handling for incorrect inputs and perhaps a bit more testing such as ensuring the toString methods for the Task subclasses work correctly.

src/main/java/duke/gui/MainWindow.java
```diff
@@ -0,0 +1,53 @@
...
+    @FXML
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    private Button sendButton;
+
+    private Duke duke;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/Jotaro.jpg"));
```
ryantay232, line 30: 👀

src/main/java/duke/todo/TaskList.java
```diff
@@ -0,0 +1,225 @@
...
+
+import duke.DukeException;
+
+import java.lang.StringBuilder;
+import java.util.List;
+import java.util.ArrayList;
+
+public class TaskList {
+    private List<Task> taskList;
+    private int counter;
```
ryantay232: Could just use .size() method instead of having a counter?

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,106 @@
+package duke.ui;
+
+import duke.DukeException;
+import duke.todo.Task;
+
+public class Ui {
+    StringBuilder messageBuffer;
```
ryantay232, line 7: Never thought of using this to arrange the string fully before printing or returning it. Thanks!

ryantay232 (commented):
Very clean and well structured code! Seems there isn't much error handling for wrong user inputs. Still learned a lot from looking at your code!

src/main/java/duke/todo/TaskList.java
moziliar: Good suggestion:) I did it in a rush and couldn't recall .size().

src/main/java/duke/ui/Ui.java
moziliar: It could be simple concatenation because of the lightweight nature of this project. However, I do feel like doing StringBuilder.

src/main/java/duke/gui/MainWindow.java
moziliar: Thanks:)

moziliar: Will be adding user input validation in next commit:)

PR #245, opened 2019-08-28
[Poh Lin Wei] Duke Increments

pohlinwei:


src/main/java/AddCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Executes this command by adding <code>this</code> task to <code>taskList</code>.
+     * Informs user that command has been executed.
+     *
+     * @param taskList task list which <code>this</code> task should be added to
+     * @param storage storage which stores all tasks on the local hard disk, if any
+     */
+    public void execute(TaskList taskList, Optional<Storage> storage) {
```
shawnkoh, line 25: It would be better if you indicated that this method ```@Override``` the super class's execute

src/main/java/Command.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public interface Command {
+    /**
+     * Executes the intended command.
+     *
+     * @param taskList task list which <code>this</code> task should be added to
+     * @param storage storage which stores all tasks on the local hard disk, if any
+     * @throws Exception if command cannot be executed
+     */
+    public void execute(TaskList taskList, Optional<Storage> storage) throws Exception;
```
shawnkoh, line 14: It would be better if you threw an explicit Exception instead of the catchall.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @param o <code>Object</code> to be compared with <code>this</code>
+     * @return boolean indicating whether the two objects are 'equal'
+     */
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof Deadline) {
+            return this.taskName == ((Deadline) o).taskName
+                    && this.details == ((Deadline) o).details;
+        } else {
```
shawnkoh, line 70: there is no need for an else { } statement. The return method in the guard clause already exits the function.

src/main/java/Duke.java
```diff
@@ -1,10 +1,44 @@
...
+        try {
+            desiredStorage = Optional.of(new Storage(entries));
+        } catch (Exception e) {
+            Ui.showError(e);
+        }
+        Optional<Storage> storage = desiredStorage;
+        storage.ifPresent(s -> taskList.addPreviousTasks(s.load()));
+
+        // greet user
+        Ui.sayHi();
```
shawnkoh, line 20: comment can be made redundant if you renamed sayHi to greetUser.

src/main/java/EmptyDescriptionException.java
```diff
@@ -0,0 +1,21 @@
+
+class EmptyDescriptionException extends Exception {
+    private String type;
+
+    public EmptyDescriptionException(String task) {
+        super();
```
shawnkoh, line 6: Redundant super() call.

src/main/java/Event.java
```diff
@@ -0,0 +1,66 @@
...
+public class Event extends Task {
+    private static SimpleDateFormat dateBeautifier = new SimpleDateFormat("dd/MM/yy h:mma");
+
+    private String details;
+    private Calendar startTime;
+    private Calendar endTime;
+
+    /**
+     * Returns an event with a task name, start time and end time.
+     * @param taskName task of name
```
shawnkoh, line 16: coding style: line break after header comment.

src/main/java/Event.java
```diff
@@ -0,0 +1,66 @@
...
+     * when given a string representation of it.
+     *
+     * @param name name of task
+     * @return <code>Task</code> representing the given input
+     */
+    static Task stringToTask(String name, String timeInfo) {
+        String[] dateTimeArr = timeInfo.split(" ");
+        String date = dateTimeArr[0];
+        String[] times = dateTimeArr[1].split("-");
+        String time1 = times[0];
```
shawnkoh, line 48: potential IndexOutOfBounds exception

src/main/java/InvalidCommandException.java
```diff
@@ -0,0 +1,13 @@
+/**
+ * This indicates that an error was encountered when trying to parse user's command.
+ */
+class InvalidCommandException extends UnsupportedOperationException {
+    public InvalidCommandException() {
+        super();
```
shawnkoh, line 6: redundant constructor.

src/main/java/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param task task to be added to file
+     */
+    public void addTask(Task task) {
+        try {
+            BufferedWriter w = new BufferedWriter(new FileWriter(path, true));
+            w.write(task.getInfo() + System.lineSeparator());
+            w.close();
+        } catch (IOException e) {
+            // ignore since we are sure that we can read and write to file
```
shawnkoh, line 63: actually it may still fail. it would be better if you threw the exception or handled it by printing an error message.

src/main/java/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+            tasks.forEach(task -> {
+                try {
+                    w.write(task.getInfo() + System.lineSeparator());
+                } catch (IOException e) {
+                    System.err.println("Unable to update storage: " + e.getMessage());
+                }
+            });
+            w.close();
+        } catch (IOException e) {
+            // we are certain that we are able to read and write to file
```
shawnkoh, line 84: as above. it may still fail.

src/main/java/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    private class LoadException extends IOException {
+        public LoadException() {
+            super();
```
shawnkoh, line 90: unnecessary constructor

src/main/java/Task.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Calendar;
+import java.util.Date;
+
+/**
+ * An <code>abstract</code> class representing a task which has name and status (whether it is done) fields.
+ */
+abstract public class Task {
+    private static final int CURRENT_YEAR = Calendar.getInstance().get(Calendar.YEAR);
+    private static SimpleDateFormat dateFormatter = new SimpleDateFormat("d/M HHmm");
+    protected static final String SYMBOL_COMPLETE = "\u2714";
```
shawnkoh, line 12: i like how you declared these as constants.

src/main/java/Task.java
```diff
@@ -0,0 +1,96 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+
+/**
+ * An <code>abstract</code> class representing a task which has name and status (whether it is done) fields.
+ */
+abstract public class Task {
+    private static final int CURRENT_YEAR = Calendar.getInstance().get(Calendar.YEAR);
+    private static SimpleDateFormat dateFormatter = new SimpleDateFormat("d/M HHmm");
```
shawnkoh, line 11: it would be better if dateFormatter was decoupled from Task.

consider having a DateUtil class.

src/main/java/Task.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    protected static Calendar parseDate(String str) {
+        Calendar c = Calendar.getInstance();
+        try {
+            Date date = dateFormatter.parse(str);
+            c.setTime(date);
+            c.set(Calendar.YEAR, CURRENT_YEAR);
+        } catch (ParseException e) {
+            // won't fall into this block since this function is only used when str is valid
```
shawnkoh, line 74: It is better to practice defensive programming where possible instead of ignoring exceptions.
In this instance, if you're really sure it wont happen, you should throw the exception instead.

shawnkoh (commented):
Nice use of functional programming!
Your code is good. But it would be better if it was decoupled a little bit more, and did not ignore exceptions.

src/main/java/AddCommand.java
pohlinwei: I see! In hindsight, yeah, that is a good idea. I have seen people doing it but I previously just found it quite redundant. Thanks for pointing it out!

src/main/java/Command.java
pohlinwei: I see! I have thought of that. But do you think that it is too verbose if I explicitly list all exceptions only to perform the same action for each exception later?

src/main/java/Duke.java
pohlinwei: Haha. I just forgot to delete the comment 😅 Thanks though. I'll remove it later. (:

src/main/java/Deadline.java
pohlinwei: Could you clarify this? I don't get really get it. 😅 If I don't have the else statement and o is not an instance of deadline, wouldn't an exception be thrown? 

Thanks!

src/main/java/EmptyDescriptionException.java
pohlinwei: Would you suggest not including the constructor method all together then? I included the super call so that I won't have an empty constructor. (If you are wondering, I am trying to avoid empty constructor so that I won't have "blanks" in my Javadoc) Thanks once again!

src/main/java/Deadline.java
shawnkoh:
```
def abc {
  if (true) {
    return anything;
  }
  print("abc");
}
```

abc will never print

src/main/java/Task.java
pohlinwei: Didn't think of that. However, now that you mention it, I think that it is indeed a good point. Thanks!

src/main/java/Task.java
pohlinwei: However, if I throw the exception, shouldn't I catch it? I decided not to do that because it is too much work. HAHA.

src/main/java/Command.java
shawnkoh: Yeah it can be annoying to. But I think it’s a good practice to have.

src/main/java/Task.java
shawnkoh: Haha, it is more work - there is an easier way though, you can just continue throwing it in the code that calls this method

shawnkoh:
I wasn't satisfied with how I was handling exceptions previously so I went to read up on exceptions [here](https://github.com/tatsuyaoiw/effective-java/blob/master/chapter-9.md) and consulted a really experienced friend, and found that I violated two good practices of exception, so I would like to share them here.

1. avoid catching runtime exceptions - they should be handled using if/else instead. a good example of this would be user input errors (e.g. delete a instead of delete 2)
2. exceptions should be for truly exceptional cases that are out of your control; such as when you try to save the data to the file and for some reason it does not work i.e. IOException. For exceptions that you can control like FileNotFoundException, they should be handled.

Consider checking out my duke repo again to see how I handle exceptions now. I don't think its perfect code but it should definitely be far better than what it was previously.

src/main/java/Task.java
pohlinwei: I see! Okay, thanks for all the advice. All the best with your project 😄

PR #246, opened 2019-08-28
[Shao Yi] Duke Increments

shaoyi1997:


src/main/java/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        SimpleDateFormat format = new SimpleDateFormat("dd/MM/yyyy HHmm");
+        Task tk;
+        int idxOfSlash;
+
+        try {
+            if (input.length() >= 4) {
+
+                // add Deadline
+                if (input.length() >= 8 && input.substring(0, 8).equals("deadline")) {
```
dalsontws: These conditions can work, but maybe you can try to use a .contains function?
for e.g.
input.contains("deadline")
This can simplify the entire conditional statement.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+                    idxOfSlash = input.indexOf("/by ");
+                    if (idxOfSlash == -1 || input.split(" ")[1].contains("/")) {
+                        throw new DukeException("The description of the deadline is incorrect.");
+                    }
+                    tk = new Deadline(input.substring(9, idxOfSlash - 1),
+                            format.format(format.parse(input.substring(idxOfSlash + 4))));
+                }
+
+                // add Event
+                else if (input.length() >= 5 && input.substring(0, 5).equals("event")) {
```
dalsontws: Same as line 46.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+                    idxOfSlash = input.indexOf("/at ");
+                    if (idxOfSlash == -1 || input.split(" ")[1].contains("/")) {
+                        throw new DukeException("The description of the event is incorrect.");
+                    }
+                    tk = new Event(input.substring(6, idxOfSlash - 1),
+                            format.format(format.parse(input.substring(idxOfSlash + 4))));
+                }
+
+                // add ToDos
+                else if (input.substring(0, 4).equals("todo")) {
```
dalsontws: Same as line 46.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private String input;
+
+    /**
+     * Constructor for the AddCommand object.
+     *
+     * @param ipt the string representation of the command input.
+     */
+    public AddCommand(String ipt) {
+        input = ipt;
```
dalsontws: Not compulsory, but would be good to have a 'this' keyword to reference the current object. 
e.g. this.input = ipt
Would become ambiguous if there are similar variable names.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Constructor for Deadline task.
+     * Initializes the description and due datetime.
+     *
+     * @param des description of the deadline.
+     * @param by datetime for which the deadline is due.
+     */
+    public Deadline(String des, String by) {
+        super(des);
+        this.by = by;
```
dalsontws: "this" keyword used for your tasks constructors, but some methods not used. 
Good to be consistent with the usage!

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private int taskNumber;
+
+    /**
+     * Constructor for the delete command.
+     *
+     * @param taskNum the index of the task to be deleted
+     */
+    public DeleteCommand(int taskNum) {
+        taskNumber = taskNum;
```
dalsontws: Same as the addCommand method in the AddCommand Class

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private int taskNumber;
+
+    /**
+     * Constructor for the done command.
+     *
+     * @param taskNum the index of the task to be marked done.
+     */
+    public DoneCommand(int taskNum) {
+        taskNumber = taskNum;
```
dalsontws: Same as the addCommand method in the AddCommand Class

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private String input;
+
+    /**
+     * Constructor for the find command.
+     *
+     * @param ipt the string of search query.
+     */
+    public FindCommand(String ipt) {
+        input = ipt;
```
dalsontws: Same as the addCommand method in the AddCommand Class

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Indicates whether this is an exit command.
+     *
+     * @return false.
+     */
+    public boolean isExit() { return false; }
```
dalsontws: Would be clearer and improve readability to have your return statement on a new line!
Maybe you can try running the CheckStyle tool? 
Would help to identify these issues!

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param tasks tasklist that stores all the tasks
+     * @param ui ui object that deals with user interaction
+     * @param storage storage object that deals with writing into the text file
+     * @return string representation of the command response
+     * @throws DukeException exception is thrown when the index falls outside the number of tasks
+     */
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        try {
+            if (taskNumber > tasks.getNumOfTask() || taskNumber < 1) {
+                throw new DukeException("List only has " + tasks.getNumOfTask() + " items.");
```
wilfredbtan: Should the DukeException be thrown within this try catch block within an if statement? Logic might be confusing. Maybe you can consider a way to catch this error and throw it below as well.

src/main/java/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param input the string representation of the command input by the user.
+     * @return the command object given by the input.
+     * @throws DukeException thrown when comm
+     */
+    public static Command parse(String input) throws DukeException {
+        if (input.equals("bye")) {
+            return new ExitCommand();
+        } else if (input.length() <= 3) {
+            throw new DukeException("I'm sorry, but I don't know what that means :-(");
+        } else if (input.substring(0, 4).equals("find")) {
```
wilfredbtan: Is there a better method to check whether a string "contains" something? Also does this catch the errors resulting from invalid inputs for find, done, delete, etc?

PR #248, opened 2019-08-28
[Alex Koh Nan Yu] Duke Increments

alxkohh:
Duke increments


src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                    sb.append(list.get(i).toTextFileString());
+                } else {
+                    sb.append(list.get(i).toTextFileString());
+                    sb.append("\n");
+                }
+            }
+            String content = sb.toString();
+            writeToFile(content);
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
```
seahlynn, line 100: Perhaps you can catch IOException and deal with it by throwing a DukeEception (like you did in the load() function above! So all the errors thrown during the execution of the program can be standardised to DukeExceptions! Although I don't really see anything wrong with this HAHAHA

src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> outputList = new ArrayList<>();
+        try {
+            File f = new File(filePath);
+            Scanner s = new Scanner(f);
+            while (s.hasNext()) {
+                String savedTask = s.nextLine();
+                String[] savedTaskSplit = savedTask.split("\\|");
+                if (savedTaskSplit[0].equals("T")) {
```
seahlynn, line 34: Perhaps you could use switch-case statements instead of if-else loops here to make the different scenarios clearer to the reader!

seahlynn (commented):
Overall clear code with accurate naming of variables!! However, could try to use switch-case statements (if possible) to replace if-else loops! Good job m8!!

src/main/java/TaskList.java
```diff
@@ -1,4 +1,75 @@
-package PACKAGE_NAME;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 
 public class TaskList {
+
+    private static DateTimeFormatter dukeDateTimeFormatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+    private ArrayList<Task> dukeTaskList = null;
```
seahlynn: Not necessary to declare it as null in the variables field (I think??) Please correct me if I'm wrong!! (Also this is not a very constructive feedback, I'm just curious whether it's necessary HAHA)

src/main/java/TaskList.java
```diff
@@ -1,4 +1,75 @@
...
+    }
+
+    public void addTaskAfterValidation(String dateTimeString, Task t) {
+        try {
+            LocalDateTime dateTime = LocalDateTime.parse(dateTimeString, dukeDateTimeFormatter);
+
+            if (t instanceof Event) {
+                Event e = (Event) t;
+                e.setDateTimeAt(dateTime);
+            } else {
```
seahlynn, line 84: Perhaps it could be better to write "else if ( t instanceof Deadline) { ... } " and use the else-block for catching exceptions etc, because according to coding standards the final else "should mean everything else", and not the final option!

seahlynn (commented):
Variables and Methods are named very clearly and concisely!! Good job!!

src/main/java/TaskList.java
```diff
@@ -1,4 +1,75 @@
...
+                Deadline d = (Deadline) t;
+                d.setDateTimeBy(dateTime);
+            }
+
+            addTask(t);
+            System.out.println("Got it. I've added this task:");
+            System.out.println(t);
+            System.out.println("Now you have " + dukeTaskList.size() + " tasks in the list.");
+        } catch (Exception e) {
+            System.out.println("Task not added to list because the input format for date and time is unrecognised. " +
```
seahlynn: Might be more appropriate to throw a DukeException here with the same error message as displayed!

src/main/java/UnrecognisedInputCommand.java
```diff
@@ -0,0 +1,7 @@
+public class UnrecognisedInputCommand extends Command {
```
seahlynn, line 4: You are the first person I see with such a command!!! Why didn't you just throw an exception to deal with unrecognised input? (I am genuinely curious!!)

src/main/java/Parser.java
```diff
@@ -0,0 +1,37 @@
+public class Parser {
+
+    public static Command parse (String input) {
+        String[] inputSplit = input.split(" ");
+        if (input.equals("bye")) {
```
seahlynn, line 14: Might be more appropriate to use switch-case statements here to organise the code and make the different commands more apparent to the readers!

seahlynn (commented):
Command classes are named very clearly and distinctly! Think it's very interesting how you handle unrecognised input as a class rather than an exception! I'm not sure whether that would be redundant though!

src/main/java/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+            return new ExitCommand();
+        } else if (input.equals("list")) {
+            return new DisplayListCommand();
+        } else if (inputSplit[0].equals("delete")) {
+            int inputIndex = Integer.parseInt(inputSplit[1]);
+            int actualIndex = inputIndex - 1;
```
seahlynn, line 20: I really like how you split the different processing of the commands into temp variables which makes it very easy to follow!!! :)

src/main/java/DisplayListCommand.java
```diff
@@ -0,0 +1,9 @@
+public class DisplayListCommand extends Command {
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        System.out.println("Here are the tasks in your list:");
```
seahlynn: I think you should try to use Ui here? Instead of system.out.println!

src/main/java/AddEventCommand.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public void execute (TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        if (inputEvent.trim().length() == 5) {
+            throw new DukeException("☹ OOPS!!! The description of a event cannot be empty.");
+        } else {
+            int slashLocation = slashLocator(inputEvent);
+            int firstIndex = slashLocation - 1;
+            int secondIndex = slashLocation + 4;
+            String eventDescription = inputEvent.substring(6, firstIndex);
+            String eventAt = inputEvent.substring(secondIndex);
```
seahlynn: Think it would be good to throw a DukeException for the users forgetting to input event time!

seahlynn (commented):
Simple and concise tests with clear method names! Easily understandable test codes good job!! Maybe you could also test other classes not limited to "tasks" like Parser and TaskList, because I think these classes are also a pretty integral part of the main code!

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,25 @@
+import java.util.ArrayList;
+
+public class FindCommand extends Command {
+
+    String keyword = "";
+
+    public void setKeyword(String keyword) {
```
seahlynn, line 7: Perhaps "keyword" could be passed into the object when it is constructed, instead of having to create another method to set keyword under the object!

seahlynn (commented):
I noticed here that the different tasks commands are named abit differently? It might be better to standardise all the namings to "AddEventCommand", "AddDoneCommand" and "AddTodoCommand"! Also, it seems that the Ui object isn;t really used in most of the execute methods, so it might perhaps be better for Ui to be removed as a parameter for execute() because it seems redundant!

src/main/java/Task.java
```diff
@@ -1,30 +1,56 @@
...
     }
 
+    /**
+     * Converts a task object into string form to be used
+     * in storage text file.
+     * @return string representation of a task on the storage text file
+     */
     public String toTextFileString() {
         String status = "";
         if (this.isDone == true) {
```
seahlynn: Think this might be redundant! if(this.isDone) { ... } would be sufficient in checking if isDone is true or false

seahlynn (commented):
Very concise description of the different classes and their methods! Your javadocs comments really makes clear to the reader what the code/method id supposed to do!! However, I would like to suggest that in the description of the execute methods for the different command objects, perhaps you could be more specific about when (the scenario) a DukeException would be thrown, to help readers understand the purpose of the exception!

data/tasks.txt
```diff
@@ -0,0 +1,3 @@
+D|1|eat|2/2/2019 2200
```
jonchan51: Good to make it a habit to avoid committing data files like this, so that if a user decides to clone your repository, they do not start with data they aren't supposed to have.

src/main/java/AddDeadlineCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Overridden method. Executes the add deadline command.
+     * @param tasks list of tasks
+     * @param ui user interface
+     * @param storage storage file
+     * @throws DukeException exception specific to Duke application
+     */
+    @Override
+    public void execute (TaskList tasks, Ui ui, Storage storage) throws DukeException {
```
jonchan51: minor styling, `execute(TaskList tasks...`

src/main/java/AddDeadlineCommand.java
```diff
@@ -0,0 +1,43 @@
...
+        if (inputDeadline.trim().length() == 8) {
+            throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
+        } else {
+            int slashLocation = slashLocator(inputDeadline);
```
jonchan51, line 24: I think it would be clearer if you added a comment explaining your method here, why -1? why +4?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Encapsulates a deadline.
+ * Each deadline has a time when the deadline takes effect,
+ * stored in both String and LocalDateTime formats.
+ * A subclass of Task.
+ */
+public class Deadline extends Task {
+
+    protected String by;
```
jonchan51: Is there a reason why you chose to have it protected instead of private?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,130 @@
...
+            if (t instanceof Event) {
+                Event e = (Event) t;
+                e.setDateTimeAt(dateTime);
+            } else {
+                Deadline d = (Deadline) t;
+                d.setDateTimeBy(dateTime);
+            }
+
+            addTask(t);
+            System.out.println("Got it. I've added this task:");
```
jonchan51, line 90: Think it would be better to abstract out all forms of user interaction to your Ui class

src/main/java/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+                    String description = savedTaskSplit[2];
+                    String by = savedTaskSplit[3];
+                    Task deadlineTask = new Deadline(description, by);
+                    if (savedTaskSplit[1].equals("1")) {
+                        deadlineTask.markAsDone();
+                    }
+                    outputList.add(deadlineTask);
+                }
+            }
+        } catch (FileNotFoundException e) {
```
jonchan51, line 62: Would be good to catch the case where the input format is incorrect (Normally ArrayOutOfBounds if they modified and added extra details.

src/main/java/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+            DeleteCommand delete = new DeleteCommand();
+            delete.setIndexToRemove(actualIndex);
+            return delete;
+        } else if (inputSplit[0].equals("done")) {
+            int inputIndex = Integer.parseInt(inputSplit[1]);
+            int actualIndex = inputIndex - 1;
+            DoneCommand done = new DoneCommand();
+            done.setIndexToMarkDone(actualIndex);
```
jonchan51: this could have been done in the constructor for DoneCommand and DeleteCommand!

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
...
+        }
+    }
+
+    /**
+     * Runs the application. A scanner is instantiated to read standard input.
+     */
+    public void run() {
+        ui.showWelcome();
+        Scanner sc = new Scanner(System.in);
+        while (sc.hasNextLine()) {
```
jonchan51, line 53: i think doing it this way means that the "bye" command does not stop the app

jonchan51 (commented):
Easy to read through your code with your variable naming, even with the slightly more obscure parts its still understandable. Think most of the things I noticed are additional things that aren't necessary, so good job so far!

src/main/java/Storage.java
alxkohh: Yes, good idea since there are many cases to consider. I used if-else blocks due to personal preference, hope the readability was still fine for you!

src/main/java/TaskList.java
alxkohh: Yes, you are right, thanks for pointing out! Instance and Class variables do not need to take a null value when they are declared. Local variables (Those declared inside methods) however, need to take a null value when first declared, from what I understand. Thank you :)

src/main/java/UnrecognisedInputCommand.java
alxkohh: I see, it was just part of my design to abstract out as many commands as possible. Abstraction, together with descriptive names for abstracted units (Methods, Classes, etc) also helps me to understand my own code better when the project gets more developed and demanding.

src/main/java/Parser.java
alxkohh: Thank you! Hope they were as helpful to me as they were to you :)

src/main/java/Task.java
alxkohh: You are right! Thanks for reminding, will amend accordingly.

src/main/java/AddDeadlineCommand.java
alxkohh: Thanks for pointing out, I have removed the white space.

src/main/java/TaskList.java
alxkohh: Good idea to make the whole application more object oriented, will refactor accordingly! Thank you!

src/main/java/Duke.java
alxkohh: Actually "bye" does close the app. The app will exit with an exit code of 0. I invoked System.exit(0) in a method inside the Ui class. Just for your information, thank you for the comment though!

src/main/java/AddDeadlineCommand.java
alxkohh: Thanks for this suggestion, I think that comment would be very helpful for both myself and also other developers. Will add that in accordingly, thank you!

PR #249, opened 2019-08-28
[Daniel Tan] Duke Increments

CarbonGrid:


shihaoyap:
Hi Daniel! While looking through your program, i did not find any coding style violations. Well done! :) I found your code structure to be really neat and readable and it follows good OOP principles. I also like the fact that you did prepare for multiple scenarios of invalid input function, and your program was able to handle the different input exceptions really well, when i was testing it 👍 

Overall the program was well written! I tried hard to break it but was not able to haha! Well done! i look forward to seeing your program improve! 

Suggestions - Could explore creating a UI (as stated in Level-10) to be used to interact with the users.

CarbonGrid:
> Hi Daniel! While looking through your program, i did not find any coding style violations. Well done! :) I found your code structure to be really neat and readable and it follows good OOP principles. I also like the fact that you did prepare for multiple scenarios of invalid input function, and your program was able to handle the different input exceptions really well, when i was testing it 👍
> 
> Overall the program was well written! I tried hard to break it but was not able to haha! Well done! i look forward to seeing your program improve!
> 
> Suggestions - Could explore creating a UI (as stated in Level-10) to be used to interact with the users.

@shihaoyap Thanks for the in-depth review. 

Happy to meet the expectations. :smiley: 
Sorry for the letdown. 
Couldn't commit Level-10 early cause still :mask:

src/main/java/Launcher.java
```diff
@@ -0,0 +1,8 @@
+import javafx.application.Application;
+
```
lrchema, line 2: please consider adding javadoc comments to this class

src/main/java/Main.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private Duke duke = new Duke("data/duke.ser").usingMainUi();
+
+    /**
+     * MainWindow loaded from MainWindow.
+     * DialogBox initialised from DialogBox.java instead of loaded from fxml.
+     */
+    @Override
+    public void start(Stage stage) {
+        assert stage != null;
```
sianghwee, line 23: Give a description/comment for your assert to aid troubleshooting

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,76 @@
+import duke.DialogBox;
+import duke.Duke;
+
+import duke.MainUi;
```
sianghwee: It's good that you never use wildcard imports but it would be better if you grouped the imports by packages.

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,76 @@
...
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+        MainUi tmpUi = new MainUi();
+        tmpUi.showWelcome();
+        dialogContainer.getChildren()
+            .add(DialogBox.getDukeDialog(new Label(tmpUi.getResponse()), new ImageView(dukeImage)));
+    }
+
+    public void setDuke(Duke d) {
```
sianghwee, line 45: Write a javadoc to explain this method.

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,90 @@
...
+    private ImageView displayPicture;
+
+    /**
+     * Initializes a DialogBox with Label as text content & ImageView as Profile Image.
+     * DialogBox always force enable word wrap for Label and prevents truncations.
+     *
+     * @param l  String textContent
+     * @param iv Profile Image for DialogBox
+     */
+    public DialogBox(Label l, ImageView iv) {
```
sianghwee, line 26: I feel that it might be better if you use variable names that are more descriptive especially for such a long method.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,80 @@
...
+        assert input != null;
+        assert ui instanceof MainUi;
+
+        try {
+            Command c = Parser.parse(input);
+            c.execute(tasks, ui, storage);
+        } catch (DukeException e) {
+            ui.show(e.getMessage());
+        } finally {
+            return ((MainUi) ui).getResponse();
```
sianghwee, line 79: Why did you need to typecast your Ui?

src/main/java/duke/MainUi.java
```diff
@@ -0,0 +1,21 @@
+package duke;
```
sianghwee: Perhaps you can consider having a separate sub-package (duke.ui) for all your Ui to improve readability.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,18 @@
+package duke.command;
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+import duke.exception.DukeException;
+
+public class FindCommand extends Command {
```
lrchema, line 7: I think it would be good to add javadoc comments here

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+     * @throws LoadFileFailDukeException if file or raw data can't be loaded
+     */
+    @SuppressWarnings("unchecked")
+    public TaskList load() throws LoadFileFailDukeException {
+        try {
+            ObjectInputStream ois = new ObjectInputStream(new FileInputStream(filePath));
+            Place.setRefLatLong((HashMap) ois.readObject());
+            Place.setRefAlias((HashMap) ois.readObject());
+            return (TaskList) ois.readObject();
+        } catch (Exception e) {
```
blimyj: I don't think it's a good idea to catch a general exception. For example, if it is a FileNotFoundException, a new file could be created with a clearer error message printed.

src/main/java/duke/MainUi.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class MainUi extends Ui {
```
sianghwee: Include a Javadoc for your classes as it is unclear what each class does.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param str Expects string to be parse as duke.command.Command.
+     * @return duke.command.Command object as per parsed.
+     * @throws DukeException if command can't be parsed.
+     */
+    public static Command parse(String str) throws DukeException {
+        assert str != null;
+
+        str = str.trim();
+        int lastIndex = str.indexOf(' ');
```
lrchema: This seems a little complicated. Can you achieve the same outcome by using .split()?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        case "delete":
+            return new DeleteCommand(str.substring(lastIndex));
+        case "done":
+            return new DoneCommand(str.substring(lastIndex));
+        case "find":
+            return new FindCommand(str.substring(lastIndex));
+        case "list":
+            return new ListCommand();
+        case "deadline":
+            temp = str.substring(8)
```
sianghwee: Perhaps you can try not to use magic literal in your code as it will be more intuitive for the next developer.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+        this.filePath = filePath;
+    }
+
+
+    /**
+     * Try to load raw data from filePath and deserialize everything.
+     *
+     * @throws LoadFileFailDukeException if file or raw data can't be loaded
+     */
+    @SuppressWarnings("unchecked")
```
sianghwee, line 33: Perhaps you can try to resolve the warning instead of suppressing the warning as it might lead to more underlying bugs.  Is there a reason for this decision? 😄

sianghwee (commented):
Sorry, we were only able to review to the Storage.java as we were pressed for time. Thank you for sharing your code and we were able to learn a lot from it.

src/main/java/duke/Storage.java
CarbonGrid: I tried searching for an appropriate way to handle reading Objects. Unfortunately, all solutions I have come across so far has no "clean" way of dealing with Object casting as there is an implied order of processing.

src/main/java/duke/Storage.java
CarbonGrid: Agreed. Scope should always be minimal.

src/main/java/duke/Parser.java
CarbonGrid: Definitely yes, I did a bit of hand-optimization here. As you can see, split is only good for one word. As you know split would process split at all white-spaces. But I just figured I could implement using lazy look-ahead. Hence, great thought from you that I would improve.

src/main/java/duke/MainUi.java
CarbonGrid: Observant enough. 😄  I did the rest of the package structure in such a way that inheriting class under parent class is of same sub-package. However, forgotten about MainUi extends Ui.

src/main/java/duke/MainUi.java
CarbonGrid: Without constructor JavaDoc, I believe that is necessary. Good suggestion.

src/main/java/duke/Parser.java
CarbonGrid: Bad practice indeed.

src/main/java/duke/command/FindCommand.java
CarbonGrid: I'm not sure if it is needed here. But I think I should at least have a Javadoc at least in the constructor or execute().
After all one would know what FindCommand probably is from inheritance but one wouldn't know what it does on execution.

src/main/java/MainWindow.java
CarbonGrid: Not sure when that line break got in. But good observation.

src/main/java/Main.java
CarbonGrid: I guess it differs from person to person. But IMO, it should be self-explanatory enough that I wouldn't think of placing any comment.

src/main/java/MainWindow.java
CarbonGrid: IMO, this is self-explanatory and simple. Sometimes it's hard for me to justify whether getters & setters deserve JavaDoc.

src/main/java/duke/Duke.java
CarbonGrid: MainWindow uses MainUi while Ui doesn't have any definition of getResponse(). There is alternative ways of cleaner implementation but I believe but I thought it would be simpler to cast it. Might put this on forum for discussion on good practice for non-existent subclass methods in super-class.

src/main/java/Launcher.java
CarbonGrid: I was considering to remove this class. After all, it's a Launcher that does nothing.

src/main/java/duke/DialogBox.java
CarbonGrid: Justified.

src/main/java/Main.java
sianghwee: Hello, I think I was not clear in my initial comment. What I meant was that when you assert, perhaps you can do `assert stage != null : "Stage not found"; ` as that will make it clear what the problem is when your assert fails.

CarbonGrid: 😄 I was wondering if you meant an assert message, which is great. But I concluded you suggest me to leave a comment by the assert.

PR #250, opened 2019-08-28
[Liu Guangyuan Gary] Duke Increments

gary-lgy:


docs/README.md
```diff
@@ -1,20 +1,121 @@
...
+### Available in both CLI and GUI
+GUI:
+
+![GUI version](./gui_demo.gif)
+
+CLI:
+
+![CLI version](./cli_demo.gif)
+
+While the GUI may be more intuitive to use, the CLI version provides a simpler
```
nattanyz, line 33: a clear readme overall! :) having both GUI and CLI accessible from the same JAR file is quite a good idea!

src/main/java/duke/Launcher.java
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            launchGui();
+        }
+    }
+
+    private static boolean hasValidArgs(String[] args) {
+        return args.length == 0 || args.length == 1 && VALID_ARGS.contains(args[0]);
+    }
+
+    private static String getMode(String[] args) {
```
nattanyz, line 41: clean implementation of command-line argument parsing! would it make sense to use an enum for the two types of UI?

src/main/java/duke/cli/Cli.java
```diff
@@ -0,0 +1,146 @@
...
+        show("Warning: " + warning);
+    }
+
+    /**
+     * Shows each error on a new line.
+     *
+     * @param error the error to show
+     */
+    @Override
+    public void showError(final String error) {
```
nattanyz, line 112: since the implementations of showMessage, showWarning and showError don't differ, would it make sense to combine these methods? or is this difference due to GUI implementation? :)

src/main/java/duke/command/AddTaskWithTimeCommand.java
```diff
@@ -0,0 +1,15 @@
+package duke.command;
+
+import java.time.LocalDateTime;
+
+/**
+ * Command that adds a new task with a datetime field.
+ */
+public abstract class AddTaskWithTimeCommand extends AddTaskCommand {
```
nattanyz, line 8: abstracting out the tasks with datetime fields is a good idea! +1 from me :)

src/main/java/duke/gui/MainWindow.java
```diff
@@ -0,0 +1,131 @@
...
+        if (input.isBlank()) {
+            return;
+        }
+        addUserDialog(input);
+        userInput.clear();
+
+        try {
+            Command command = Parser.parse(input);
+            command.execute(tasks, this, storage);
+            if (command.shouldExit()) {
```
nattanyz, line 92: ah, so that's how you implemented the exit on Duke's GUI! this got me stumped for quite a while. good job :)

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * The Command base class.
+ */
+public abstract class Command {
+    /**
+     * Checks whether this Command is valid.
+     * @param tasks the current tasks
+     * @throws DukeInvalidCommandException if this Command is invalid
+     */
+    protected abstract void check(final TaskList tasks) throws DukeInvalidCommandException;
```
nattanyz, line 18: it's not immediately clear why checking if the command is valid -- involving everything from command syntax to missing arguments to invalid operations depending on the state of the task list -- should be encapsulated in the same method and take the TaskList as an argument. however, you seem to have implemented it quite well! I like how you've made all these various exceptions DukeInvalidCommandException. 👍

nattanyz (commented):
hi Gary, thanks for your thorough review of my code earlier on! :) it was pretty helpful -- and quite enlightening for me to see how much my code has improved since! sorry for not responding to your comments earlier. hope some of these comments are helpful to you, and I'll add more if I ever have the time! :)

PR #252, opened 2019-08-28
[Yeo Yik Hwee Ernest] Duke Increments

ernestyyh:
Add Duke increments up to A-JavaFx


src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                String[] savedTask = s.nextLine().split(" \\| ");
+
+                switch (savedTask[0]) {
+                case "T":
+                    Todo t = new Todo(savedTask[2]);
+                    if (savedTask[1].equals("1")) {
+                        t.markAsDone();
+                    }
+                    taskList.add(t);
+                    break;
```
halfwhole: Perhaps marking the task as done and adding it to the task list could be extracted out, instead of doing it over and over across each `case` statement (one for todo, one for deadline, and one for event)?

src/main/java/Todo.java
```diff
@@ -0,0 +1,14 @@
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
+
+    public String formattedString() {
+        return String.format("T | %d | %s", super.isDone ? 1 : 0, super.description);
```
halfwhole, line 39: An `@Override` could be helpful here!

src/main/java/Ui.java
```diff
@@ -0,0 +1,87 @@
+import java.util.Scanner;
+
+public class Ui {
+    public void initiate(Storage storage, TaskList taskList) {
+        String logo = " ____        _        \n"
```
halfwhole, line 37: This method's functionality should probably belong to the `Duke` class instead? Seems that the main loop for the program is now placed here, which isn't what we typically expect a UI component to do -- to only handle inputs/outputs.

halfwhole (commented):
Great! Nice use of tests and OOP patterns.

SimNee:
A good OOP practice but for your different commands, maybe you could use classes like "ByeCommand", "ListCommand" etc to make it more OOP.
For the Exceptions classes maybe do not need so many classes also.
Else, the rest of the comments already mentioned by halfwhole. 
But overall good job :)

PR #253, opened 2019-08-28
[Choong Yong Xin] Duke Increments

choongyx:
Up to Week 4: Level-1, Level-2, Level-3, Level-4, Level-5, Level-6, Level-7, Level-8, A-MoreOOP, A-Packages, A-JUnit, A-Jar, A-JavaDoc, A-CodingStandard, Level-9, A-CheckStyle, Level-10.


data/tasks.txt
```diff
@@ -0,0 +1,10 @@
+T | ? | eat
+E | ✘ | meeting | 20180909 0900
+E | ✘ | meeting | 12/12/2019 900
```
weikiat97: Why are the dates and time in different format over here? Also, for event it is a time period as mentioned in the question so might want to add a range instead of just a start time.

src/main/java/Command.java
```diff
@@ -14,5 +14,5 @@
 
     abstract boolean isExit();
 
-    abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException, IOException;
+    abstract String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException, IOException;
```
weikiat97: Consider including JavaDocs for the method.

src/main/java/Duke.java
```diff
@@ -3,14 +3,36 @@
...
+    private TextField userInput;
+    private Button sendButton;
+    private Scene scene;
+
+    private Image user = new Image(this.getClass().getResourceAsStream("/images/User.png"));
+    private Image duke = new Image(this.getClass().getResourceAsStream("/images/Quack.png"));
+
+    public Duke() throws IOException {
+        String filePath = "/data/tasks.txt";
         ui = new Ui();
```
weikiat97: Do you still need a UI now since everything is done in the GUI?

src/main/java/Duke.java
```diff
@@ -22,26 +44,119 @@ public Duke(String filePath) throws IOException {
...
+    private void handleUserInput() {
+        String text = userInput.getText();
+        Label userText = new Label(userInput.getText());
+        String responseText = getResponse(userInput.getText());
+        Label dukeText = new Label(responseText);
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(userInput.getText(), user),
+                DialogBox.getDukeDialog(responseText, duke)
+        );
+        if (text.equals("bye")) {
```
weikiat97: Good that you handled the exit condition here!

src/main/java/Duke.java
```diff
@@ -22,26 +44,119 @@ public Duke(String filePath) throws IOException {
...
-                isExit = c.isExit();
-            } catch (DukeException e) {
-                ui.showError(e.getMessage());
-            } finally {
-                ui.showLine();
-            }
+    /**
+     * Sets up GUI.
+     */
+    @Override
```
weikiat97: Setup of GUI should be under the MainWindow.

weikiat97 (commented):
Overall, there is some inconsistency in the JavaDocs (certain methods current do not have them). It is good that you handled different kinds of DukeException which makes it clearer for the reader. It is pretty neat and readable!

src/main/java/TodoCommand.java
```diff
@@ -35,15 +35,26 @@ String execute(TaskList tasks, Storage storage) throws DukeException, IOExceptio
...
+     * Forms the output string.
+     *
+     * @param tasks TaskList containing the tasks
+     * @param newTodo new Todo to be added
+     * @return output string
+     */
+    private String getOutput(TaskList tasks, Todo newTodo) {
+        String output = "Got it. I've added this task: \n";
+        output += (newTodo + "\n");
+        output += "Now you have " + tasks.taskList.size() + " tasks in the list.\n";
```
bitterg0d, line 57: Good job on your refactoring! Your variables are very appropriately named. The only thing I can suggest for improvement is maybe tailoring the output line to print "task" or "tasks" depending on whether the number of tasks in your list is <= 1 or otherwise

PR #254, opened 2019-08-29
[Phoebe] Duke Increments

febee99:
Up to Week 3 Increments.


data/tasks.txt
```diff
@@ -0,0 +1,7 @@
+D|0| return book |Mon, 2 Dec 2019, 06:00 PM
+T|1|make my bed
+D|0|Quiz 1 | tomorrow
+T|0|Make bed
+T|0|make bed
+T|0|read book
+D|0|return my stuff |tomorrow
```
niqiukun: You may consider adding this file into your `.gitignore` to prevent it from being tracked by git.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,53 @@
...
+import duke.tasks.TaskList;
+import duke.ui.Ui;
+
+public class Duke {
+    private final Storage storage;
+    private final Ui ui;
+    private TaskList tasks;
+
+    /**
+     * Initialise new duke.Duke session.
```
niqiukun: Note that according to the coding standard the JavaDoc should starts with `Initialises ...`.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,135 @@
...
+     * Executes the AddCommand, adds the task to the LinkedList and saves it to the file.
+     *
+     * @param tasks   The TaskList containing all existing tasks.
+     * @param ui      The Ui for printing purposes.
+     * @param storage The Storage for saving tasks to file.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+
+        tasks.allTasks.add(task);
+        StringBuilder sb = new StringBuilder("Got it. I've added this task:\n" + task);
```
niqiukun: Although this is fine to me (I did like this also), alternatively, you may consider placing all these messages inside the `Ui` class and call the respective methods in other classes.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param tasks   The TaskList containing all existing tasks.
+     * @param ui      The Ui for printing purposes.
+     * @param storage The Storage for saving tasks to file.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+
+        StringBuilder sb = new StringBuilder("Noted. I've removed this task:\n" + tasks.allTasks.get(index));
+        storage.deleteTaskFromFile(tasks.allTasks.get(index));
+        tasks.allTasks.remove(index);
```
niqiukun: There could be exception handling here which handles `IndexOutOfBoundsException`. Alternatively, you may consider making it more OOP by encapsulating the deletion within the `TaskList` class. As a result, you only have call `tasks.delete(index)` here.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,51 @@
...
+        if (tasks.allTasks.size() == 1) {
+            sb.append("\nNow you have ")
+              .append(tasks.allTasks.size())
```
niqiukun: Similarly, you may encapusulate the `getSize()` method within the `TaskList` class to prevent the underlying implementation (`ArrayList<Task>`) to be accessed directly.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,25 @@
...
+     * @param storage The Storage for saving tasks to file.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage);
+
+    /**
+     * Checks if it is a bye command.
+     *
+     * @return A boolean: true if it is a bye command.
+     */
+    public abstract boolean isExit();
```
niqiukun, line 23: Consider an alternative implementation: you can implement `isExit` as follows:
```Java
public boolean isExit() {
    return this instanceof ExitCommand;
}
```
As a result, you don't need to reimplement this method in all its child classes.

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,35 @@
...
+import duke.ui.Ui;
+
+public class ExitCommand extends Command {
+
+    /**
+     * Initialises an ExitCommand.
+     */
+    public ExitCommand() {
+
+    }
```
niqiukun, line 13: This may not be necessary?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param tasks   The TaskList containing all existing tasks.
+     * @param ui      The Ui for printing purposes.
+     * @param storage The Storage for saving tasks to file.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+
+        // Print all existing items in the list
+        ui.printMessage("Here are the matching tasks in your list:");
+        for (Task x : tasks.allTasks) {
+            if (x.toString().toLowerCase().contains(keyword)) {
```
niqiukun, line 33: Similar to `DeleteCommand`, you may consider making it more OOP by encapsulating the finding process within the `TaskList` class.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+        String[] details = fullCommand.trim().split("\\s+");
+
+        try {
+            switch (details[0]) {
+            case "list":
+                return new ListCommand();
+            case "find":
+                return new FindCommand(details[1]);
+            case "done":
+                return new DoneCommand(Integer.parseInt(details[1]) - 1);
```
niqiukun, line 38: You may consider moving the "minus one" process to the lowest level implementation, i.e. inside `TaskList` class, such that all these details will not be revealed to upper level classes.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Reads the input file line by line to add tasks into the TaskList.
+     *
+     * @return The LinkedList containing all loaded tasks.
+     * @throws DukeException In the event that the file cannot be read properly.
+     */
+    public LinkedList<Task> load() throws DukeException {
+        LinkedList<Task> allTasks = new LinkedList<>();
```
niqiukun, line 39: Usually, it is recommended to use `ArrayList` instead of `LinkedList` for faster access of `Tasks`.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Append a new task to the end of the file.
+     *
+     * @param t The task to be added.
+     */
+    public void appendTaskToFile(Task t) {
+        try {
+            BufferedWriter output = new BufferedWriter(new FileWriter(filepath, true));
```
niqiukun, line 99: You can use the `FileWriter` to write file directly without the need for `BufferedWriter`. See the following example:
```Java
FileWriter writer = new FileWriter(filePath, true); // Note the true here to enable append mode
writer.write("This is one line\n");
writer.flush();
writer.close();
```

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Deletes a task from the file.
+     *
+     * @param t The task to be deleted.
+     */
+    public void deleteTaskFromFile(Task t) {
```
niqiukun: For deletion, it may be better to use an integer index as parameter than to use the `Task` object. Not only because the parameter passed between methods will be simpler, it will also avoid the situation where the wrong task is deleted. Consider the case when there are two tasks with the same `toStore()` value. Although they are the same in terms of content, they should be distinct objects.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+        LinkedList<Task> allTasks = new LinkedList<>();
+        try {
+            File f = new File(filepath);
+            if (!f.exists()) {
+                f.createNewFile();
+            }
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(f));
+            String currLine;
+            while ((currLine = bufferedReader.readLine()) != null) {
+                String[] formattedText = currLine.split("\\|");
```
niqiukun: Since the conversion from `Task` objects to strings in files is done in the respective child classes of `Task`, consider placing the opposite conversion in the `Task` class as well. For example, the `Task` class could have the following method:
```Java
public static Task toTask(String text) throws DukeException {
    String[] formattedText = text.split("\\|");
    Task t;

    // Omitted

    return t;
}
```

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private final String by;
+
+    /**
+     * Initialises a Deadline task.
+     *
+     * @param description The deadline description.
+     * @param by          The time at which the deadline is up.
+     */
+    public Deadline(String description, String by) {
```
niqiukun: Using `String` to store the actual deadline is okay to me, but for greater consistency, you may also consider storing it as `Date` objects. When you export the tasks to a file, the deadlines could be converted from `Date` to `String` and hence follow the same format.

src/main/java/duke/tasks/TaskList.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    /**
+     * Initialises an empty TaskList.
+     */
+    public TaskList() {
+        this.allTasks = new LinkedList<>();
+    }
+
+}
```
niqiukun, line 88: As mentioned above, there could be more OOP by encapsulating various methods here, including `addTask`, `deleteTask`, `findTask`, etc.

niqiukun (commented):
Hi Phoebe, your code is quite well written :) Keep up the good work! I have left several comments in the code with alternative implementations so that you may consider improving it further.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,53 @@
...
+        storage = new Storage("data/tasks.txt");
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    public static void main(String[] args) {
```
bjhoohaha: I am not sure whether you may have missed out this but there is missing javadoc for main method.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @return The string representation of the new date object.
+     */
+    public String parseDate() {
+        try {
+            SimpleDateFormat sdf = new SimpleDateFormat("dd/MM/yyyy HHmm", java.util.Locale.ENGLISH);
+            Date myDate = sdf.parse(time.replaceAll("-", "/"));
+            sdf.applyPattern("EEE, d MMM yyyy, hh:mm a");
+            return sdf.format(myDate);
+        } catch (ParseException e) {
```
bjhoohaha: Good effort in parsing the date and applying a different pattern to format to another date format that may be more readable.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return The parsed description.
+     * @throws DukeException In the event that the command is in the wrong format.
+     */
+    public String parseDescription(String delimiter) throws DukeException {
+        String[] details = description.split(delimiter);
+        if (details.length == 1) {
+            if (details[0].length() != 0) {
+                throw new DukeException("Incorrect format. \nPlease try again with the format below: \n"
+                        + type + " ($task_name) " + delimiter + " ($date/day)");
+            }
```
bjhoohaha: Very considerate developer in telling user to input the correct format!

src/main/java/duke/command/DeleteCommand.java
bjhoohaha: Adding on to @niqiukun the `IndexOutOfBoundsException` may then be handled in `tasks.delete(index)`

src/main/java/duke/command/AddCommand.java
bjhoohaha: You may also want to consider extending different errors from `DukeException`. This would help make the code more readable and minimize the many '+' when trying to build your string. you can simply create a new `DukeException` and passing the arguments `$(task_name)` and `$(date/day`) as parameters

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Executes the Delete Command and removes the task from the Linked List and file.
+     *
+     * @param tasks   The TaskList containing all existing tasks.
+     * @param ui      The Ui for printing purposes.
+     * @param storage The Storage for saving tasks to file.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+
+        StringBuilder sb = new StringBuilder("Noted. I've removed this task:\n" + tasks.allTasks.get(index));
```
bjhoohaha: Good use of String Buffer to increase efficiency rather than building a new string each time.

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,37 @@
...
+     * @param tasks   The TaskList containing all existing tasks.
+     * @param ui      The Ui for printing purposes.
+     * @param storage The Storage for saving tasks to file.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+
+        // Print all existing items in the list
+        ui.printMessage("Here are the tasks in your list:");
+        tasks.allTasks.forEach(x -> ui.printMessage((tasks.allTasks.indexOf(x) + 1) + ". " + x));
+    }
```
bjhoohaha, line 27: Nice use of `Stream` here

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+            case "deadline":
+            case "event":
+                return new AddCommand(details[0], fullCommand.split(details[0]));
+            case "bye":
+                return new ExitCommand();
+            default:
+                throw new DukeException("I'm sorry, but I don't know what that means :-( ");
+            }
+        } catch (ArrayIndexOutOfBoundsException e) {
+            // If the command misses a description entirely
```
bjhoohaha, line 51: You may want to consider handling the exceptions at each case. Having a try catch for "done" and "delete" thus minimise the switch case you have to do here.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+            writer.close();
+            reader.close();
+
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        tempFile.renameTo(inputFile);
+    }
+}
```
bjhoohaha, line 139: Interesting alternative implementation in deleting the line from the database as well thus data is saved even though the program exits!

src/test/java/duke/command/AddCommandTest.java
```diff
@@ -0,0 +1,108 @@
...
+                    ((AddCommand) Parser.parse("todo Make the bed")).parseDescription());
+        } catch (DukeException e) {
+            System.err.println(e);
+        }
+    }
+
+    @Test
+    public void parseDescription_emptyToDo_exceptionThrown() {
+        try {
+            ((AddCommand) Parser.parse("todo")).parseDescription();
```
bjhoohaha: You may want to consider doing `Assertion.assertThrows( )` This would minimize having each try catch for each method test

src/test/java/duke/command/AddCommandTest.java
```diff
@@ -0,0 +1,108 @@
...
+        } catch (DukeException e) {
```
bjhoohaha: You may want to consider using a fail() method. This would ensure that when fail() is reached the test would have failed
```suggestion
 public void parseDescription_emptyToDo_exceptionThrown() {
	        try {
	            ((AddCommand) Parser.parse("todo")).parseDescription();
	            fail() // fail method here which will give an error by Junit if this line is reached. 
	            // This line should not be reached because previous line would throw the exception
```

bjhoohaha (commented):
Good Job! This is generally well done. Alternative implementation of storing the task each time it is changed, hence data is saved even though program exits.

Developer is also very considerate for user by throwing an error as well as telling user how to enter the correct input.

However, developer may want to consider having more abstraction to increase code readability as well as making it more OOP. Duke project is hard and takes a lot of time, so jiayou and I look forward to seeing the final version of this product as well! All the best to you.

PR #256, opened 2019-08-29
[Sakshi Pradyumn] Duke Increments

SakshiPradyumn:
All levels Week 3


src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+    private TaskList taskList;
+    private Ui ui;
+
+    public Parser(Storage storage, TaskList taskList, Ui ui) {
+        this.storage = storage;
+        this.taskList = taskList;
+        this.ui = ui;
+    }
+
+    public void parse() {
```
dragontho: Very intuitive way of handling exceptions! I see you are going for the "for each command handle it own exceptions" approach. But I think it can get a bit cluttered the more try/catch blocks you add. Maybe I could suggest trying the whole block of commands and catching them at the end?

src/main/java/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+        File f = new File(fPath); // create a File for the given file path
+        Scanner S = new Scanner(f); // create a Scanner using the File as the source
+        while (S.hasNext()) {
+            String temp = S.nextLine();
+            cl.readFromFile(temp);
+        }
+    }
+*/
+
+    public List<Task> load() throws FileNotFoundException, DukeException {
```
dragontho, line 45: Very clean way of loading the text file! Could learn a thing or two from this.

src/main/java/Ui.java
```diff
@@ -0,0 +1,51 @@
+public class Ui {
+    private String response;
+
+    private void getResponseDirect() {
+        System.out.println("    ____________________________________________________________\n     "
```
dragontho: Nice way of abstracting the annoying borders! I've seen that you rewrite the same borders for all of your exceptions as well. Maybe you could do the same form of abstraction to them as well?

src/main/java/Ui.java
```diff
@@ -0,0 +1,51 @@
+public class Ui {
+    private String response;
```
dragontho, line 5: I like how you generalise every single command into a single response line then call getResponseDirect() at the end of your UI calls. Very nice!

dragontho (commented):
Overall, code was written well and organised. A little more abstraction could be used but otherwise fantastic optimisation of code!

atharvjoshi (commented):
Hello! Good job on making it to Week 4 of the iP hehe :)

- I like your thorough unit testing. I hope to practice such rigour in my unit testing as well. 

- It is interesting that you classify anything task related (e.g. creation, conversion to string) into the TaskList class. This gave me a new perspective of looking at what a 'TaskList' classification can achieve. I myself performed Task Generation and conversion of tasks in the list to one long string for printing both in my Parser class.

- I could not figure out why the classes ChatLike and Ui are co-existing in your code. Perhaps I am not thinking hard enough - aren't both performing very similar functions?

- Would like to suggest practicing a little more encapsulation. For instance, in your Parser's parse method, you could have separated the 'parsing' i.e. process of converting human input into a task understandable by Duke from the 'command execution' i.e. the actual subtasks to be done by the parsed task. This could increase the readability of the parse method (it took me the longest to read and understand). In the same method, the exception handling could then be delegated to the 'command execution' handler, while the parser handler is only responsible for 'parsing' i.e. understanding the command (and not executing it). 

- I used enums to enforce type safety on the kinds of commands Duke can understand and I used a class for each command (that inherits a 'Command' class) such that each command defines an abstract 'execute' method to run itself. I found that very useful because I too earlier did string slicing and matching (as you have done in your parse() method) and found that a little tedious compared to simply instantiating a new command based on the first word entered by the user. 

All the best for the coming weeks!

SakshiPradyumn:
Hello @dragontho! I really liked the way you have put similar classes in
folders like tasks, commands, ....

Thank you for your observations.

I get your point that my code could've been neater. I will try that in next
week's submission.

Your Ui class deserves a thumbs-up!

My Ui and Parser looks quite same, and I should work on deleting extra
lines.

Great work on JavaDoc; made me realise their importance now.

Well, something I have been asking people but with no success: How do you
delete a commit?
For initial commits I used to make mistakes, but somehow there were no
issues.

Also, apart from your beautiful code, can you suggest some other
classmate's code which has a unique way of implementation?

See ya!

SakshiPradyumn:
Hello @atharvjoshi! I really liked the way you have put similar classes in
folders like tasks, commands, ....

Thank you for your observations.

I will try more encapsulation in next week's submission.

Your enum imperative is wonderfully effective! I will try to incorporate
enum in my next project
Sorry for the time you wasted on my Parser. I did the bare minimum right
now because of my other commitments. Will work on it sometime.

Inspiring work on JavaDoc as well!

Also, apart from your beautiful code, can you suggest some other
classmate's code which has a unique way of implementation?

See ya!

PR #257, opened 2019-08-29
[Wang Tianyi] Duke Increments

arjunwangty:
All levels Week 3


src/main/java/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+     * Creates a new parser instance.
+     *
+     * @param s Storage instance to be used.
+     * @param t TaskList containing task list.
+     * @param u Ui instance to be used.
+     */
+    public Parser(Storage s, TaskList t, Ui u) {
+        this.storage = s;
+        this.tasks = t;
+        this.ui = u;
```
waifonglee: i think there is no need to use "this" in this situation since we can easily distinguish between 'ui' and 'u'! Same goes for 'storage' and 's', and 'tasks' and 't' :)

src/main/java/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+                    try {
+                        Task task = tasks.done(Integer.valueOf(command.substring(5)));
+                        ui.done(task);
+                        try {
+                            storage.writeToFile(tasks.generateInfo());
+                        } catch (IOException e) {
+                            System.out.println("Something went wrong: " + e.getMessage());
+                        }
+                    } catch (Exception e) {
+                        System.out.println("    ____________________________________________________________\n     " +
```
waifonglee: i believe we should put our line break before an operator :)

src/main/java/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+    private List<Task> tasks = new ArrayList<Task>();
+    private String filePath;
+
+    /**
+     * Creates a new Storage instance with the given path to the file.
+     *
+     * @param path path to the file.
+     */
+    public Storage(String path) {
+        this.filePath = path;
```
waifonglee: You can omit "this" ! :)

waifonglee (commented):
Looks good! I really like how thorough the test classes are!

ghost:
Hi Tianyi, good job with sticking to the coding standard very closely!
Your javadocs are very well written and informative, which is something I hope to pick up from you :)

One suggestion I would like to make is:
https://github.com/nus-cs2103-AY1920S1/duke/pull/257/files#diff-ea273e827996a4f97f92dbac061b7b5cR41

It would be cleaner if written something like:
```
String[] commands = command.split(" ");
if (commands[0].equals("done")) {
```

Furthermore, the code will look more organized too :) For example,
```
if (commands[0].equals("done")) {
	...
} else if (commands[0].equals("todo")) {
	...
} else {
	System.out.println("OOPS!!! I don't know what that means :-(");
}
```

Good job with splitting the code into classes to make it more OOP!
I have also tested the code and the saving and loading of tasks works!
All JUnit tests are well written too.
Keep up the good work :)

PR #258, opened 2019-08-29
[Ryan Tay] Duke Increments

ryantay232:


src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Constructor for event.
+     *
+     * @param description description of event
+     * @param date        date and time of event
+     */
+    public Deadline(String description, String date) {
+        super(description);
+        String[] dateSplit = date.split("\\s");
```
blimyj: The use of LocalDateTime.parse could really help to simplify the code here!

blimyj: I think the use of LocalDateTime.parse & DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm") could really help to simplify your code here!

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,60 @@
...
+        return dateTime.getDayOfMonth() + "/" + dateTime.getMonthValue() + "/" + dateTime.getYear() +
+                " " + String.format("%02d", dateTime.getHour()) + String.format("%02d", dateTime.getMinute());
+    }
+
+    /**
+     * Returns String in text file format.
+     *
+     * @return a String to write on text file
+     */
+    public String toFile() {
```
blimyj, line 34: The name "toFile" sounds a bit confusing to me, perhaps you can consider renaming this method to "toSaveFileFormat"?

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param input  original String input from user
+     * @return ArrayList of tasks
+     */
+    public List<Task> doCommand(int result, String input) {
+        String[] s = input.split("\\s");
+        switch (result) {
+            case 0:
+                return tasks;
+            case 1:
+                System.out.print(Ui.line);
```
blimyj: There seems to be a lot of repeated use of Ui.indent and Ui.line, perhaps you could abstract this out to a separate formatter method?

blimyj (commented):
Your code was really easy to understand for me! Though I feel that the "toFile" method could be renamed to slightly improve clarity. You could also perhaps consider abstracting out the formatting of your output so that you need not repeatedly call Ui.indent. Great work!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Gets a command from the user.
+     *
+     * @param input original String input from user
+     * @return an integer that represents a command
+     */
+    public int getCommand(String input) {
+        String[] i = input.split("\\s");
+        if (i[0].equals("bye")) {
```
moziliar: Have you considered using a switch statement? In a sense, they are equivalent to each other. But switch is used for long if-else statement where the conditions are similar.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,38 @@
...
+            return 0;
```
moziliar: Over here, number does not carry any meaning to a developer first time looking at your code. It would be better to have an Enum class or final static variables to grant these integer meaningful names.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class Storage {
+    /**
+     * Charset for reading and writing .txt files.
+     */
+    final static Charset ENCODING = StandardCharsets.UTF_8;
+    /**
+     * Path for data.
+     */
+    private Path path;
```
moziliar: Do we need to Javadoc private attributes? They are not supposed to be visible to the outside world but should stay local in the class. Javadoc will reveal the actual implementation of this class, which breaks abstraction somewhat.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+     * Reads text file and formatted into ArrayList of tasks to be used on other classes.
+     *
+     * @return ArrayList of tasks
+     * @throws IOException
+     */
+    public ArrayList<Task> readDuke() throws IOException {
+        ArrayList<Task> tasks = new ArrayList<Task>();
+        List<String> lines = read();
+        for (String s : lines) {
+            String[] split = s.split("\\|");
```
moziliar, line 78: Could it have a more meaningful name like splitInput or splitString? "split" itself is a verb which does not serve well as a variable name meant for an array of strings.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+        Files.write(path, lines, ENCODING);
+    }
+
+    /**
+     * Reads text file and formatted into ArrayList of tasks to be used on other classes.
+     *
+     * @return ArrayList of tasks
+     * @throws IOException
+     */
+    public ArrayList<Task> readDuke() throws IOException {
```
moziliar, line 74: Should Storage deal with the creation of individual tasks? It could be a job for Parser, since it is created to parse text anyways.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+        return Files.readAllLines(path, ENCODING);
+    }
+
+    /**
+     * Writes text file to path.
+     *
+     * @param lines text lines to write onto .txt as List of Strings
+     * @throws IOException
+     */
+    private void write(List<String> lines) throws IOException {
```
moziliar, line 64: If the action is one-liner, could it be just written inside writeDuke? I see some clean code pattern here nonetheless.

ryantay232:
test 
https://github.com/ryantay232/duke/blob/0126566726ed12c511f45f06642fa0b95e05049c/src/main/java/duke/Duke.java#L47

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Gets a response from user and sends back the output.
+     *
+     * @param input input String from user
+     * @return output String from duke
+     */
+    public String getResponse(String input) {
+        Command command = new Command(parser.getCommand(input), input);
```
moziliar, line 46: It seems that parser.getCommand(input) outputs a Enum value of Comd. My approach here is to get Parser to parse the input and output the appropriate Command. Something seems off with my approach as I think Parser should not know about Command. But over here you have 2 input variables inside your Command constructor. Looks repeated. What's your view on this?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,48 @@
...
+public class Command {
+    private Comd comd;
+    private String input;
+    private String output;
+    private ArrayList<Task> taskList;
+
+    public Command() {
+    }
+
+    public Command(Comd comd, String input) {
```
moziliar, line 20: How about using subclassing to represent the different kind of Commands? You could capitalise on polymorphism.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,48 @@
...
+import duke.parser.Comd;
+
+/**
+ * A class for storing information for commands.
+ */
+public class Command {
+    private Comd comd;
+    private String input;
+    private String output;
+    private ArrayList<Task> taskList;
```
moziliar, line 15: Should Command class know about taskList?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,39 @@
...
+    public Parser() {
+    }
+
+    /**
+     * Gets a command from the user.
+     *
+     * @param input original String input from user
+     * @return an integer that represents a command
+     */
+    public Comd getCommand(String input) {
```
moziliar, line 16: Could it just output a Command?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Class for deadline tasks.
+ */
+public class Deadline extends Task {
+    private LocalDateTime dateTime;
+    private DateTimeFormatter formatter;
```
moziliar, line 11: Would it be better to have formatter inside Parser as a static final constant?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Constructor for event.
+     *
+     * @param description description of event
+     * @param date        date and time of event in String format
+     */
+    public Event(String description, String date) {
+        super(description);
+        this.formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
```
moziliar, line 24: Would it be better to have the pattern as a static final constant in Parser?

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,191 @@
...
+     * Constructor for TaskList.
+     *
+     * @param tasks ArrayList of tasks
+     */
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * Carry out operations according to commands from user.
```
moziliar: Command pattern mandates an execute method to be overriden by the subclasses (implementation). User would just need to use `command.execute()` to execute the command instead of making use of task list to do it (It is not task list's job to execute the task. It merely stores the tasks)

moziliar (commented):
Good to have segmentation inside TaskList. Would it be better to create subclasses for command to execute the operations?

PR #259, opened 2019-08-29
[Timothy Yu Zhiwen] Duke Increments

Kyzure:
All levels Week 3.


src/test/java/duketest/EventTest.java
```diff
@@ -0,0 +1,80 @@
...
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class EventTest {
+
+    private static final String TICK = "✓";     // \u2713
+    private static final String CROSS = "✗";    // \u2717
+
+    @Test
+    public void instanceTest() {
+        // Test for Date, Time and Both
```
EvonDong: It would be better if you separate them into different test methods, so that you can detect where the bug lies. Apart from that, everything else is good!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * Prints to user guidance in case they do not know what to do.
+     * 
+     * <p>This is the user guide for the user. 
+     * It indicates the user what to do and what is the format of the inputs.
+     * It takes in as many user Inputs as possible to output the indicated help required
+     * E.g. User Inputs 1 | Prints out help for 'List' Command
+     */
+    public void printHelp() {
```
EvonDong: I like how you added this method to guide the user. This was nicely done.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,175 @@
+/**
+ * Represents a Parse Manager, which parses the inputs given by the user, so our other classes
+ * are able to read it with ease.
+ */
+class Parser {
+    /**
```
EvonDong: Your comments are very specific and detailed. Well done.

src/main/java/duke/Action.java
```diff
@@ -0,0 +1,7 @@
+/**
+ * Represents the different Actions a user can do. 
+ * This enum is also the first word which the user inputs
+ */
+enum Action {
+    LIST, EXIT, HELP, TODO, DEADLINE, EVENT, DONE, DELETE, FIND;
+}
```
cutieprobe: Good use of `Enum` to represent the actions!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Represents a ArrayList which stores task and implemenets Serializable, allowing it to be
+ * serialized by the Storage class.
+ * 
+ * @see Storage#Storage(String)
+ */
+class TaskList implements Serializable {
```
cutieprobe: Great usage of serialization to simplify the storage process!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a deadline, with a given date/time/both 
+ * which indicates when the user needs to complete it by.
+ */
+class Deadline extends Task {
+    private LocalDate date;
+    private LocalTime time;
```
cutieprobe: Is it better to use `java.util.Date` to storage the date/time together in one field?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,68 @@
...
+                try {
+                    this.time = LocalTime.parse(dateTimeArr[0], TIME_FORMATTER);
+                    this.date = null;
+                } catch (DateTimeParseException error) {
+                    throw new DukeException("Please enter a legitamate date, time or both");
+                } catch (Exception error) {
+                    throw new DukeException("The format entered is wrong. Use \'help\' "
+                            + "for formatting styles");
+                }
+            }
```
cutieprobe: I'm not sure about the nested try-catch block. For example, if exceptions occur when parsing the `date`, but no error occur in parsing the `time`, then will `date` be set to null and no exception is thrown? Maybe better to refactor it into a single try-catch block?

src/main/java/duke/Deadline.java
Kyzure: Good idea, but I use them separately as I wanted more leeway for the user. They can type both date, time or both instead of just date time, unless java.util.Date is able to do that.

src/main/java/duke/Action.java
cutieprobe: Good use of `Enum` to represent the actions!

cutieprobe (commented):
Overall, great work!

PR #260, opened 2019-08-29
[Irene] Duke Increments

irene-lly:


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+     * @throws DukeException If task is empty.
+     * @throws ParseException If date or time of task does not follow the format.
+     */
+    public Task parse() throws DukeException, ParseException {
+        Task parsedTask = null;
+        String[] wordArray = str.split(" ");
+        String taskWithoutType = str.replace(wordArray[0], "").trim();
+
+        SimpleDateFormat format = new SimpleDateFormat("dd/MM/yyyy HHmm");
+        SimpleDateFormat formatter = new SimpleDateFormat("dd MMM yyyy, hh:mm a");
```
geshuming: This is slightly confusing. Why are there 2 different types of format here? 

`formatter` is quite vague. Maybe we can use `dateTimeFormat` instead? Hence, `dateTimeFormat.format(str)` is more obvious

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,49 @@
...
+    public Duke(String filePath) throws IOException {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.loadTasks());
+        } catch (DukeException e) {
+            tasks = new TaskList();
+        } catch (FileNotFoundException e) {
+            File file = new File(filePath);
+            file.createNewFile();
```
geshuming: `tasks` is potentially null here. 

Maybe you can use the `finally` block to check if `tasks` is null, and create a new `TaskList` there.

src/main/java/extra.java
```diff
@@ -0,0 +1,164 @@
+/*import java.io.FileNotFoundException;
```
geshuming: Can we try not to include dead code into the master branch? It makes it confusing as to why such code exists.

geshuming (commented):
Overall I like the adherence to the style guide.

src/main/java/TaskList.java
```diff
@@ -17,6 +20,8 @@ public void addToList(String input) {
         String[] arrOfWords = input.split(" ");
         String taskWithoutType = input.replace(arrOfWords[0], "").trim();
         Task task = new Task(input);
+        SimpleDateFormat format = new SimpleDateFormat("dd/MM/yyyy HHmm");
+        SimpleDateFormat formatter = new SimpleDateFormat("dd MMM yyyy, hh:mm a");
```
mirozo: Remove the unwanted one

mirozo (commented):
Interesting how you used the TaskList class to handle user input. May I know why you processed user inputs overhere? In my opinion, I don't recommend it, because TaskList should only support behaviors related to a List. Can be quite confusing for other developers to understand your code.

PR #261, opened 2019-08-29
[Lim Hui Min] Duke Increments

huiminlim:


src/main/java/duke/Duke.java
```diff
@@ -59,21 +60,21 @@ String getDukeResponse(String input) {
         try {
             Command c = Parser.parse(input);
             output = c.execute(tasks, ui, storage);
-        } catch (IndexOutOfBoundsException e){
```
teika97: Comprehensive error handling. However, perhaps you could consider handling these errors in DukeException?

src/main/java/duke/parser/Parser.java
```diff
@@ -14,6 +14,7 @@
 import duke.task.Task;
 import duke.task.ToDo;
 import duke.ui.Ui;
+
 import java.util.ArrayList;
```
teika97: Since the commandList is always the same, perhaps you could initialize the commandList in the constructor or in the attribute directly without having to create a separate initialize() method.

src/main/java/duke/parser/Parser.java
```diff
@@ -71,7 +72,7 @@ private static String getCommandName(String userInput){
     }
```
teika97: I think its great your code accepts user input which is not case-sensitive. As a minor improvement, perhaps you can use commandName as a variable name given your method name is getCommandName.

src/main/java/duke/parser/Parser.java
```diff
@@ -117,7 +118,7 @@ public static Command parse(String input) throws CommandNotFoundException, Incor
 
             } else {
```
teika97: Perhaps it would be clearer if you specify the last case e.g. else if (commandName.equals("EVENT"))

teika97: For creation of new tasks to be added into taskList, perhaps it would be clearer and more readable if you declare variables for each of the parameters e.g. String description = input.substring("Todo".length()).trim() or boolean isDone = false

teika97 (commented):
Hi Hui Min,
My team and I reviewed some parts of your code. We feel that your code is well-written with effort put into abstracting methods e.g. Parser with good handling of different types of errors. Good job! :)

src/main/java/duke/parser/Parser.java
huiminlim: Thank you for the feedback, I appreciate them greatly! I will look into the feedback and implement the suggestions. Cheers!

docs/README.md
```diff
@@ -1,20 +1,300 @@
...
+Alternatively, a command can be issued in the command terminal: `java -jar duke-0.2.0.jar`.
+
+
 ## Features 
 
-### Feature 1 
-Description of feature.
+Duke allows you to track three types of task - To Do, Deadline and Events. In addition, there are also various utility functions to allow you to manage your task list. These functions include `list`, `delete`, `done`, `bye`, `help`, `find`.
+
+For all To Do, Deadline and Events, a task description is shown in chronological order in which they were added. The classification of the To Do, Deadline and Events are denoted by a task icon, as shown in table below.
```
nattanyz, line 41: a thorough readme! however, the language could be clearer and less verbose in some parts. :) some minor typos e.g. line 64: "shown to the you".

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,90 @@
...
+	
+	/**
+	 * Performs Duke start up operation to generate task list.
+	 */
+	public void performDukeStartup() throws IncorrectFileFormatException, FileNotFoundException,
+											IncorrectArgumentsException {
+		tasks = new TaskList();
+		try {
+			tasks = new TaskList(storage.load(ui));
+		} catch (IncorrectFileFormatException j) {
```
nattanyz, line 43: perhaps the exceptions pertaining to the file to be loaded can be resolved in Storage class instead?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,90 @@
...
+	 * Generate a GUI dialog box with duke response and image.
+	 *
+	 * @return Dialog box with duke response and image.
+	 */
+	public String getDukeResponse(String input) {
+		String output;
+		try {
+			Command c = Parser.parse(input);
+			output = c.execute(tasks, ui, storage);
+		} catch (IndexOutOfBoundsException e) {
```
nattanyz, line 66: would it be helpful to abstract out exceptions into a DukeException class? defining the output strings here in Duke.java makes sense, but might be hard to find and edit.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,64 @@
...
+	private CommandType commandType;
+	private Task pendingTask;
+	
+	/**
+	 * Constructor for class Duke.
+	 *
+	 * @param command     String command input from user.
+	 * @param pendingTask Task object from information given by user.
+	 */
+	public Command(String command, Task pendingTask) {
```
nattanyz, line 26: it seems like not all commands pertain to a particular task (e.g. bye, help), so would the code be more self-explanatory if pendingTask is removed from the method signature?

src/main/java/duke/command/CommandType.java
```diff
@@ -0,0 +1,8 @@
+package duke.command;
+
+/**
+ * Contains the different types of commands available in duke.
+ */
+public enum CommandType {
```
nattanyz, line 6: abstracting out command types is a good idea! :) this helps to reduce errors in parsing and extending the app's functionality.

src/main/java/duke/command/HelpCommand.java
```diff
@@ -0,0 +1,75 @@
...
+		String helpOutput = null;
+		if (isCommandSpecific) {
+			helpOutput = executeSpecificCommand(inputCommand);
+		} else {
+			helpOutput = executeGeneral();
+		}
+		return helpOutput;
+	}
+	
+	private String executeSpecificCommand(String commandInterested) {
```
nattanyz, line 30: what is the difference between the execute() and executeSpecificCommand() methods? is it possible to eliminate one of those, or improve the abstraction of helper methods for execute()?

src/main/java/duke/frontend/MainWindow.java
```diff
@@ -0,0 +1,111 @@
...
+	
+	/**
+	 * Set up Duke for main GUI, prepare the task list for first load.
+	 */
+	public void setDuke(Duke d) {
+		duke = d;
+		String loadingErrorMessage = "";
+		boolean isErrorOccurred = false;
+		
+		// Perform loading of tasks from disk
```
nattanyz, line 50: is there a way you could avoid loading tasks from disk within setDuke()? it seems like this chunk of code may be repeated from Duke.java.

src/main/java/duke/frontend/MainWindow.java
```diff
@@ -0,0 +1,111 @@
...
+		
+		if (isErrorOccurred) {
+			dialogContainer.getChildren().add(new DialogBox(loadingErrorMessage, dukeImage));
+		}
+		
+		// Throw initial duke greeting
+		dialogContainer.getChildren().add(new DialogBox(duke.getDukeWelcome(), dukeImage));
+	}
+	
+	/**
```
nattanyz, line 74: javadoc is not updated here

src/main/java/duke/frontend/MainWindow.java
```diff
@@ -0,0 +1,111 @@
...
+	 * Creates two dialog boxes, one echoing user input and the other containing Duke's reply and then appends them to
+	 * the dialog container. Clears the user input after processing.
+	 */
+	@FXML
+	private void handleUserInput() {
+		String input = userInput.getText();
+		
+		boolean isHelpCommand = input.toLowerCase().startsWith("help");
```
nattanyz, line 82: why do help commands require their own method for handling? could these operations be combined into the same method, or called from the same method? furthermore,  is MainWindow within the GUI implementation the appropriate place to handle user input parsing and handling? is this logic repeated from the CLI implementation duke?

nattanyz (commented):
hi Hui Min, great effort overall! :) I'm reviewing your code because I made some reference to it when I was just setting up my GUI, so thanks for that! :p hope you find some of these comments helpful. I'll add more in future if I have time!

PR #262, opened 2019-08-29
[Tiu Wee Han] Duke Increments

tiuweehan:


woon17 (commented):
Good OOP design. I like your JUnit testing parts. They are quite detail and useful to test the different behaviors and functions. Using Enumerations is a good idea, I should learn from you to make a class for it.

SoilingRogue (commented):
I think you have done a really great job, there were definitely several takeaways when I reviewed your code. I really liked your use of enums as well as the proper javadoc you did. Furthermore, I really appreciate the effort you put into Unit Testing, especially the use of stubs for testing specific classes. The class packaging, organizations and coding standards were all great.

There were also a few points which I felt could be improved on. Firstly, the abstraction barrier.  You should consider using relevant classes to handle assigned commands, such as TaskList class to handle the finding and returning of tasks containing the keyword, instead of directly using FindCommand class to handle it.

Next, you should use your execute method in the Command classes to process commands correctly, throw and handle DukeExceptions, such as checking of valid task list sizes, such that other classes would not have to, such as methods in TaskList class.

Lastly, your public static void main(String args[]) method is empty in the DukeCli class/ double main methods clashing with the Launcher class, not sure if you noticed that :)

Great job overall!

src/main/java/duke/command/BasicCommand.java
```diff
@@ -0,0 +1,73 @@
+package duke.command;
+
+import duke.task.TaskList;
```
jiayushe: Try to arrange your imports in alphabetical order.

src/main/java/duke/Launcher.java
```diff
@@ -0,0 +1,29 @@
...
+import duke.program.DukeGui;
+import javafx.application.Application;
+
+public class Launcher {
+    /**
+     * The entry point to the Duke CLI program.
+     *
+     * @param filepath The location of the storage file.
+     */
+    public static void launchCli(String filepath) {
```
jiayushe: I think since we are using RCS, it is preferred to remove dead code from the codebase.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Executes the command.
+     *
+     * @param tasks The list of tasks the command should act on.
+     * @param ui The ui that the result of the command should be directed to.
+     * @param storage The Storage that the command should act on.
+     * @throws DukeException If command fails to execute to completion.
+     */
+    void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
```
jiayushe: I feel that instead of overloading the `execute` method here in the `Command` class, you can handle the alternative ui output in the main class.

jiayushe (commented):
Generally well designed code and good test coverage. I left some comments regarding some minor issues on coding style and design below. Otherwise look good to me!

PR #263, opened 2019-08-29
[Denis] Duke Increments

ghost:


waifonglee (commented):
Nicely structured code that follows coding standard :). Good job packaging code to make it more structured.

src/main/java/weomucat/duke/Controller.java
```diff
@@ -0,0 +1,211 @@
...
+        for (ByeCommandListener listener : byeCommandListeners) {
+          listener.byeCommandUpdate();
+        }
+      }
+    });
+  }
+
+  /**
+   * Add a AddTaskCommandListener to the Controller. When a AddTaskCommand is received, this
+   * listener will be notified.
```
arjunwangty: Javadoc is detailed and informative 👍One minor suggestion is to separate the first sentence as a summary of the method and start the sentence with an action word in the form "Adds", "Parses, Performs" etc. :)

arjunwangty (commented):
Hi Denis, I really like the rigour of your code. Your classes are very well encapsulated, commands classes and listeners classes are also extracted out. I think I can learn how to write such well-designed codes :) I can see that your project is also properly packaged and built with Gradle. Well done!

src/main/java/weomucat/duke/Controller.java
ghost: Hi Tianyi, thank you for the kind words :) I agree with the javadocs, I will take your advice and put a summary with an action word above my javadocs! 👍

PR #264, opened 2019-08-29
[Kelvin Harris] Duke Increments

kelvinnharris:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,68 @@
...
+        ui.printWelcome();
+
+        // process input
+        String text = ui.readLine();
+        while (!text.equals("bye")) {
+            try {
+                parser.processLine(text); // add, delete, etc
+            } catch (DukeException e) {
+                System.err.println(e.getMessage());
+            } catch (NumberFormatException e) {
```
iltep64: Would it be better to wrap all the exceptions at the source with a `DukeException`, so that the catch block only needs to be concerned with one particular type only?

src/main/java/duke/command/Storage.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param tasks The tasks to be saved after running duke.
+     */
+    public void save(TaskList tasks) {
+        try {
+            ArrayList<Task> list = tasks.list;
+            String filePath = this.filePath;
+            FileWriter fw = new FileWriter(filePath);
+            StringBuilder sb = new StringBuilder();
+            for (Task task : list) {
+                sb.append(task.getFileStringFormat()).append("\r\n");
```
iltep64: If you want the native OS line endings, Java has `System.lineSeparator()`. If we're going for cross-platform compatibility, using raw `\n` is better.

src/main/java/duke/command/Ui.java
```diff
@@ -0,0 +1,90 @@
...
+     * Prints each task in the list provided.
+     *
+     * @param list The list to be printed.
+     */
+    public void printList(ArrayList<Task> list) {
+        if (list.isEmpty()) {
+            System.out.println("No task found");
+        } else {
+            System.out.println("Here are the tasks in your list:");
+            IntStream.rangeClosed(1, list.size()).forEach(x -> {
```
iltep64: Nice use of Java Streams here.

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+            ArrayList<Task> listFound = new ArrayList<>();
+            tasks.list.stream()
+                    .filter(task -> task.getDescription().contains(secondWord))
+                    .forEach(listFound::add);
+            ui.printList(listFound);
+        } else {
+            Task task;
+            if (firstWord.equals("todo")) {
+                if (words.length < 2 || words[1].equals("")) {
+                    throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
```
iltep64: Maybe the `"☹ OOPS!!!"` part can be added in the `DukeException` constructor, or the final `catch (DukeException e)` handler, instead of spreading it across all places you create a `DukeException`.

src/main/java/duke/Duke.java
kelvinnharris: Oh! Thanks for the suggestion. I was wondering why we needed `DukeException` class and could not figure out its usage.

src/main/java/duke/command/Storage.java
kelvinnharris: Apparently `\n` did not produce a new line for me. Do you think it was because I was running it on Windows Command Prompt?

src/main/java/duke/command/Ui.java
kelvinnharris: Thanks!

src/main/java/duke/command/Parser.java
kelvinnharris: Noted. Thank you!

src/main/java/duke/command/Parser.java
```diff
@@ -65,41 +65,41 @@ public void processLine(String userInput) throws DukeException, ParseException {
             }
             int index = Integer.parseInt(words[1]) - 1;
             Task removed = tasks.delete(index);
-            ui.printDeleteTask(removed, list);
+            return ui.printDeleteTask(removed, list);
 
         } else if (firstWord.equals("find")) {
             String secondWord = words[1];
             ArrayList<Task> listFound = new ArrayList<>();
             tasks.list.stream()
```
nishcafe: Good use of streams here!

src/main/java/duke/command/Parser.java
```diff
@@ -65,41 +65,41 @@ public void processLine(String userInput) throws DukeException, ParseException {
...
-                    throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+                    throw new DukeException("OOPS!!! The description of a todo cannot be empty.");
                 }
                 task = new ToDo(userInput.split(" ", 2)[1]);
             } else if (firstWord.equals("deadline")) {
                 if (words.length < 2 || words[1].equals("")) {
-                    throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
+                    throw new DukeException("OOPS!!! The description of a deadline cannot be empty.");
                 }
                 String description = userInput.split(" ", 2)[1].split(" /", 2)[0];
```
nishcafe: Maybe you could process each task in a different function for easier readability

PR #265, opened 2019-08-29
[Nadia] Duke Increments

wpuffs:


C-likethis123:
Hi Nadia, 

Overall, I like that your code is clean, with a lot of comments, and that there were many tests. 

I do have a question regarding your code. In Client.java, there is a lot of code to format how the GUI looks like. Would it be easier to create FXML files and controller classes for the FXML files? In this way the FXML files contains all the formatting code, and the controller classes contain code for the application logic. 

I look forward to your replies. 

-Jia Ying

build.gradle
```diff
@@ -0,0 +1,37 @@
+plugins {
+    id 'java'
+    id 'application'
+	id 'checkstyle'
```
jxofficial, line 4: There seems to be extra spaces here

jxofficial (commented):
Hi Nadia,

Your code was very pleasant to read. It is well structured, with each class having minimal dependency on other classes. I especially liked how your `UI` class is extremely simplified, just dealing with printing of `String` However, your `UI` `show` method seems to be unused in this version. Additionally, you seem to be using a lot of string concatenation with `+=`, and perhaps this will not be advisable in terms of time complexity if your application were to scale up in the future (refer to attached picture). But of course, it is not a concern now. Correct me if I'm wrong on any points above!

Also, I would just like to ask your rationale behind using `if else` statements instead of `switch` for your `Parser` class.

Looking forward to your reply.

Jun Xian

image

PR #266, opened 2019-08-29
[Ng Jun Hao] Duke Increments

jun-ha0:
Duke Level-1 to Level-10.


ning-sy210:
### Code
I really like how the functionality of your code was separated into many smaller classes such that each of them has a very apparent role in making your program work the way it does as a whole! 😊
Especially, I like how the different types of commands are being separated into their respective class files, and how `Message.java` has a class of its own to handle the different error messages shown to the user!

However, I do have a few concerns on certain aspects of your code:

**1. Certain `execute()` method arguments not used**

    public abstract void execute(TaskList taskList, UserInterface ui, Storage storage) throws DukeException;

In `Command.java`, the above abstract method is defined to accept a `TaskList`, `UserInterface` and `Storage` as arguments. 



    @Override
    public void execute(TaskList taskList, UserInterface ui, Storage storage) throws DukeException {
        ui.exitProgram();
    }

However, in `ExitCommand.java`, only the `UserInterface` is actually used. Such cases of irrelevant parameters can also be seen in `ListCommand.java` and `FindCommand.java`, where not all of the method parameters are used. 
Here is the problem I see with this approach: If in the future there are new commands to be added that requires new parameters (that are not `TaskList`, `UserInterface` or `Storage`), all existing `Command` subclasses will need to have their method definition changed just from the introduction of a single command. And as you can already tell, this will get very tedious as the number of commands reaches a large number.

**2. Inappropriate names for certain variables**

`
private LocalDateTime by;
 `

`
private LocalDateTime at;
`

As per the Java Coding Standard, try to use more descriptive words for these variables in `Deadline.java` and `Event.java` respectively. This is to help others that read your code! One suggestion could be `DateTimeOfDeadline` and `DateTimeOfEvent`.




    public void showAddition(Task task) {
        showToUser(Message.MESSAGE_ADDED);
        showToUser(" " + task);
    }

    public void showDeletion(Task task) {
        showToUser(Message.MESSAGE_DELETED);
        showToUser(" " + task);
    }


Similarly, for the two method names above, it would be more ideal if they were to be renamed respectively to be `showTaskAddedMessage()` and `showTaskDeletedMessage()` instead.

**3. There can be more variations of Duke Exceptions.**
Try to introduce more variations for Duke Exceptions. This is because errors are more easily identifiable by their causes rather than the messages they show. So I would think it is a good practice.

### Coding Style
Your code style follows the Java Coding Standard very closely. It is also very neat and tidy so Good Job Well Done! 👍
Just remember to take note of naming sense when it comes to variables that seem important to a segment of the program such that others reading your code can better understand it. 

Also, I would think it is good practice to include a separator line in between lines with different logic operations. An example is shown below (from `Parser.java`):

    private static Event createEventFrom(String inputLine) throws DukeException {
        try {
            String[] eventPart = inputLine.substring("event".length()).split("/at");
            String eventDescription = eventPart[0].strip();
            if (eventDescription.isEmpty()) {
                throw new DukeException(Message.MESSAGE_INVALID_EVENT_FORMAT);
            }
            LocalDateTime at = LocalDateTime.parse(eventPart[1].strip(),
                    DateTimeFormatter.ofPattern("d/M/yyyy HHmm"));
            return new Event(eventDescription, at);
        } catch (ArrayIndexOutOfBoundsException e) {
            throw new DukeException(Message.MESSAGE_INVALID_EVENT_FORMAT);
        } catch (java.time.format.DateTimeParseException e) {
            throw new DukeException((Message.MESSAGE_INVALID_DATE_FORMAT));
        }
    }

Instead of writing the above and clumping everything together, to further increase readability of your code, you can do something like this:

    private static Event createEventFrom(String inputLine) throws DukeException {
        try {
            String[] eventPart = inputLine.substring("event".length()).split("/at");
            String eventDescription = eventPart[0].strip();

            if (eventDescription.isEmpty()) {
                throw new DukeException(Message.MESSAGE_INVALID_EVENT_FORMAT);
            }

            LocalDateTime at = LocalDateTime.parse(eventPart[1].strip(),
                    DateTimeFormatter.ofPattern("d/M/yyyy HHmm"));

            return new Event(eventDescription, at);
        } catch (ArrayIndexOutOfBoundsException e) {
            throw new DukeException(Message.MESSAGE_INVALID_EVENT_FORMAT);
        } catch (java.time.format.DateTimeParseException e) {
            throw new DukeException((Message.MESSAGE_INVALID_DATE_FORMAT));
        }
    }

As you can see, now it is very clear that in the `try` block, you are first trying to find information in the input string, then checking for incorrect inputs then finally making a new `Event` from the input information. 
So although they are just empty lines, they are really useful in separating different logic in your code and can improve others' ease of reading and understanding drastically! 😎

krusagiz:
# Code

I like the modularity of your code, every class is packaged nicely into easily understood packages, it is a great application of the OOP paradigm and SOLID principles. Overall, the code is great! 😊

Here is a suggestion that might make it better:
 
- In both your `Event.java` and `Deadline.java` classes, in the the `toString()` method: (This one is from `Deadline.java`)
   ```

   public String toString() {
        return "[D]" + super.toString() + " (by: " + (this.by.getDayOfMonth() % 10 == 1
                ? DateTimeFormatter.ofPattern("d'st of' MMMM yyyy, h.mma").format(this.by)
                : this.by.getDayOfMonth() % 10 == 2
                    ? DateTimeFormatter.ofPattern("d'nd of' MMMM yyyy, h.mma").format(this.by)
                    : this.by.getDayOfMonth() % 10 == 3
                        ? DateTimeFormatter.ofPattern("d'rd of' MMMM yyyy, h.mma").format(this.by)
                        : DateTimeFormatter.ofPattern("d'th of' MMMM yyyy, h.mma").format(this.by)) + ")";
    }

    ```
    You may want to abstract a method to print the **1st, 2nd, 3rd** days of the month instead of using so many conditional operators. I think it will make it more readable?


# Style
Your code mostly follows the [coding style](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/basic.html) set out by the Professor. It is neat and tidy too. I also like how your import statements follow the SE-EDU [AB3](https://github.com/se-edu/addressbook-level3/blob/master/docs/SettingUp.adoc) conventions. Great Job! 👍

I do have a suggestion regarding a certain aspect that could make things better:

- In your `Event.java` and `Deadline.java` classes, your `LocalDateTime` variables are too short:
   ```

   private LocalDateTime at;
   
   ```
   Would it be better if you used a longer and more descriptive name such as `eventDateTime` to make it look less like a [scratch variable](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#naming)?

jun-ha0:
Thank you @ning-sy210 @krusagiz for the feedback!

PR #267, opened 2019-08-29
[Siew Yee] Duke Increments

liawsy:


src/main/java/Duke.java
```diff
@@ -1,10 +1,38 @@
...
+     * Updates and changes this file according to user input.
+     */
+    private Duke() {
+        greeter = new MessageGenerator();
+        sc = new Scanner(System.in);
+        ui = new UI("files/tasks.text");
+        isExit = false;
+    }
+
+    private void run() {
```
aidilfbk: Very clean code thanks to good use of abstraction :+1:

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+
+    List<Task> taskList = new ArrayList<>();
+    MessageGenerator msgGenerator = new MessageGenerator();
+
+    /**
+     * Creates task list and adds a placeholder value for easier tracking.
+     */
+    TaskList() {
+        //placeholder value in task
+        taskList.add(null);
```
aidilfbk: I'm not exactly sure what the `null` placeholder is for? My guess is to achieve one-based indexing but it seems to complicate the implementation of other methods. It could even be surprising if `getTaskList()` is called and the resulting `List<Task>` is used by other developers unaware of this convention.

src/main/java/duke/Date.java
```diff
@@ -0,0 +1,84 @@
...
+public class Date {
+    int day;
+    Month month;
+    int monthNum;
+    int year;
+
+    /**
+     * Documents months of a year.
+     */
+    public enum Month {
```
aidilfbk: Good use of enums 👍

src/main/java/duke/Date.java
```diff
@@ -0,0 +1,84 @@
+package duke;
+
+public class Date {
```
aidilfbk: Good minimalistic representation of a date but may I know why this solution was chosen over Java's built in [`Date`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Date.html) class? I believe you could leverage it to avoid having to validate inputs like "32/1/2019".

src/main/java/duke/InvalidCommandException.java
```diff
@@ -0,0 +1,21 @@
+package duke;
+
+class InvalidCommandException extends DukeException {
+
+    private String message = "☹ OOPS!!! ";
```
aidilfbk: There seems to be duplication of this line in other `DukeException` subclasses from the abstract parent class. I'm unsure if this was intentional.

src/main/java/duke/InvalidCommandException.java
```diff
@@ -0,0 +1,21 @@
...
+     */
+    public InvalidCommandException() {}
+
+    /**
+     * Creates String for error message.
+     * @return error message.
+     */
+    @Override
+    String errorMessage() {
+        this.message += "I'm sorry, but I don't know what that means :-(";
```
aidilfbk: Multiple calls of the `errorMessage()` method will append the additional message multiple times. I believe this is unintended behaviour.

Consider performing the append in the constructor instead, so it only executes once. A similar issue exists in other `DukeException` subclasses (i.e. [`InvalidInputException`](https://github.com/nus-cs2103-AY1920S1/duke/pull/267/files#diff-e609e4de614bc4af1b3ecb7527c00b0aR13), [`InvalidItemException`](https://github.com/nus-cs2103-AY1920S1/duke/pull/267/files#diff-7b846950507033d2eb0991da6c05b6a0R18) and [`MissingInputException`](https://github.com/nus-cs2103-AY1920S1/duke/pull/267/files#diff-aa6de11dac8d62841d1ed29540c3bafaR21)).

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Creates Duke Exception for exceptions in programs.
+     */
+    DukeException() { }
+
+    /**
+     * Creates DukeException that are related to tasks.
+     * @param taskType String taken in for error printing purposes.
+     */
+    DukeException(String taskType) {
```
aidilfbk: It seems like are two different kinds of `DukeException`s, one for general exceptions and another in relation to a task. It might be a good idea to create a new abstract class that inherits from `DukeException` for latter case.

src/main/java/duke/MessageGenerator.java
```diff
@@ -0,0 +1,138 @@
...
+
+    Formatter formatter = new Formatter();
+
+    /**
+     * Generates messages for different user input.
+     */
+    public MessageGenerator() {}
+
+    /**
+     * @return String for header of task list.
```
aidilfbk: I'm not sure if this Javadoc form where there is no short summary of the method is acceptable for this module's code style, although I would have done the same for these particular methods.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    Parser(){};
+
+    /**
+     * Takes in line of information to process into commands for program to read.
+     * @param line task description or information needed for processing.
+     * @return command used for updating tasks.
+     * @throws InvalidCommandException when program gives an invalid command type.
+     */
+    Command process(String line) throws InvalidCommandException {
```
aidilfbk: Good use of checked exceptions here but it may be unnecessary as the try-catch block is already catching the `DukeException` superclass.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,141 @@
...
+                return new Command(CommandType.DELETE, line);
+            case "find":
+                return new Command(CommandType.FIND, line);
+            default:
+                throw new InvalidCommandException();
+            }
+        } catch (DukeException e) {
+            e.printError();
+        }
+        return new Command();
```
aidilfbk: I believe this line will only be reached when the `line` input is invalid, thus would it make more sense to return a `null` here instead?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Loads and writes into given file.
+     * @param filePath String that indicates path to file.
+     */
+    Storage (String filePath) {
+        this.filePath = filePath;
+        f = new File(filePath);
+        try {
+            if (!f.exists()) {
```
aidilfbk: It's great that you checked for the existence of the file right at the constructor so it doesn't create a problem later when the rest of Duke is executing. :+1:

src/main/java/duke/Time.java
```diff
@@ -0,0 +1,49 @@
...
+        return new Time(Integer.parseInt(timeString));
+    }
+
+    /**
+     * Formats Time object.
+     * @return A string in the form HHMM.
+     */
+    @Override
+    public String toString() {
+        return time + "";
```
aidilfbk: I understand this line is to convert the `time` variable of `int` type to a `String` but consider using [`Integer.toString()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Integer.html#toString()) to make the string conversion explicit.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,55 @@
...
+    private Date date;
+    private Time time;
+
+    /**
+     * Creates Deadline (also a Task).
+     * @param num Number in the list.
+     * @param task String indicates description of task.
+     * @param date Date at which deadline occurs.
+     * @param time Time at which deadline occurs.
+     * @param type String indicates task type, in this case "D".
```
aidilfbk: I think it would be a good idea to prevent developers from having to remember to specify the right task type every time by hardcoding it in the call to the `super(num, task, type, done);` constructor to be `super(num, task, "D", done);` instead. But, I may be misunderstanding why this flexibility is desirable.

A similar issue exists in other `Task` classes (i.e. [`Event`](https://github.com/nus-cs2103-AY1920S1/duke/pull/267/files#diff-2ee21c0e5cbfdab2f3cef781d91263a5R14) and [`Todo`](https://github.com/nus-cs2103-AY1920S1/duke/pull/267/files#diff-f9f74394a8caf25c3a67670a93c)).

src/main/java/duke/MessageGenerator.java
```diff
@@ -0,0 +1,138 @@
+package duke;
+
+import java.util.List;
+
+public class MessageGenerator {
```
aidilfbk: :+1: for a `MessageGenerator` class as it will make transitioning to a GUI much easier.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,58 @@
...
+public abstract class Task {
+
+    boolean done;
+    private int no;
+    String task;
+    private String type;
+
+    /**
+     * Creates Task with an item number, task description, task type and done status.
+     * @param num Task number in the list.
```
aidilfbk: Should the Task instance know about its position/Task number in the list it is contained by? My understanding is that this information should be known only by the list itself.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,136 @@
+package duke;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class TaskList {
+
+    List<Task> taskList = new ArrayList<>();
+    MessageGenerator msgGenerator = new MessageGenerator();
```
aidilfbk: Should `TaskList` instances be calling the `MessageGenerator.print*` family of methods? My understanding is that a `TaskList` should just be a container for `Task`s and the classes that use a `TaskList` are in charge of notifying the user of any changes to the list.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    String getTaskInfo() {
+        return this.task;
+    }
+
+    /**
+     * Formats task to be written in given file.
+     * @return formatted string for writing in file.
+     */
+    public abstract String fileFormat();
```
aidilfbk: Delegating `Task` subclasses to implement how they should be formatted when written to file is 👍.

At the same time, I'm wondering if these subclasses should also be in charge of how the resulting string should be parsed to reproduce the same `Task` instance.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param taskNo Task's number in the list.
+     * @param taskType tasks of type Todo/Deadline/Event.
+     * @param arr String array that contains task description that has been processed.
+     * @return Task created from given inputs.
+     */
+    Task createNewTask(int taskNo, String taskType, String[] arr) {
+        boolean firstInDescription = true;
+        Date date = null;
+        Time time = null;
+        for (int i = 1; i < arr.length; i++) {
```
aidilfbk: It would be good to have a comment here to provide clarity that lines 112-121 are reconstructing the `Task`'s description from the space-separated command in the `arr` variable.

aidilfbk (commented):
Hi Siew Yee,

I've made my comments on specific lines where relevant and am seeking your clarification on them. However, these are my thoughts in general:

- Javadocs were clear in explaining what each method does. :+1:
- Code is well organized and architecturally sound.
- Classes are properly packaged and inheritance is used sensibly.
- There are a few small code style violations (e.g. [commas should be followed by a white space](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#layout)). Consider setting up Checkstyle ([via Gradle if you're already using it](https://github.com/nus-cs2103-AY1920S1/duke/blob/master/tutorials/gradleTutorial.md#checkstyle)) or make a habit of using your IDE to perform style conformance before committing. If you're using IntelliJ, it has a handy feature called [Reformat Code](http://jetbrains.com/help/idea/reformat-and-rearrange-code.html#reformat_file) and I use it frequently.
- Javadocs violate the code style's requirement for an [empty line between description and parameter section](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html#comments)
- It would be good to be explicit about access modifiers on all constructors, methods and properties.

I hope you found my feedback useful! 😄

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Processes String to retrieve keyword for search.
+     * @param command command for finding matching tasks.
+     * @return keyword for matching tasks.
+     */
+    String getKeyword(Command command) {
+        String line = command.getDescription();
+        String[] description = line.split(" ");
```
chen-xi-cx: variable name should be "descriptions"  according to java coding standards (Plural form should be used on names representing a collection of objects.)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    TaskList() {
+        //placeholder value in task
+        taskList.add(null);
+    }
+
+    /**
+     * @return Number of tasks in list.
+     */
+    private int noTasks() {
```
chen-xi-cx: I personally feel that naming getNumTasks() would be more intuitive because I read it as "no tasks" at first glance which was confusing. I think using "num" for "numbers" is the convention.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+    void loadTask(Task task) {
+        taskList.add(task);
+    }
+
+    /**
+     * Removes task of given identification number from list.
+     * Prints message when task is removed.
+     * @param taskNo identification number of task in list.
+     */
+    private void removeTask(int taskNo) {
```
chen-xi-cx: Same as above, I recommend using taskNum.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+    void printList() {
+        msgGenerator.printList(listify(this.taskList));
+    }
+
+    /**
+     * Checks if given task number exists in task list.
+     * @param taskNo identification number for task.
+     * @return whether or not task number exists.
+     */
+    private boolean invalidTaskNo(int taskNo) {
```
chen-xi-cx: isInvalidTaskNum() may be more in line with the coding standard,

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,44 @@
...
+    public String toString() {
+        return String.format("[T][%s] %s", done ? "✓" : "✗",task);
+    }
+
+    /**
+     * Formats task to be written in given file.
+     * @return formatted string for writing in file.
+     */
+    @Override
+    public String fileFormat() {
```
chen-xi-cx: Great addition, more elegant than the method I used.

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public boolean isExit() {
+        return isExit;
+    }
+
+    /**
+     * Processes file in storage and adds tasks to program.
+     */
+    public void processFile() {
+        for (Task task: storage.loadTasks().getTaskList()) {
```
chen-xi-cx: I think you need a whitespace before ":"  but I am not too sure.

src/main/java/duke/Parser.java
chen-xi-cx: This line may not be executed even after u catch the error. (see [Flow control in try catch finally in Java](https://www.geeksforgeeks.org/flow-control-in-try-catch-finally-in-java/)). You may want to consider using the finally block for this line (as below)

`finally {
return new Command();
}`

chen-xi-cx (commented):
Good effort. Code is easy to read with the help of javadoc comment.
Please take note of the following:
1. You can divide your packages into smaller packages to better categorize the code files.
2. For javadoc comment, you need to have an empty line between description and @param etc
3. You need to add access modifier to all your methods and attributes. See [here](https://qr.ae/TWygFD) for why it is good to do so.

test/java/DukeTest.java
```diff
@@ -0,0 +1,28 @@
...
+import duke.Todo;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class DukeTest {
+
+    /**
+     * Tests if a Date object returns the correct string format for printing.
```
Oscar-B-Liang: I think the code can be more well-tested to add more test cases.

src/main/java/Exceptions/InvalidCommandException.java
```diff
@@ -0,0 +1,21 @@
+package Exceptions;
+
+public class InvalidCommandException extends DukeException {
+
+    private String message = "☹ OOPS!!! ";
+
+    /**
+     * Creates InvalidCommandException when commands that are not understood by program are given.
```
Oscar-B-Liang: I like this part that you specify duke exceptions into more detailed cases.

src/main/java/UI/MessageGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+import Task.Task;
+
+import java.util.List;
+
+public class MessageGenerator {
+
+    Formatter formatter = new Formatter();
+
+    /**
+     * Generates messages for different user input.
```
Oscar-B-Liang: Here is a minor suggestion that it might be better to give the message generating function to the TaskList, UI and each specific task. I think this can reduce the workload in the future if we want to modify these commands. Do you agree?

Oscar-B-Liang (commented):
Hello. One of my peers did not create his PR, so I came to review your code, please excuse me.
Your code is in general very well organized and have high quality. I added some minor comments. The code can look even better developed if there are more test cases.
Hope these comments are useful for you.

src/main/java/Data/Command.java
```diff
@@ -0,0 +1,37 @@
+package Data;
```
Wingedevil: Perhaps it will be better to call the package something else other than Data?

src/main/java/Data/Parser.java
```diff
@@ -0,0 +1,174 @@
...
+        case "find":
+            return new Command(CommandType.FIND, line);
+        case "update":
+            return new Command(CommandType.UPDATE, line);
+        default:
+            return new Command(CommandType.INVALID);
+        }
+    }
+
+    public String[] getUpdateInfo(Command command) throws MissingInputException, InvalidInputException {
```
Wingedevil: Do include javadocs for methods that are not straight-forward .

src/main/java/Data/Parser.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param arr String array that contains task description that has been processed.
+     * @return Task created from given inputs.
+     * @throws MissingInputException when description is incomplete.
+     */
+    Task createNewTask(int taskNo, String taskType, String[] arr) throws MissingInputException, InvalidInputException {
+        boolean firstInDescription = true;
+        String desc = "";
+        Date date = null;
+        Time time = null;
+        for (int i = 1; i < arr.length; i++) {
```
Wingedevil: Put it in a different method??

src/main/java/Task/Date.java
```diff
@@ -0,0 +1,95 @@
+package Task;
+
+import Exceptions.MissingInputException;
+
+public class Date {
```
Wingedevil: Consider using the inbuilt java class for DateTime to ease your pain.

PR #268, opened 2019-08-29
[Lim Yong Shen, Kevin] Duke Increments

lyskevin:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,75 @@
...
+        ui.close();
+    }
+
+    /**
+     * Executes the program.
+     * @param args Command line arguments (unused).
+     */
+    public static void main(String[] args) {
+        new Duke("/Users/lyskevin/Desktop/Y2 Sem 1/CS2103T/duke/src/main/java/duke/data/tasks.txt").run();
+    }
```
cheongsiuhong, line 80: I believe truncating the path name starting from the root would work as well. eg "duke/data/tasks.txt". Anyone who clones your repo won't be able to get the storage function to work with this path!

src/main/java/duke/commands/IncorrectCommand.java
```diff
@@ -0,0 +1,43 @@
+package duke.commands;
+
+/**
+ * Implements the incorrect command, which handles invalid user input.
+ * @author Lim Yong Shen, Kevin
+ */
+public class IncorrectCommand extends Command {
+
```
cheongsiuhong, line 8: Perhaps using exceptions would be a better approach? This makes this class look like a working class part of the functionality.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+            return prepareEventCommand(input);
+        case ExitCommand.COMMAND_WORD:
+            return prepareExitCommand();
+        case FindCommand.COMMAND_WORD:
+            return prepareFindCommand(input);
+        case ListCommand.COMMAND_WORD:
+            return prepareListCommand();
+        case TodoCommand.COMMAND_WORD:
+            return prepareTodoCommand(input);
+        default:
```
cheongsiuhong, line 61: Maybe try using `enum`.

cheongsiuhong (commented):
Not many comments from me, the code generally looks good. Maybe try having more JUnit tests (but I know it's bothersome >.>) !

src/main/java/duke/parser/Parser.java
```diff
@@ -69,7 +69,7 @@ public static Command parse(String input) {
      */
     private static String getArguments(String input) throws ArrayIndexOutOfBoundsException {
         try {
-            return input.split(" ", INPUT_STRING_SPLIT_LIMIT)[ARGUMENTS_INDEX];
+            return input.split(" ", INPUT_STRING_SPLIT_LENGTH_LIMIT)[ARGUMENTS_INDEX];
```
kangyeelim, line 74: It might work with just throwing the exception e without catching it then throwing in the catch block?

src/main/java/duke/parser/Parser.java
```diff
@@ -35,37 +35,26 @@ public static Command parse(String input) {
             return prepareIncorrectCommand();
         }
         String commandWord = getCommandWord(input);
-        Command command;
         switch (commandWord) {
         case DeadlineCommand.COMMAND_WORD:
```
kangyeelim, line 43: switch cases make things cleaner :)

src/main/java/duke/commands/FindCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    public FindCommand(String keyword) {
+        this.keyword = keyword;
+    }
+
+    /**
+     * Executes this FindCommand and returns its CommandResult.
+     * @return This FindCommand's CommandResult.
+     */
+    @Override
+    public CommandResult execute() {
```
kangyeelim, line 31: Maybe it is better to put the find functionality with the substring method in FindCommand rather than TaskList?

kangyeelim:
Overall, the code is quite clean and it was a pleasure to review :) I think it may be better to print out slightly more specific exception messages for the users (tho it takes a bit of time)

src/main/java/duke/Duke.java
lyskevin: Hi, I did do this prior to this commit but then I think I accidentally messed with the IntelliJ file structure and multiple paths didn't work for me. I left this here in order to finish the level but I've fixed it and it's back to a relative path now :)

src/main/java/duke/commands/IncorrectCommand.java
lyskevin: I did consider using an Exception but I made the decision to create this class because I felt like any incorrect inputs could still be encapsulated within their own command. I appreciate the feedback and will look into using custom Exceptions in future.

src/main/java/duke/parser/Parser.java
lyskevin: Good point! I was thinking of using enums for other purposes but I chose not to do so here since the criteria for comparison is the first word of the input string. I did actually use an enum in an older version of the code which ultimately got refactored away. I'll keep this in mind for the GUI portion of the project!

src/main/java/duke/parser/Parser.java
lyskevin: Thanks for the feedback; I didn't know about that. I'll look into it!

src/main/java/duke/parser/Parser.java
lyskevin: I agree :)

src/main/java/duke/commands/FindCommand.java
lyskevin: That's true... I'll keep this in mind when refactoring the code. I'll look into cleaning up the code after implementing the GUI :)

PR #270, opened 2019-08-29
[John Ng] Duke Increments

JohnNzj:


Ellieyee (commented):
For Parser.java: 
Can consider using a switch case instead of if-else statements as it can be harder to trace the effects. The switch case implemented in line 42 can be abstracted out into a separate function for readability (or can consider storing pre-loading into a hashmap). Overall, comments are concise and accurately follows the javadoc coding standards 👍

src/main/java/Ui.java
```diff
@@ -124,5 +124,16 @@ public void listTask(TaskList tasks) {
             System.out.println(i + "." + tasks.getList().get(i - 1));
         }
     }
+
+    public void printFind(String keyword) {
```
Ellieyee: Can consider adding an alternative message if no matching tasks are found.

src/main/java/Command.java
```diff
@@ -72,6 +72,10 @@ public void execute(TaskList tasks, Ui ui, Storage storage) {
             tasks.delete(deletionNumber);
             ui.printDelete(toDelete);
             break;
+        case "find":
```
Ellieyee, line 126: Each case can be extracted out into separate functions or sub-classes for readability.

src/main/java/Deadline.java
```diff
@@ -1,12 +1,26 @@
...
+ */
 public class Deadline extends Task {
 
     protected String by;
 
+    /**
+     * Creates a deadline task.
+     *
+     * @param description A string representing the description of the task
+     * @param by A string representing the deadline of that task.
```
Ellieyee, line 12: String "by" can be renamed to clearer description E.g taskDeadline

src/main/java/Event.java
```diff
@@ -1,12 +1,25 @@
...
+ */
 public class Event extends Task {
 
     protected String when;
 
+    /**
+     * Creates a Task of type Event.
+     *
+     * @param desription A string representing the task's description.
+     * @param when A string representing time of the task.
```
Ellieyee, line 12: String when can be renamed to clearer description E.g eventDuration

src/main/java/Command.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Responds accordingly to the user's command.
+     *
+     * @param tasks A TaskList object containing the list of tasks.
+     * @param ui An Ui object that helps to interact with the user.
+     * @param storage A Storage object that contains a file with the tasks in it.
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
```
ang-zeyu: Since commands are limited to a fixed set of values, maybe it would be a good idea to use an enum?

I think it also offers a trivial runtime benefit as you only call CommandsEnum.valueOf(taskName) once ( and then catch IllegalArgumentException if invalid), and compare for enum equality.
( In C / C++ / C# enums are fundamentally implemented with integers, I believe it is the same for Java but I cant find a good source on it thus far )

(instead of comparing string equality till the correct string is reached)

ang-zeyu: As per Ellieyee's comment below, you could also extract and make abstractions for each of the individual commands.

src/main/java/Duke.java
```diff
@@ -1,10 +1,46 @@
...
+    public Duke(String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        tasks = new TaskList(storage.load());
+    }
+
+    /**
+     * Loads and saves files by applying other classes.
+     */
+    public void run() {
```
ang-zeyu: 👍 Effective and clean REPL!

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
+/**
+ * The Event class represents a type of task.
+ */
+public class Event extends Task {
+
+    protected String when;
```
ang-zeyu, line 6: When could be renamed to a more semantic 'period', or a seperate 'start', and 'end date'!

src/main/java/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Gets the string representation of a date by converting them from dd/mm/yyyy hhmm
+     * to english sentences.
+     *
+     * @param time A string representing date in the form of dd/mm/yyyy hhmm
+     * @return A string representation of a date in english sentences.
+     */
+    public static String convertDateAndTime(String time) { // in the form of dd/mm/yyyy hhmm
```
ang-zeyu, line 26: You could use java's inbuilt java.time (aka JodaTime) library to do the parsing ( and even formatting output ) instead!
https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html
Classes relavant ( I believe ): DateTimeFormatter, TimeFormatter, LocalDateTime, LocalTime

Regarding the java documentation (while browsing the docs):
There are many date time related classes outside of the java.time which are deprecated and harder to use (pre java8 apis).

src/main/java/Storage.java
```diff
@@ -0,0 +1,201 @@
...
+     * @param number An int representing the task number.
+     */
+    public void updateDone(int number) {
+        try {
+            String s = "";
+            Scanner scanFile = new Scanner(file);
+            for (int i = 1; i <= TaskList.size(); i++) {
+                if (i == number) {
+                    if (i == TaskList.size()) {
+                        s = s + scanFile.nextLine().replaceFirst("0", "1");
```
ang-zeyu, line 46: What if a task had already been marked as done, and the task description has a 0 in it?

_Suggestion \:_
- Either split the nextLine by the pipeline character seperator '|' and edit the second element of the array
- Or change the done command not to execute if the task was already marked as done!

src/main/java/Storage.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @return An ArrayList containing tasks from the file.
+     */
+    public ArrayList<Task> load() {
+        ArrayList<Task> tasks = new ArrayList<>();
+        try {
+            Scanner sc = new Scanner(file);
+            while (sc.hasNext()) {
+                String type = sc.next();
+                switch (type) {
```
ang-zeyu, line 81: - Another good place to consider the use of 👍 enums 👍  
- Could also check for validity of storage data (e.g. if the file is corrupted / of the wrong format ) and throw the appropriate exceptions if the conditions are not met.

src/main/java/Storage.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+        return tasks;
+    }
+
+    /**
+     * Add the task of type Todo to the file.
+     *
+     * @param taskname A string representing the task description.
+     */
+    public void addTodo(String taskname) {
```
ang-zeyu: _Suggestion \:_
Generalise the logic when adding a new task to avoid adding methods every time a new type of Task is added.
For example (pseudo code) \:
    `fileWriter.write(String.format("%s | %s | %s", task.getType().toString(), task.isDone(), task.getDescription());`

src/main/java/Ui.java
```diff
@@ -0,0 +1,139 @@
+import java.util.Scanner;
+
+/**
+ * The Ui class provides methods to deal with user interactions
+ */
+public class Ui {
```
ang-zeyu, line 4: Clean abstraction of ui overall 👍

ang-zeyu (commented):
Missing summary message from first PR review :

👍 Clean code overall! 👍 
Could seperate the abstractions for commands and storage in future iterations to increase code reusability.

src/main/java/Command.java
```diff
@@ -0,0 +1,104 @@
...
```
FelixNWJ: Since the execute method is going to be different for all commands, its better to make this method (and by extension the class) abstract instead of having a switch case, as it makes your execute method have too much functionality. It is better to break it up instead of possible

src/main/java/Duke.java
```diff
@@ -1,10 +1,46 @@
...
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
```
FelixNWJ: Systematic coding of each individual action that you want to execute to a corresponding method.

src/main/java/Parser.java
FelixNWJ, line 26: As what Zeyu stated above, unless your particular output format is not able to be returned by any methods in the java Time classes, its better to use their library since it saves readers the need to read your implementation and makes the code look neater as well.

src/main/java/Storage.java
```diff
@@ -0,0 +1,201 @@
...
+    public Storage(String filePath) {
+        file = new File(filePath);
+    }
+
+    /**
+     * Updates the task in the file to be marked as done.
+     *
+     * @param number An int representing the task number.
+     */
+    public void updateDone(int number) {
```
FelixNWJ: Method looks succint and each line does one task, well done!

src/main/java/Task.java
```diff
@@ -0,0 +1,50 @@
+/**
+ * The Task class represents the things to do.
+ */
+public class Task {
```
FelixNWJ, line 4: Since the Task class is primarily used for inheritance for other subclasses, and we do not actually save a Task parent object, its better to make the class abstract instead.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,65 @@
+import java.util.ArrayList;
+
+/**
+ * The TaskList class contains the task list and
+ * has methods to modify the tasks in the list.
+ */
+public class TaskList {
+
+    private static ArrayList<Task> tasks;
```
FelixNWJ: In the event that your Duke uses many TaskLists for a certain feature, I think that the tasks list should be implemented as a non static attribute, since one tasklist should not be used for all Duke objects created.

src/main/java/Ui.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public void printFind(String keyword) {
+        System.out.println("Here are the matching tasks in your list:");
+        int count = 1;
+        for (int i = 0; i < TaskList.size(); i++) {
+            if (TaskList.getList().get(i).getDescription().contains(keyword)) {
+                System.out.println(count + "." + TaskList.getList().get(i));
+                count++;
+            }
```
FelixNWJ, line 94: Code looks clean here and focuses on the functionality of the UI, well done!

src/main/java/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * The Parser class deals with making sense of the user's command.
+ */
+public class Parser {
+    /**
+     * Gets a Command object
+     *
+     * @param command A string representing the user's command.
+     * @return A Command object.
+     */
```
FelixNWJ, line 14: Is there a need to create a Parser in this case? Since it directly returns a Command object without much modification

FelixNWJ (commented):
Overall code is clean and neat, while adhering to the coding style guide. However, abstractions and modularisation could be implemented more so that the code is more readable

PR #271, opened 2019-08-29
[Gu ChunZhi] Duke Increments

Crisgu:
-


src/main/java/Duke.java
```diff
@@ -1,10 +1,236 @@
...
+                    break;
+                case "done":
+                    int n = Integer.parseInt(a[1]);
+                    done(tasks.getlist(), n-1);
+                    break;
+                case "todo":
+                    try{
+                        todo(tasks.getlist(),s);
+                    }
+                    catch (ErrorException e) {
```
hcwong: Nice use of switch, maybe can also move the try catch block outside the switch cases, since you're always catching `ErrorException`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,236 @@
...
+
+        }
+    }
+    /**
+     * Executes the command and returns the result.
+     * @param s user command
+     * @return result of the command
+     */
+    public void other(String s) throws ErrorException{
+        throw new ErrorException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
hcwong, line 223: Hmm if this function is only used to throw errors and used once, can just put in the switch statement directly?

src/main/java/Duke.java
```diff
@@ -1,10 +1,236 @@
...
+    }
+    /**
+     * Executes the command and returns the result.
+     * @param p user command
+     * @return result of the command
+     */
+    public void delete(ArrayList<Task>list,int p){
+        Task rt = list.get(p);
+        ui.drawline();
+        System.out.println("     Noted. I've removed this task:");
```
hcwong, line 258: Can also put this formatted printing as a function in the `UI` class, since you already have the class. Same for the rest

src/main/java/Task.java
```diff
@@ -0,0 +1,54 @@
+/**
```
hcwong, line 1: Nice OOP! 📦

src/main/java/Storage.java
```diff
@@ -0,0 +1,96 @@
...
+        try {
+            File txtfile = new File(filePath);
+            Scanner s = new Scanner(txtfile);
+            while (s.hasNext()) {
+            String txttasks = s.nextLine();
+            String[] txtsplit = txttasks.split(",");
+            if (txtsplit[0].equals("T")) {
+                String description = txtsplit[2];
+                Task todo = new Todo(description);
+                if (txtsplit[1].equals("1")) {
```
hcwong, line 34: The 3 inner `if txtsplit[1].equals("1") ...` and `arrList.add()` can probably be abstracted to out, since all 3 conditionals do the same thing.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,71 @@
+import java.time.LocalDateTime;
```
hcwong, line 1: Nice file!

src/main/java/Duke.java
```diff
@@ -1,10 +1,236 @@
...
+        ui.drawline();
+        System.out.println("     added: "+t.getName());
+        ui.drawline();
+    }
+    /**
+     * Executes the command and returns the result.
+     * @param k user command
+     * @return result of the command
+     */
+    public void done(ArrayList<Task>list,int k){
```
hcwong: `execute` might be a better name for this function

Crisgu: Thank you for the advise, I will take care of that

PR #274, opened 2019-08-29
[Kenneth Fung] Duke Increments

kenneth-fung:


SendorasRight:
Nice work. I liked the small touches like putting a 3-second timer before the app exits, and the fact that you extracted out the main() from Duke into its own class. Just a few comments from skimming through the repo:

* Have you considered creating a package for the Parser, Storage, TaskList and Ui files? They could go under something like duke.utils or duke.tools to improve the organization of your code. Same goes for DukeException as well, especially if you plan on adding more classes with similar behaviours in the future.

* For all your classes, you seem to be catching general Exceptions in a handle-everything kind of catch block. If I may, I would suggest catching specific Exceptions (like IOException etc.) that you already have prepared an error handler for. If not, your Duke program may fail silently in the future with an error that you did not prepare Duke to handle. Moreover, you might want your Duke program to respond differently to different errors in the future.

* Instead of constantly putting repetitive blocks of if-code during a parsing method to check if an array is of the proper size, one way of writing it would be to just brazenly access the array anyway while containing the entire method body in a try-catch block checking for IndexOutOfBoundsException. Just a suggestion, but your method works fine too!

* You represented date and time using a LocalDateTime object, but implemented your own parsing algorithms. You might want to look into the Java 11 API for [DateTimeFormatter](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html) since they already have a (mostly) bug-free implementation that's already optimised for that specific purpose. It would shorten your code from 30 lines to just 4 or 5. Plus, it would easier to regression test the rest of your Duke program as well, since parsing dates would now no longer be your problem!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,113 @@
...
+            return new CommandExit();
+        default:
+            String unknownCommand = commandWords[0].length() > 20
+                    ? commandWords[0].substring(0, 20) + "..."
+                    : commandWords[0];
+            throw new DukeException("I'm sorry, I don't know what '" + unknownCommand + "' means :(\n\n"
+                    + "Type 'help' to get a list of commands I can read.");
+        }
+    }
+
```
ooimingsheng: Consider not using string literals.  Perhaps, have a final static String object for each of your command objects. For example, 

Current code:  

 case "bye":

Proposed change"

case CommandExit.COMMAND_WORD

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    @Override
+    public String toString() {
+        DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy, h:mma");
+        return String.format(
+                "[E][%s] %s (at: %s)",
+                isDone ? "+" : " ",
+                description,
+                time.format(dateTimeFormatter)
+        );
```
ooimingsheng: Consider using
String doneIcon = isDone ? "+" : " ";

 as you have done in Task.java. I feel its much more readable this way.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,165 @@
...
+                    reader.close();
+                } catch (IOException e) {
+                    System.err.println(e.getMessage());
+                }
+            }
+        }
+
+        return tasks;
+    }
+
```
ooimingsheng: I havent done this myself, but you might want to consider using REGEX to show to the person reading the code how you are parsing the string inputs

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,165 @@
...
+                try {
+                    reader.close();
+                } catch (IOException e) {
+                    System.err.println(e.getMessage());
+                }
+            }
+        }
+
+        return tasks;
+    }
```
ooimingsheng: This is also an incredibly long function. Might want to consider splitting into smaller methods.

ooimingsheng (commented):
Overall, it was quite well done. Please see the suggestions below for code specific comments.

PR #275, opened 2019-08-29
[Shaun Ng] Duke Increments

ShaunNgTX:


src/main/java/Task.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+    
+    public String toStringDone() {
+        return "Nice, I've marked this item as done\n\t\t[" + this.getStatusIcon() + "] " + this.description;
+    }
+
+    //...
+}
+
+class Deadline extends Task {
```
christabelteo: It would be good if your subclasses can have a new Java file on its own.

christabelteo: Same for your Event and Todo classes

text-ui-test/runtest.bat
```diff
@@ -0,0 +1,21 @@
+@ECHO OFF
```
christabelteo, line 1: I like how you try to automate the testing by having an input file, actual and expected output. *thumbs up*

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Scanner;
+
 public class Duke {
```
christabelteo: It would be good if you can add more OOP to your Duke program to handle each process of the program separately, in their own respective class files.

PR #276, opened 2019-08-29
[Abhiman Yadav] Duke Increments

Abhiman2211:


oscarsu97:
Commit message eg "Added J-unit testing", it should be "Add J-unit testing".

oscarsu97 (commented):
FilePath shouldn't include the whole file path but instead eg" data/duke.txt", so that not only just @Abhiman2211 can use it. Can try using Path class to get the root directory of the user using.

tishyakhanna97 (commented):
Good Exception Handling for the find task portion

tishyakhanna97 (commented):
Can try to have more descriptive variable names instead of i

PR #278, opened 2019-08-29
[Sudharshan Madhavan] Duke Increments

uggi121:


src/main/java/Parser.java
```diff
@@ -66,4 +70,21 @@ public void evaluateDone(String input) throws DukeException {
             }
         }
     }
+
```
joshuaseetss: I think you forgot to add the javadocs to this method!

src/main/java/TaskList.java
```diff
@@ -172,6 +172,16 @@ public void addDeadline(String description) throws DukeException {
             parent.printTaskAdded();
         }
     }
+
```
joshuaseetss: Same as above, I think you forgot Javadocs for this method too as well.

src/main/java/Date.java
```diff
@@ -25,7 +24,6 @@ public static boolean verifyDateTimeDeadline(String dateTime) {
     }
```
joshuaseetss: There's actually a package in java that deals with date formats and what not. 
It also helps us to convert from String -> Date format. It is called 
java.text.SimpleDateFormat;
But this is actually impressive!

src/main/java/Parser.java
```diff
@@ -0,0 +1,90 @@
+import java.io.IOException;
+import java.util.ArrayList;
+
+import duke.exception.DukeException;
+import duke.task.Task;
+
+/**
+ * Represents the class responsible for parsing and evaluating user input.
+ */
```
joshuaseetss: This is a good way to make the code more OOP! Good use of having a Parser class

uggi121: Hey! I completely remade my Duke. Thanks for the feedback though, good insights :)

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,116 @@
...
+        } else if (cleanedInput.startsWith(EVENT_COMMAND_START)) {
+            return makeEvent(cleanedInput);
+        } else {
+            throw new InvalidTaskDukeException("Invalid add command!");
+        }
+    }
+
+    private Todo makeTodo(String task) throws InvalidTodoDukeException {
+        try {
+            assert task.startsWith(TODO_COMMAND_START) : "This is not a to-do command";
```
yhtingg, line 75: Good use of assertions to validate the task! I didn't think of using it in this way, it's really insightful :)

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,27 @@
...
+     * Executes the command.
+     * @param tasks List of tasks
+     * @param ui User-Interface
+     * @param storage Storage object
+     * @return Duke's response as a string.
+     * @throws DukeException If something goes wrong.
+     */
+    public abstract String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    public abstract boolean checkExit();
```
yhtingg, line 20: Perhaps it would be simpler to return this instanceof ExitCommand, without having to use an abstract method :)

PR #279, opened 2019-08-29
[Wu Xia] Duke Increments

tswuxia:


linyutinglyt:
Very neatly arranged commits. Easy to refer.

src/main/java/Date.java
```diff
@@ -0,0 +1,67 @@
...
+ *  Represents a timing on a specific date. A <code>Date</code> object corresponds
+ *  to a date and a time. e.g.<code>27/08/2019 1800</code>
+ */
+public class Date {
+    protected int day;
+    protected String mon;
+    protected int yr;
+    protected double time;
+    protected String meridian;
+    private String[] months = {"January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"};
```
Yoshi275: Interesting way of storing dates. I'm quite curious why you created a Date class, since Java also has a Date class (albeit a flawed one).

src/main/java/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Reads and react to user command
+     * @param tasks Current task list.
+     */
+    public void read(TaskList tasks) {
+        while(true) {
+            String cmd = sc.nextLine();
+            Scanner cmdSc = new Scanner(cmd);
+            if (cmd.toLowerCase().equals("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
```
Yoshi275: Why not refactor this to be done by the UI instead, so all printing is done by the UI?

src/main/java/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+                    case "todo":
+                        try {
+                            if (cmdSc.hasNext()) {
+                                tasks.addToDo(cmdSc.nextLine());
+                            } else {
+                                throw new DukeException("The description of a todo cannot be empty.");
+                            }
+                        }
+                        catch (DukeException exp) {
+                            System.out.println("OOPS!!! " + exp.getMessage());
```
Yoshi275: Same issue, I think potentially you can refactor such that the UI handles all logic of what to print?

src/main/java/Storage.java
```diff
@@ -0,0 +1,70 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Retrieve the list of tasks from local file.
+     * @return A list of current tasks in record.
+     */
+    public List<Task> load() {
+        Path file = Paths.get(filePath);
+        List<Task> tasks = new LinkedList<>();
```
Yoshi275: I'm curious why you ended up using a LinkedList. Potentially, an ArrayList may be more efficient, especially when randomly accessing various parts of the array, and often not in sequence? Then again, modification of list may be more efficient using a LinkedList.

Yoshi275 (commented):
Overall, your code is quite readable and accessible. I think my main concern is the function of the UI class, if it doesn't print the response to the user in multiple cases. Considering refactoring further! Good job, keep it up!

PR #280, opened 2019-08-29
[Gabriel Ong Lian Xin] Duke Increments

Q-gabe:
In fulfilment of iP Week 3.


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Returns a string representation of the Deadline Task.
+     * @return a string representation of the deadline Task.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString()
+                + " (by: " + by.format(dateFormatter) + ")";
```
justarock111, line 46: Alternatively, you can format your dates in the constructor first and use it for future methods.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    /**
+     * Returns a string representation of the Deadline Task.
+     * @return a string representation of the deadline Task.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString()
```
justarock111, line 45: Good idea!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public Deadline(String description) throws DukeException {
+        super(description);
+        String[] splitDescription = description.split(" /by ", 2);
+
+        try {
+            this.description = splitDescription[0];
+            this.by = LocalDateTime.parse(splitDescription[1], dateFormatter);
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new DukeException("Please enter a deadline using \"/by\".");
```
justarock111: Very clear error messages.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.io.File;
+
+/**
+ * Duke.java
+ * CS2103T
+ * @author Gabriel Ong
+ *
+ * This program is an interactive task list that takes in several
+ * preset commands from the user to create tasks, view the list of
```
justarock111, line 7: Very detailed JavaDoc comments!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class Duke {
+
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Returns a instance of duke.
+     * @param filePath a filepath indicating the storage directory for duke
```
justarock111: I think there would need to be a ending period for each line of JavaDocs comments, as there might be checkstyle error.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+     * @throws DukeException for various input errors or if the
+     * command is invalid.
+     */
+    public String parse(String input) throws DukeException {
+        /*
+         * Check if command suffix to parse, split accordingly and run command
+         */
+        // "done" commands
+        if (input.startsWith("done ")) {
+            if (input.length() > DONE_SUFFIX_LENGTH) {
```
justarock111, line 34: Good use of constants!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,147 @@
+package duke;
+
+/**
+ * This class is responsible for parsing input from the UI, taking the
+ * String input and returning a String output.
+ */
+public class Parser {
+    private TaskList taskList;
+    private static final int DONE_SUFFIX_LENGTH = 5;
```
justarock111, line 9: Alternatively, you can define under an enum, like enum LENGTH {...}

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+        } else if (input.startsWith("todo")) {
+            return addToList(input, TaskType.Todo);
+        } else if (input.startsWith("deadline ")) {
+            return addToList(input, TaskType.Deadline);
+        } else if (input.startsWith("event ")) {
+            return addToList(input, TaskType.Event);
+        } else if (input.equals("list")) {
+            return this.taskList.getList();
+        } else {
+            throw new DukeException("I'm sorry, but I don't know what"
```
justarock111, line 64: Alternatively, you can define another class like that extends duke exception, like InvalidCommandException. So that you can just call it, so you would not need to type the message every time.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Loads a TaskList from the storage binary at the designated file path.
+     * @return a TaskList object loaded from the binary.
+     * @throws DukeException when an existing save was not found.
+     */
+    public TaskList load() throws DukeException {
+        try {
+            FileInputStream load = new FileInputStream(this.saveFile);
```
justarock111, line 34: Nice idea to use FileInputStream.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,90 @@
...
+        output.append("Here are the tasks in your list:\n");
+
+        // List and print all items stored
+        for (Task item: this.taskList) {
+            //noinspection StringConcatenationInsideStringBufferAppend
+            output.append(count++ + "." + item + '\n');
+        }
+
+        // Remove terminal newline character if at least 1 item inserted
+        if (count > 1) {
```
justarock111, line 73: Good idea!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,90 @@
...
+            output.deleteCharAt(output.toString().length() - 1);
+            return output.toString();
+        } else {
+            throw new DukeException("The task list is empty.");
+        }
+    }
+
+    public String getSearchList(String regex) throws DukeException {
```
justarock111, line 88: Alternatively, you can use streams instead as it it a much simpler to turn you ArrayList into a stream, then map to the String representation of element, then filter it, then collect it back to an ArrayList.

src/main/java/duke/TaskType.java
```diff
@@ -0,0 +1,10 @@
+package duke;
+
+/**
+ * Task representation upon parsing input.
+ */
+public enum TaskType {
```
justarock111, line 6: Just wondering, why would we need an enm for TaskType? Thank you!

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,80 @@
+import duke.DukeException;
+import duke.Parser;
+import duke.TaskList;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class ParserTest {
+    @Test
+    public void write_ToDoCommand_writtenCorrectly() {
```
justarock111, line 10: Good naming of test methods!

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test
+    public void write_deadLineCommand_empty_exceptionThrown() {
+        Parser testParser = new Parser(new TaskList());
+        String testToDoStr = "deadline ";
+
+        try {
+            assertEquals(testParser.parse(testToDoStr), "Got it. I've added this task:\n  " +
+                    "[D][✘] return book (by: 12/12/1212 1212)" + "\nNow you have 1 tasks in the list.");
```
justarock111: Just wondering, how were you able to get the console to print the icon? As I have errors printing icon accurately. Thank you!

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,80 @@
...
+            fail();
+        } catch (DukeException e) {
+            System.out.println("Exception successfully thrown: " + e.getMessage());
```
justarock111, line 62: I am not to sure about this, but Assert.equals might have to be used to check that the exception is thrown successfully, as printing it out would not allow JUnit to know that the test is passed. 

Also, there is a way to use JUnit to test if methods throw exception.
https://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests
However, I had error when doing this, but you can try it!

src/main/java/duke/Deadline.java
Q-gabe: I implemented this actually to keep the date stored as a LocalDateTime object and only convert it to a String using this formatter when actually needed. I guess storing a String representation could also work. Thank you for the suggestion!

src/main/java/duke/Duke.java
Q-gabe: Parameters actually don't need periods at the end, the official javadocs also do not actually add periods for their parameters. Descriptions do however!
Screenshot 2019-09-04 at 10 13 49 AM

src/main/java/duke/Parser.java
Q-gabe: I'm unsure if assigning an enum for a continuous variable like length is a good idea, but I will consider it considering that the cases here are quite limited. Thank you!

src/main/java/duke/Parser.java
Q-gabe: Good idea, thanks for the suggestion! That would make better OOP sense as well. (:

src/main/java/duke/TaskList.java
Q-gabe: I guess a stream implementation would look slightly neater, but personally I am more comfortable with a for-each loop and it might be clearer for collaborators too. Good suggestion though!

src/main/java/duke/TaskType.java
Q-gabe: I foresaw that I might need to implement more OOP features into the parser function since I found implementing a separate group of classes for commands, as suggested on the CS2103 website, a bit superfluous. Hence, I decided to use an enum for TaskType to determine the command type for the parser. (:

src/test/java/ParserTest.java
Q-gabe: Hmm, I am unsure, I just used the unicodes as was given, namely the "\u2713" and "\u2718" for the tick or X symbols respectively. Were you able to initially print the icons for the classes first? Or was this only a problem upon unit testing?

src/test/java/ParserTest.java
Q-gabe: Hmm, I kind of understood it as if the JUnit tests pass either if 1) fail() is not called or 2) an AssertEquals or any other JUnit-based assertions passes. The printing isn't really there to let JUnit pass the test but rather as a secondary check for the user to see the error message. I suppose I can implement a second AssertEquals against the error message of the exception to see if the expected error message is being thrown. That would be more thorough indeed.

As far as I found, this try-catch method of testing whether an exception is thrown successfully is actually quite standard besides the newer JUnit 5 and 4.7 methods as stated by one of the responders in the link you provided. But it's interesting regardless, I will try out the newer commands. Thank you for the great suggestion!

Q-gabe: Pardon the double post, but thank you for reviewing the code so thoroughly! It means a lot! (: 💯

src/main/java/duke/Duke.java
fadisng: Oh wow that's pretty specific thanks for sharing!

fadisng (commented):
Uses a range of Java classes in your code which is very interesting. Clean code and appropriate comments which makes it easy to read and understand.

PR #281, opened 2019-08-29
[Kyungho Min] Duke Increments

john0227:


Xelyion (commented):
Javadoc and comments are informative and concise. Might be possible to use java.util.Date instead of writing an entire date class.

KendrickAng:
Great work! I love that you took the effort to separate the Duke GUI and Duke CLI. You've also stylized your GUI well, and even defined your own Date class, which opens much flexibility for future changes. The javadocs are written well too, with the use of tags to add more detailed comments.

I'm impressed with the rigour of your code. It's overall well-abstracted, and the project is properly packaged and built with Gradle - it worked out of the box when I cloned it. There's a lot for me to learn from your code and abstraction!

john0227:
@Xelyion 
Thanks for the suggestion! I tried to but gave up halfway 😂. I'll work on it later!

john0227:
@KendrickAng 
Thanks for the comment! Now I feel good about myself 😁

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Runs the Duke application.
+     */
+    public void run() {
+        // Greet the user
+        this.ui.greet();
+
+        // Handle user input
+        boolean isExit = false;
```
Kyzure: isExit should be shouldExit - As Exit is not a noun.

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Returns false.
+     *
+     * @return False.
+     * @see Command#isExit()
+     */
+    @Override
+    public boolean isExit() {
```
Kyzure: The method lacks the explanation as to why isExit is always false.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public abstract String getResponse(TaskList taskList, CommandStack commandStack, Storage storage)
+            throws DukeException;
+
+    /**
+     * Shows whether executing this Command should quit the Duke or not.
+     *
+     * @return True if Duke should quit, false otherwise.
+     */
+    public abstract boolean isExit();
```
Kyzure: You can make this not abstract, but instead return false, and the command that requires exiting should override this instead and return true.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,184 @@
...
+     * @throws DukeIoException If an error occurs while saving.
+     */
+    @Override
+    public String[] undo(TaskList taskList, CommandStack commandStack, Storage storage) throws DukeIoException {
+        assert this.deletedTasks != null : "DeleteCommand.java (line 110) : deletedTasks should not be null.";
+
+        // Add the deleted tasks back.
+        String[] message;
+        switch (this.deletedIndex) {
+        case "-1":
```
Kyzure: Use a static variable or constant instead of throwing a magic String/Number

PR #282, opened 2019-08-29
[Yen Ren Zuo Brian] Duke Increments

brianyenna:
Part of Task 4 (Week 3 Individual Project Tasks)


src/main/java/duke/commands/Command.java
```diff
@@ -0,0 +1,42 @@
...
+import duke.exceptions.DukeException;
+import duke.utils.Storage;
+import duke.utils.TaskList;
+import duke.utils.Ui;
+
+/**
+ * Parent class for all Commands.
+ * Implements the logic that is fundamental to all Commands.
+ */
+public class Command {
```
mrnewguy, line 12: Well abstracted and good use of inheritance for other command classes.  But is the default behavior really to save the tasks? Is there a better default behavior?

src/main/java/duke/utils/DukeDate.java
```diff
@@ -0,0 +1,32 @@
+package duke.utils;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
```
mrnewguy, line 4: Good use of the SimpleDateFormat. Perhaps you can try writing your own date library too?

src/main/java/duke/utils/Parser.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Scanner;
+import java.util.ArrayList;
+
+/**
+ * Class used to parse all the string input coming from the user, manage error handling
+ * and generate Commands that can then be invoked via their .execute() methods to trigger certain
+ * behaviours. This class is meant as an abstraction to abstract away the details and complexities
+ * behind parsing using input.
+ */
+public class Parser {
```
mrnewguy, line 22: Good use of switch case statements to keep code concise. However, instead of having multiple parse methods (e.g parseToDo, parseEvent etc), perhaps you could have a parent parse class and have several children inheriting from the parse class. Another method which can shorten this Parser.java class is to have multiple Command Classes (e.g. eventCommand, toDoCommand) instead of just 1 type of AddCommand.

src/test/java/duke/utils/ParserTest.java
```diff
@@ -0,0 +1,67 @@
...
+import duke.exceptions.DukeException;
+import duke.commands.Command;
+import duke.commands.ByeCommand;
+
+import java.util.ArrayList;
+
+public class ParserTest {
+    @Test
+    public void testParse_Bye_ByeCommand() {
+        try {
```
mrnewguy, line 17: Don't think there is a need to put try catch statement in your tests. Just running an assert statement with the Junit library should be sufficient

mrnewguy (commented):
Overall great code, just that I think further inheritance could greatly reduce the length of the Parser.java file.

gachia:
Excellent usage of inheritance and abstraction. Quite well done.

src/main/java/duke/commands/ByeCommand.java
```diff
@@ -0,0 +1,26 @@
...
+     * @param allTasks TaskList object containing all tasks.
+     * @throws DukeException re-thrown from underlying method calls.
+     */
+    public void execute(Ui ui, Storage storage, TaskList allTasks) throws DukeException {
+        //Will cause while loop in Duke to exit
+        this.isExit = true;
+
+        //Parent method invoked to save TaskList
+        super.execute(ui, storage, allTasks);
+    }
```
gachia: Curious to know if the ByeCommand is able to execute and exit the program if a file path is not specified or broken. I encountered it on my code and yours look similar.

src/main/java/duke/commands/Command.java
brianyenna: You are definitely right. I was debating over whether this is an ideal default behaviour. The reason I went down this path is because I figured every command ought to result in the file being saved as a safety mechanism. But thinking a little further, this seems wasteful, as some commands never make any modifications, so saving the file is a waste of time and computational resources. 

I'll update this in a future commit. Thanks for pointing it out :)

src/main/java/duke/utils/Parser.java
brianyenna: That's definitely a possible approach! I'll look into it. Thank you!

src/main/java/duke/commands/ByeCommand.java
brianyenna: Hmm I didn't encounter this issue in my code. For me, the file path is saved as a static variable, so it does not change once the application starts. Could you clarify on the issue you are facing? Do you mean to say that pass your ByeCommand object a file path?

gachia: It's more of an edge case test, where the file path does not exist or is broken when the application starts. So when you try to save, a DukeException gets thrown and you aren't able to exit the program with a ByeCommand because the boolean variable was unable to set to True.

PR #285, opened 2019-08-29
[Sheng Ying] Duke Increments

ning-sy210:


jun-ha0:
<h2>Code:</h2>

I like the comprehensive handling of exceptions and usage of software design principles (Single Level of Abstraction, Liskov Substitution Principle, etc). 😃 

1.  For the private void run() method in Duke.java, I have the following comments:
```
       while (!instruction.equals("bye")) {
            try {
                parser.parseInstruction(instruction, taskList);
            } catch (DateTimeParseException e) {
                exceptionHandler.showParseDateTimeError();
                Ui.showLine();
            } catch (VoidDukeCommand e) {
                exceptionHandler.showVoidDukeCommandError();
                Ui.showLine();
            } catch (IncorrectDukeCommand e) {
                exceptionHandler.showDukeCommandEvaluationError(e);       
                Ui.showLine();
            } catch (InvalidDukeCommand e) {
                exceptionHandler.showUnknownDukeCommandError();
                Ui.showLine();
            } catch (IOException e) {
                exceptionHandler.showDukeIoError();
                Ui.showLine();
            } catch (DukeException e) {
                exceptionHandler.showDukeError(e);   
                Ui.showLine();
            } 
            instruction = sc.nextLine();
        }
        sc.close();
        ui.exit();
```   

- Good practice of closing scanner.
- It may be a good idea to write ui.showLine() by using the ui object that was instantiated, instead of Ui.showLine().
- It may be a good idea to put ui.showLine() in a finally block to shorten code:


```
public void run() {
    ui.showWelcome();
    boolean isExit = false;
    while (!isExit) {
        try {
            isExit = c.isExit();
        } catch (DukeException e) {
            ui.showError(e.getMessage());
        } finally {
            ui.showLine();
        }
    }
}
```

2. Add more OOP by extracting the following classes: Command, AddCommand, DeleteCommand, ExitCommand etc. that inherits from an abstract Command class. These allows the Parser to adhere to Separation of Concerns Principle (SoC) which can achieve better modularity, separate the code into distinct sections, such that Parser addresses only parsing. You can get some inspiration from how the code of the [addressbook-level2](https://github.com/se-edu/addressbook-level2) is organized.

3. There were some coupling between the classes due to additional features like temporarySearchList functionality.

4. Lastly, good use of singleton pattern to manage the various exceptions.

<h2>Style:</h2>

Neat and tidy. Good job! 👍 

Disclaimer: The following are optional. They are just some ideas to make the source code neater.

- It may be a good idea to follow SE-EDU [AB3](https://github.com/se-edu/addressbook-level3/blob/master/docs/SettingUp.adoc) convention for ordering of import statements.


- It may be useful to check out the codestyle import ordering with [IntelliJ.](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003038760-Codestyle-Import-ordering-with-the-new-version-of-IntelliJ-2018-3)

- Lastly, it may be a good idea to add Javadocs for all public classes and methods.

krusagiz:
# Code

I think that you did well in customising the error messages and implementing the code according to *SOLID* software design principles (Single Level of Abstraction, Liskov Substitution Principle, etc). 🎉

Here are some suggestions:

1. For your `Parser.java`, think it might be good to declare error messages in a constant like this:
```

private static final String EVENT_EMPTY_DESC_MSG = "The description of an task.Event cannot be empty!"

```

2. I think you do not need to put your `Duke.java` into a package by itself since it is the driver class? 

3. You might want to add more OOP by extracting the commands from the `Parser` class? This is to adhere to Separation of Concerns Principle (SoC), as the `Parser` class should only parse input.

# Style

Your code mostly follows the [coding style](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/basic.html) set out by the Professor. It is neat and tidy too. Good Job! 👍

ning-sy210:
Thanks to @krusagiz @jun-ha0 for your valuable feedback! 😃

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,60 @@
...
+            taskList = new TaskList(storage);
+        }
+    }
+
+    String getResponse(String input) {
+        Parser parser = new Parser();
+        String dukeResponse;
+        String errorMessage;
+
+        try {
```
ChrisKheng, line 49: Maybe instead of putting all the exceptions in one place, maybe you could put the exceptions into the Parser class and split the exceptions into methods to improve the clarity of the code?

src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,12 @@
+package exception;
+
+/**
+ *  Represents a general exception that is not caught by any of the other exceptions.
+ */
+public class DukeException extends Exception {
+    private static final long serialVersionUID = 1L;
```
ChrisKheng, line 7: You didn't use this class variable though

src/main/java/misc/Parser.java
```diff
@@ -0,0 +1,330 @@
...
+     * Writes Tasks as Strings that are saved into a local save file. The parsed String
+     * contains information about the task's completion status, Task type and Date/Time (if any).
+     * @param task The Task to be parsed into a String and saved into a local save file.
+     * @return The parsed String of the Task's details, null if task is of unknown type.
+     */
+    String writeTaskAsText(Task task) {
+        int isDoneStatus = task.isDone ? 1 : 0;
+        
+        if (task instanceof ToDo) {
+            return String.format("T : %d : %s", isDoneStatus, task.description);
```
ChrisKheng: Maybe you should use .toString() instead to make use of polymorphism, which can make your code shorter?

src/main/java/misc/Parser.java
```diff
@@ -0,0 +1,330 @@
...
+        return null;
+    }
+
+    /**
+     * Reads Strings (from a local save file) as Tasks. These tasks retain information of
+     * all user tasks when the program last exits.
+     * @param stringTask The String to be parsed into a Task.
+     * @return A Task in the user's task list when the program last exits. 
+     */
+    Task readTextAsTask(String stringTask) {
```
ChrisKheng, line 46: Maybe this method should be in the Storage class? Because it seems that this method is more related to processing data from the database `Storage` class.

PR #288, opened 2019-08-30
[Toh Chooi Ern, Jasmine] Duke Increments

tohcejasmine:


src/main/java/Duke.java
```diff
@@ -1,10 +1,44 @@
...
+    boolean inProgram;
+    private void run() {
+        //Initialise Scanner object
+        Scanner input = new Scanner(System.in);
+        //Continually receive commands until exit
+        inProgram = true;
+        while (inProgram) {
+            String userCommand = input.next();
+            if (userCommand.equals("bye")) {
+                System.out.println("-----");
```
AHaliq: ~~you could create a function that takes a string as argument and that prints it out with the horizontal lines instead of repeatedly printing them before and after every output e.g:~~ Just saw your next commit with the display function.

```Java
public void formatPrint(String str) {
    System.out.println("-----");
    System.out.println(str);
    System.out.println("-----");
}
```

src/main/java/Duke.java
```diff
@@ -19,26 +24,45 @@ public static void main(String[] args) {
...
-                System.out.println("-----");
-                System.out.println("    " + "Bye. Hope to see you again soon!");
-                System.out.println("-----");
-                inProgram = false;
-            } else {
-                System.out.println("-----");
-                System.out.println("    " + userCommand);
-                System.out.println("-----");
+            String userCommand = input.nextLine();
+            switch (userCommand) {
```
AHaliq: Could use an enum here

src/main/java/Duke.java
```diff
@@ -64,5 +65,27 @@ private void display(String response) {
         System.out.println("-----");
     }
 
+    //List out added tasks
+    private void displayTasks() {
```
AHaliq: good approach to extract into its own function!

src/main/java/Duke.java
```diff
@@ -64,5 +65,27 @@ private void display(String response) {
...
+        System.out.println("-----");
```
AHaliq: could have used your display function instead of printing out ----- again.

src/main/java/Duke.java
```diff
@@ -64,5 +65,27 @@ private void display(String response) {
...
+                    task.description);
+        }
+        System.out.println("-----");
+    }
 
-}
+    private void displayMarkDone(int doneIdx) {
+        Task doneTask = taskList.get(doneIdx-1);
+        doneTask.markDone();
+        System.out.println("-----");
```
AHaliq: same here, consider utilising your display function

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
```
AHaliq: `this.attribute` (for example `this.description`) or `attribute` (for example `return (isDone?...`), either way is fine, but stick to one convention

src/main/java/Duke.java
```diff
@@ -46,46 +49,109 @@ private void run() {
...
                     break;
                 default:
-                    taskList.add(new Task(userInput));
-                    display("added: " + userInput);
+                    display("\"" + firstCommand + "\"" +
+                            " is not a valid command. Try again.\n\n" +
+                            indentString +
+                            "List of valid commands: \n" +
+                            indentString +
+                            "todo, deadline, event, list, done, bye");
```
AHaliq: nice touch guiding the users to the right commands!

src/main/java/Duke.java
```diff
@@ -50,45 +62,71 @@ private void run() {
...
+                    //Empty Description (only "deadline" or "deadline /by")
+                    if (userInputArr.size() == 1 || userInputArr.get(1).equalsIgnoreCase(subCommand)) {
+                        throw new EmptyDescriptionException("event");
                     } else {
-                        display("Please write something after \"event\"!");
+                        int firstByIdx = userInputArr.indexOf(subCommand);
+                        int lastByIdx = userInputArr.lastIndexOf(subCommand);
+                        //No "/by" or multiple "/by"s provided
+                        if (firstByIdx == -1 || firstByIdx != lastByIdx) {
+                            throw new IncorrectInfoInputException(subCommand);
```
AHaliq: good use of exceptions!

AHaliq (commented):
A good straight forward approach using switch statements for command types.

src/main/java/Parser.java
```diff
@@ -0,0 +1,132 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+
+// Deals with making sense of the user command
+public class Parser {
+
+    // Constructor
+    public Parser() { }
+
+    public Command parse(String userInput, TaskList taskList) throws DukeException {
```
cambrian-dk: Maybe a comment here would be useful to explain what the parse method does generally, on top of the comments within the method

src/main/java/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+            break;
+        case "delete":
+            // Empty/no list index of task provided
+            if (userInputArr.size() == 1 || userInputArr.size() > 2) {
+                throw new DukeException("Please put the list index of the " +
+                        "completed task after \"delete\" and nothing else.");
+            } else {
+                // Check if integer is provided
+                try {
+                    int deleteIdx = Integer.parseInt(userInputArr.get(1));
```
cambrian-dk: Maybe it would reduce nesting if the logic of the commands was refactored into the class files of those commands.

cambrian-dk (commented):
Good coverage of incorrect usage with exceptions, some design choices can be made to make the code more SLAP(but that's a subjective choice)

PR #290, opened 2019-08-30
[Ryo Chandra Putra Armanda] Duke Increments

ryoarmanda:


minpyaemoe:
In my opinion, the flow of your code is logical and understandable to readers. Also, I noticed that you managed to implement abstraction and inheritance well by breaking down classes to improve the quality of your code. Many lines of codes are properly formatted and looks clean to me. Furthermore, I also observed a good way to use enumeration and constants in respective classes from your codes.

The only suggestion I would like to make is that by writing Java doc comments for methods and classes in your codes, I believe that readers will be able to have an idea of what your blocks of code does in short amount of time.

Nevertheless, to me, your coding style is straightforward and good in overall. I hope my review to your code is useful to you.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
+package duke;
+
+import duke.command.Command;
+
```
khiangleon, line 10: Have you considered including JavaDoc comments for this whole class?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param input The input string from the command-line.
+     * @return Command The command describing the action.
+     * @throws DukeException if the input syntax is invalid.
+     */
+    public static Command parseCommand(String input) throws DukeException {
+        String[] tokens = input.split(" ", 2);
+        String[] args;
+        String[] times;
+        switch (tokens[0]) {
```
khiangleon: This block of code is extremely neat and easily readable and understood. I have just noticed that writing switch-case statements in this way is much neater than writing deeply nested if-else statements that some us are used to.  And the way you abstract out all the try-catch statements into checkExistence has also helped in ensuring the code quality here. Well done! 👍

src/main/java/duke/command/CommandType.java
```diff
@@ -0,0 +1,22 @@
+package duke.command;
+
```
khiangleon, line 2: Good use of enums!

khiangleon:
Overall, neat and clean code is observed in a well-structured manner. You have practiced using SLAP along the way, which makes your code really readable. I have actually learnt a lot from your code!

I think one possible way to further improve your product would be to continue developing more test cases and error handling for user inputs.

Fantastic job! Keep up the good work! 👍

PR #294, opened 2019-08-30
[YongKuan] Duke Increments

SendorasRight:
PR for Duke.


kenneth-fung:
Great work. Despite the differences in our implementations of some levels, I was able to follow the thought process behind much of your code quite quickly thanks to its well thought out structure and classes. Here are a few things that I found noteworthy:

- I like that the user can decide the file path for storage. However, might there perhaps be other ways for them to go about setting the file path? Currently it is set through a command line argument (if I'm understanding it correctly), which I feel is not very intuitive both for the user and as a developer reading the code. On the user side, there might be cases where they try to add multiple command line arguments, possibly resulting in an undesirable user experience when the app does not behave as they expect. Developer-wise, for me there was a moment of confusion when I saw that Duke's constructor was taking the argument ```arg[0]```; this might be notable as nowhere else in your code did I experience that. As a suggestion, if you intend to keep this feature, perhaps it could be done within the app itself with a new Command? Especially since in future iterations there will be a need to do a proper UI through JavaFX, although with the UI you could also decide to discard the feature and just use the default file path (which is fine by me too).

- On the topic of storage, I quite liked that you managed writing and reading to the file through input/output streams, particularly in just a few lines of code. I was unaware it could be done in this manner at all, and it is nice to learn something new.

- The following point is more subjective. When looking at the implementation for ```ByeCommand```, I was surprised to see it threw an exception instead of returning normally. As it is, it works fine thanks to ```DukeShutDownException```, but may I ask why you chose to do it this way? Personally, I am uneasy about the command to exit the application throwing an exception, since (as I understand it) exceptions are indications of problems that occur while a program is running. However, it does work perfectly, so in the end this is more of an opinion-based thing to consider.

- Finally, I like that DateTime is a class by itself, with its own ```toString()``` method. This is not something I considered doing at any point during development, and after looking through your solution it appears far neater. In fact, that likely goes for your entire code-base; each file is not very long but they all have clear purposes, and the overall structure serves to divide the app into well-functioning components. Good job.

kenneth-fung (commented):
Great work. Despite the differences in our implementations of some levels, I was able to follow the thought process behind much of your code quite quickly thanks to its well thought out structure and classes. Here are a few things that I found noteworthy:

I like that the user can decide the file path for storage. However, might there perhaps be other ways for them to go about setting the file path? Currently it is set through a command line argument (if I'm understanding it correctly), which I feel is not very intuitive both for the user and as a developer reading the code. On the user side, there might be cases where they try to add multiple command line arguments, possibly resulting in an undesirable user experience when the app does not behave as they expect. Developer-wise, for me there was a moment of confusion when I saw that Duke's constructor was taking the argument arg[0]; this might be notable as nowhere else in your code did I experience that. As a suggestion, if you intend to keep this feature, perhaps it could be done within the app itself with a new Command? Especially since in future iterations there will be a need to do a proper UI through JavaFX, although with the UI you could also decide to discard the feature and just use the default file path (which is fine by me too).

On the topic of storage, I quite liked that you managed writing and reading to the file through input/output streams, particularly in just a few lines of code. I was unaware it could be done in this manner at all, and it is nice to learn something new.

The following point is more subjective. When looking at the implementation for ByeCommand, I was surprised to see it threw an exception instead of returning normally. As it is, it works fine thanks to DukeShutDownException, but may I ask why you chose to do it this way? Personally, I am uneasy about the command to exit the application throwing an exception, since (as I understand it) exceptions are indications of problems that occur while a program is running. However, it does work perfectly, so in the end this is more of an opinion-based thing to consider.

Finally, I like that DateTime is a class by itself, with its own toString() method. This is not something I considered doing at any point during development, and after looking through your solution it appears far neater. In fact, that likely goes for your entire code-base; each file is not very long but they all have clear purposes, and the overall structure serves to divide the app into well-functioning components. Good job.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,130 @@
...
+        case "todo":
+            return Parser.parseTodoCommand(parameters);
+        case "event":
+            return Parser.parseEventCommand(parameters);
+        case "deadline":
+            return Parser.parseDeadlineCommand(parameters);
+        case "find":
+            return Parser.parseFindCommand(parameters);
+        default:
+            throw new DukeException("☹ OOPS!!! I'm sorry, "
```
ooimingsheng: consider declaring the command word  as a final static String

e.g. public final static COMMAND_WORD = "bye"

case ByeCommand.COMMAND_WORD

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,179 @@
...
+     bars. Newline is not added to input string.
+     Currently not used.
+
+     @param output  The string to be printed
+    */
+    private void formattedPrint(String output) {
+        System.out.println("____________________________________________________________\n"
+                + output
+                + "____________________________________________________________\n");
+    }
```
ooimingsheng: Might want to abstract out these "____________________________________________________" into a private final variable. I believe it would make it more readable

src/main/java/duke/command/NewTaskCommand.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public void execute() throws DukeException {
+        this.taskList.add(this.task);
+        this.ui.displayMessage("Got it. I've added this task:\n  "
+                + task
+                + "\nNow you have "
+                + this.taskList.getSize()
+                + " task(s) in the list.");
+        this.storage.saveToDisk(this.taskList);
+    }
```
ooimingsheng, line 41: might want to put the "\n" character on the line above. It is more intuitive the see the end of line character at the end of the line in my opinion.

ooimingsheng (commented):
Overall, it was quite well done. Please see the suggestions below for code specific comments.

PR #295, opened 2019-08-31
[Chan Dalis] Duke Increments

dalisc:


src/main/java/Duke.java
```diff
@@ -1,10 +1,177 @@
...
+            tasks = new TaskList();
+        }
+    }
+
+    public void run() throws IOException {
+        this.ui.showLogo();
+        Scanner sc = new Scanner(System.in);
+        boolean chat = true;
+
+        while (chat) {
```
jcjjjared: chat is a boolean and should sound like one. So maybe isChatting would be a better boolean name? :)

jcjjjared (commented):
Greetings! Very nice code. 😄 Your Java documentation looks very thorough and descriptive.
Maybe your code could be better organised by splitting the commands for the run method in the Main class into different "Command" classes though. 👍 This will keep the run() method in your Duke class shorter and easier to read. Keep up the good work!! 💯 🌵 💃

src/main/java/Duke.java
dalisc: You're right! Thanks for the catch!

crazoter:
Hi dalisc, as jared has mentioned your code is clean and meets the criteria easy. I also love your commit messages. I can see that the javafx portion is still a WIP, so keep up the good effort (don't forget to modify the getResponse method to suit your app's purposes if you haven't already.

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,16 @@
...
+import duke.handler.Storage;
+import duke.ui.Ui;
+
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class ByeCommand extends Command {
+    @Override
+    public void execute(ArrayList<Task> tasks, Ui ui, Storage storage) throws IOException {
+        response = "Bye! Hope to see you again soon.";
```
lumos309: Should be handled by `UI`.

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param fullCommand The full string input of the user
+     * @return An object that is a child class of Command
+     * @throws DukeException DukeException is thrown if the command is not valid
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        assert !fullCommand.equals(""): "The command is empty.";
+        fullCommand = fullCommand.trim();
+        String[] commandWords = fullCommand.split("\\s+");
+        String commandType = commandWords[0];
+        if (commandType.equals("todo")) {
```
lumos309, line 24: This entire block might look neater if handled by a `switch` block instead. Each case could also be abstracted out.

src/main/java/duke/command/SortCommand.java
```diff
@@ -0,0 +1,79 @@
...
+    String by;
+
+    public SortCommand(String by) {
+        this.by = by;
+    }
+
+    @Override
+    public void execute(ArrayList<Task> tasks, Ui ui, Storage storage) throws DukeException {
+        if (by.equals("deadline")) {
+            ArrayList<Deadline> deadlines = new ArrayList<>();
```
lumos309, line 22: The sorting logic could be abstracted out .

src/main/java/duke/handler/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+            BufferedReader reader = new BufferedReader(new FileReader(this.filePath));
+            String read;
+            while ((read = reader.readLine()) != null) {
+                String type = read.substring(1, 2);
+                String status = read.substring(4, 5);
+                String info = read.substring(7);
+
+                Task task;
+                if (type.equals("T")) {
+                    task = new ToDo(info.trim());
```
lumos309, line 63: See above comment in `Parser`.

src/main/java/duke/handler/TaskList.java
```diff
@@ -0,0 +1,34 @@
+package duke.handler;
```
lumos309, line 1: Wherever the add/delete methods are, they should be in this class instead.

src/main/java/duke/handler/TaskList.java
```diff
@@ -0,0 +1,34 @@
...
+     * @param list The list of tasks that already existed.
+     */
+    public TaskList(ArrayList<Task> list) {
+        this.list = list;
+    }
+
+    /**
+     * Constructs a TaskList instance with no tasks added to the list yet.
+     */
+    public TaskList() {
```
lumos309, line 24: Default constructor not needed.

crazoter:
I've run through your code again and as before, the code is relatively clean (lumos has already pointed out areas of improvement). Perhaps you may want to add some test cases?

Looking forward to the user guide and application release.

PR #296, opened 2019-08-31
[Keith Teo] Duke Increments

keiteo:


src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param description description of the task
+     * @param by deadline of the task
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        try {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+            time = LocalDateTime.parse(by, formatter);
+        } catch (Exception err) {
```
dorcastan: Might it be better to catch the specific type of exception thrown, rather than Exceptions in general?

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+                String taskToBeFound = fullCommand.substring(5);
+                return new FindCommand(taskToBeFound);
+            default:
+                throw new DukeException("     ☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+        } catch (StringIndexOutOfBoundsException e) {
+            throw new DukeException("     Please enter the description!");
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new DukeException("     Please enter the time/date!!");
+        } catch (Exception e) {
```
dorcastan, line 57: Do you think it'd be useful to catch NumberFormatExceptions separately? That could help the user to figure out what's wrong if any of the `Integer.parseInt` calls fails!

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Represents a parser object to process commands.
+ */
+public class Parser {
+
+    /**
+     * Parses user commands.
+     * @param fullCommand command from user input
+     * @return a Command
+     * @throws DukeException if the command in invalid
```
dorcastan: Not sure if you've already spotted this, but I think there's a small typo: "if the command *is* invalid" 😆

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,107 @@
+package duke.util;
+
+import duke.exception.DukeException;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+import java.io.File;
```
dorcastan, line 9: Separating the import statements like this made it easier to read! 👍

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+     * Writes data to file.
+     * @param taskList ArrayList of Task objects to be written
+     * @throws DukeException if file is not found
+     */
+    void writeToFile(ArrayList<Task> taskList) throws DukeException {
+        StringBuilder toSave = new StringBuilder();
+        for (Task task : taskList) {
+            toSave.append(task.saveData().append("\n"));
+        }
+        try {
```
dorcastan, line 73: Just wondering, is there a reason why you decided to throw an IOException? I chose a different approach by creating the required directories/files if the file doesn't exist ([link to code](https://github.com/dorcastan/duke/blob/e461af616853b0bdbab91020ccb1ff86f0c1abbe/src/main/java/duke/HardDiskStorage.java#L80-L83)), but I wanted to know if there are trade-offs in that approach.

src/test/java/TaskListTest.java
```diff
@@ -0,0 +1,21 @@
...
+import duke.util.Ui;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+class TaskListTest {
+
+    @Test
+    void deleteTask_noSuchTask_exception() {
```
dorcastan: Would you like to standardize the name of this method? The other test methods end with exceptionThrown, though both ways are clear to read.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+        return new StringBuilder("D|").append(super.saveData()).append("|").append(by);
+    }
+
+    /**
+     * Returns a string representation of this task.
+     * @return a string representation of this task
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + time + ")";
```
dorcastan, line 56: The format of `time` can also be controlled using a [format string](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html#syntax)! This is just a suggestion; I personally find the default LocalDateTime output not as easy to understand. 😆

src/main/java/Duke.java
```diff
@@ -1,10 +1,60 @@
...
+ * This is the main class of the program.
+ */
 public class Duke {
+
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Initialises the Duke object.
```
dorcastan: If I'm not wrong, the CS2103T coding standard includes a line break between the JavaDoc description and parameter tags!

dorcastan (commented):
Nice work! The clean code and concise JavaDoc comments for all public classes and methods helped with readability.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param action type of task - either deadline or event
+     * @param description description of the task
+     * @param time time of task
+     * @throws DukeException if the action is neither "event" nor "deadline"
+     */
+    public AddCommand(String action, String description, String time) throws DukeException {
+        super(false);
+        if (action.equals("event")) {
+            task = new Event(description, time);
+        } else if (action.equals("deadline")) {
```
gabrielchao, line 40: There are definitely advantages to combining the Event and Deadline creation into one class, such as reducing the amount of boilerplate code two separate classes would require. However, if the requirements for Event and Deadline ever changed, they might have to be separated, so perhaps preempting this and making DeadlineCommand and EventCommand classes might be a good idea?

src/main/java/duke/task/Deadline.java
gabrielchao: Indeed, the `|` operator can be used to catch multiple exceptions, as in
```suggestion
        } catch (IllegalArgumentException | DateTimeParseException e) {
```

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+        this.filepath = filepath;
+        storage = new File(filepath);
+    }
+
+    /**
+     * Loads data from the file.
+     * @return ArrayList of Task objects
+     * @throws DukeException if file is not found
+     */
+    public ArrayList<Task> load() throws DukeException {
```
gabrielchao, line 51: Minor suggestion, would it better implement 'programming to an interface' methodology if the code worked with the base interface rather than the implementing class wherever possible? i.e.
```suggestion
    public List<Task> load() throws DukeException {
```

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,108 @@
...
+    public void showLine() {
+        System.out.println(line);
+    }
+
+    /**
+     * Shows the message when user deletes a task.
+     * @param task deleted task object
+     * @param numOfTasks number of tasks left on the list
+     */
+    void showDeleted(Task task, int numOfTasks) {
```
gabrielchao: On one hand, putting all the output logic inside the `Ui` class makes it so only one class has to be changed when dealing with UI issues. On the other hand, this tightly couples the `Ui` class with the `Command` classes, making them highly dependent on this class. Just some food for thought, for me as well!

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public TaskList find(String taskToBeFound) throws DukeException {
+        ArrayList<Task> searchResults = new ArrayList<>();
+        for (Task task : taskList) {
+            if (task.getDescription().contains(taskToBeFound)) {
+                searchResults.add(task);
+            }
+        }
+        if (searchResults.isEmpty()) {
+            throw new DukeException("     File not found! :(");
```
gabrielchao, line 97: Hmmm, since the extra spaces at the start are a specific UI requirement, would it be better to somehow add the indentation with a method at the `Ui` or higher class level instead?

gabrielchao (commented):
Overall well written and thoughtful code. Good job!

src/main/java/Duke.java
keiteo: Thanks for pointing it out, will get it fixed!

src/main/java/duke/util/Parser.java
keiteo: Thanks for pointing it out, will get it fixed!

src/test/java/TaskListTest.java
keiteo: Oh yeah it's better to standardise it, thanks!

PR #297, opened 2019-08-31
[Park Ye Won] Duke Increments

yewon0303:


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Parses the input by user into Command, and other relevant details
+     * such as the deadline, event date, or delete task index.
+     *
+     * @param fullCommand entered by the user
+     * @return Command
+     */
+    public Command parse(String fullCommand) {
```
Wingedevil, line 23: This function is very long, perhaps it is better if it is better abstracted?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.FileWriter;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    public static final String DEFAULT_STORAGE_PATH =
+            "/Users/Yewon/Documents/NUS/Y2S1/CS2103T/iP/duke/src/main/java/dukeTaskList.txt";
```
Wingedevil: Maybe a relative file path will be better?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,63 @@
+package duke.task;
+
+public class Task {
+    public String type; //Todo, Deadline, Event
+    public String task;
```
Wingedevil, line 5: This member is public while you supply a accessor, is it intentional?

kerwin97 (commented):
May i ask why the change from private to public for TextUi, Storage and Tasklist in Duke.java? Also perhaps you could comment on what Command.java execute() method does. Overall I do understand all your code and what it does! It looks very clean and simple and object oriented! Perhaps you could add in more JavaDoc Comments I think that would be great!

PR #299, opened 2019-08-31
[Gabriel Chia] Duke Increments

gachia:
iP Duke Increments


src/main/java/AddCommand.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Represents the Command for adding Tasks and its subclasses.
+ * A subclass of Command.
+ */
+public class AddCommand extends Command {
+
+    private String taskCmd;
+    final String timePattern = "d MMMM yyyy, h:mma";
+    DateTimeFormatter dateTimeFormat = DateTimeFormatter.ofPattern(timePattern);
```
mrnewguy: Good use of the java DateTimeFormatter! However, I think that you could try implementing your own DateTimeFormatter class too!

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,79 @@
...
+                throw new DukeException("☹ OOPS!!! The description of an event cannot be empty.");
+            }
+            userWords = taskName.split("/at");
+            if (userWords.length == 1) {
+                throw new DukeException("☹ OOPS!!! The date/time of an event cannot be empty or is wrongly typed.");
+            }
+            tasks.addTask(new Event(userWords[0].trim(), userWords[1].trim()));
+            break;
+        default:
+            throw new DukeException("Wrong type of command entered" + taskCmd);
```
mrnewguy: Nice use of switch case statements, keeps code clean. However, does this imply that the default case is always an "exception"?

src/main/java/Ui.java
```diff
@@ -0,0 +1,80 @@
+import java.util.Scanner;
+
+/**
+ * Represents the User Interface of Duke, including the user input fields.
+ */
+public class Ui {
```
mrnewguy, line 7: In general, I think the abstractions are well done (e.g.  showError, showGoodbye). It makes the code readable and more importantly highly extensible

src/main/java/AddCommand.java
mrnewguy: Good use of the java DateTimeFormatter! However, I think that you could try implementing your own DateTimeFormatter class too!

mrnewguy (commented):
Overall well done. Code is clean and well-abstracted. I especially like the use of switch cases and a "finally" block to throw errors. However, I am not sure whether this is the convention or whether there is a better way to do it.

src/main/java/AddCommand.java
gachia: Yes, a default command isn't a recognisable command by Duke and should throw an exception.

brianyenna (commented):
Good code overall! Clear and concise. But perhaps there could be more separation of concerns. For instance, in the switch-case statement used in Duke.java, perhaps the handling of each of the individual commands could be abstracted away to a separate method call dedicated to that method. There are also some minor coding style violations (mainly pertaining to the inconsistent spacing between the curly braces and surrounding code)

src/main/java/Duke.java
```diff
@@ -1,163 +1,48 @@
...
-                        System.out.println(lineSpace + "Nice! I've marked this task as done:\n"
-                                + list.get(taskNo - 1) + "\n" + lineSpace);
-                        saveData(filePath, list);
-                        break;
-                    default:
-                        throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
-                }
-            } catch (Exception e) {
-                System.out.println(e.getMessage());
+                String fullCommand = ui.readCommand();
```
brianyenna, line 64: Perhaps a possible alternative implementation is to get the Parser to parse the entire input, not just the command, and allow the Parser to decide which are the parameters to pass on to the various Command objects. Hence, the different Command objects have constructors that take in the remaining parameters (such as the date and time for the DeadlineTask). This way, there is no need to make further UI calls within the Command objects, and the Command object can instead be used solely for decision-making logic. Perhaps we can debate over the merits of either approach :)

brianyenna (commented):
Great code overall. Gave me some good ideas as well :)

src/main/java/Duke.java
gachia: Your alternative implementation seems cleaner. Mine was a hack-ish way of getting the data part of the input since the Scanner object was inside the UI class.

PR #302, opened 2019-08-31
[Manas Vegi] Duke Increments

ManasVegi:


src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,115 @@
...
+        return time;
+    }
+
+    /**
+     * Prints time of the deadline in a user-friendly readable format.
+     * @return string representing time of the deadline in a particular format.
+     */
+    public String printTime() {
+        String jargon = "";
+        switch (time.get(Calendar.DAY_OF_MONTH) % 10) {
```
hellodommy: Following this logic, 11, 12 and 13 would become 11st, 12nd and 13rd which is incorrect.
I made this mistake as well! 😞

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,198 @@
...
+     */
+    public Calendar getEndTime() {
+        return endTime;
+    }
+
+    /**
+     * Returns a string representing startTime in a particular user-friendly format.
+     * @return string representing the startTime in a printing friendly format.
+     */
+    public String printStartTime() {
```
hellodommy, line 69: Would it be possible to move this to `Task` class so it does not have to be repeated in `Event` and `Deadline` classes?

src/main/java/duke/command/Ui.java
```diff
@@ -0,0 +1,118 @@
+package duke.command;
+
+import duke.task.Task;
+import java.util.ArrayList;
+
+/**
+ * Interacts with user by providing messages when required.
+ */
+public class Ui {
+    public static String hr = "______________________________________________________________________";
```
hellodommy, line 11: This could be declared as `final`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,140 @@
...
+                        } catch (DukeException | IOException de) {
+                            ui.showException(de);
+                        } catch (NumberFormatException nfe) {
+                            ui.showNumberFormatError("done");
+                        }
+                    } else if (command.equals("bye")) {
+                        ui.showGoodBye();
+                        break;
+                    } else if (command.equals("list")) {
+                        ui.showTasks(tasks);
```
hellodommy: Sorry, I don't understand why is `ui.showTasks(tasks)` called twice 😅

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    
+    /**
+     * Returns a string representing endTime in a particular user-friendly format.
+     * @return string representing the endTime in a printing friendly format.
+     */
+    public String printEndTime() {
+        String jargon = "";
+        switch (endTime.get(Calendar.DAY_OF_MONTH) % 10) {
+        case 1 : jargon = "st ";
```
hellodommy: Following the recommended [Java Coding Standard](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html) for switch statements, this line should be
`case 1:`
`jargon = "st "`

hellodommy (commented):
Good job on implementing the OOP and packaging 😊
The `run()` method in `Duke` could be cleaner by moving the parsing to other relevant classes. The numerous nested `try` and `if` blocks could be easily prone to error.
I think the code example for Week 3 project under A-MoreOOP Stretch Goal is a good guide to structure `Duke`.

```
public void run() {
    ui.showWelcome();
    boolean isExit = false;
    while (!isExit) {
        try {
            isExit = c.isExit();
        } catch (DukeException e) {
            ui.showError(e.getMessage());
        } finally {
            ui.showLine();
        }
    }
}
```

tanlk99 (commented):
**General comments**

- I've noticed that your classes are quite broad in their function, e.g. `Duke` executes the command that is passed to it from `Parser`. There is also some code duplication mainly in `Deadline` and `Event`, which makes maintaining code more difficult. If you don't mind, I would like to suggest a couple of classes you could abstract out:

   - `Executor`: executes commands parsed by `Parser`, called from `Duke`
   - `DateParser`: has two functions, `parseTime` which takes a `String` and outputs a `Calendar`, and `printTime` which takes a `Calendar` and outputs a `String`. I can see this being useful in simplifying `Parser`, `Storage`, `Event` and `Deadline`.

- You might also want to look into using `.gitignore` to automatically manage repository files (e.g. might not want to include `.class` files). It's quite easy to use; all you have to do is specify which files to exclude from your commits.

**Specific comments**
- Using `startTime` and `endTime` for Event is a brilliant idea (events don't occur at a single point in time after all!).

https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/task/Deadline.java#L52-L60
- Nice attention to detail in adding suffixes for the days of the month. However, this logic seems to fail for the 11th, 12th and 13th.

https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/command/Storage.java#L97-L102
- Perhaps it might be better to use polymorphism for this part? To be precise, it might be better to define a `getLetter()` method for the subclasses and call that instead.

https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/task/Event.java#L79-L81
- This method of handling month names certainly does work, but I would suggest using `SimpleDateTimeFormat` to format the month name like this:
```
nameOfMonth = new SimpleDateFormat("MMMM").format(startTime.getTime());
```

https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/Duke.java#L73-L74
- There might be an accidental extra line here.

https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/command/Ui.java#L10
- I'm not sure that setting the horizontal line string to `public static` is a good idea; other classes can easily change it either deliberately or by accident. It might be better to use `private static final` instead.

- This point is honestly quite minor and can sound like nitpicking, but I feel that ArrayList should be spelled in proper upper case in Javadoc.

src/main/java/duke/task/Deadline.java
ManasVegi: I just figured that! I've corrected it now. Thanks :)

src/main/java/duke/task/Event.java
ManasVegi: I did not move it to Task because ToDo tasks do not have time so having a printTime method for it's superclass might not make lot of sense. 
Would that matter?

src/main/java/duke/command/Ui.java
ManasVegi: Oh right! Thanks!

src/main/java/Duke.java
ManasVegi: Oh shoot! Thanks for finding the typo!

ManasVegi:
> **General comments**
> 
> * I've noticed that your classes are quite broad in their function, e.g. `Duke` executes the command that is passed to it from `Parser`. There is also some code duplication mainly in `Deadline` and `Event`, which makes maintaining code more difficult. If you don't mind, I would like to suggest a couple of classes you could abstract out:
>   
>   * `Executor`: executes commands parsed by `Parser`, called from `Duke`
>   * `DateParser`: has two functions, `parseTime` which takes a `String` and outputs a `Calendar`, and `printTime` which takes a `Calendar` and outputs a `String`. I can see this being useful in simplifying `Parser`, `Storage`, `Event` and `Deadline`.
> * You might also want to look into using `.gitignore` to automatically manage repository files (e.g. might not want to include `.class` files). It's quite easy to use; all you have to do is specify which files to exclude from your commits.
> 
> **Specific comments**
> 
> * Using `startTime` and `endTime` for Event is a brilliant idea (events don't occur at a single point in time after all!).
> 
> https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/task/Deadline.java#L52-L60
> 
> * Nice attention to detail in adding suffixes for the days of the month. However, this logic seems to fail for the 11th, 12th and 13th.
> 
> https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/command/Storage.java#L97-L102
> 
> * Perhaps it might be better to use polymorphism for this part? To be precise, it might be better to define a `getLetter()` method for the subclasses and call that instead.
> 
> https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/task/Event.java#L79-L81
> 
> * This method of handling month names certainly does work, but I would suggest using `SimpleDateTimeFormat` to format the month name like this:
> 
> ```
> nameOfMonth = new SimpleDateFormat("MMMM").format(startTime.getTime());
> ```
> 
> https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/Duke.java#L73-L74
> 
> * There might be an accidental extra line here.
> 
> https://github.com/nus-cs2103-AY1920S1/duke/blob/65e72448f52a4ffec4838fa96f8449a66399ee78/src/main/java/duke/command/Ui.java#L10
> 
> * I'm not sure that setting the horizontal line string to `public static` is a good idea; other classes can easily change it either deliberately or by accident. It might be better to use `private static final` instead.
> * This point is honestly quite minor and can sound like nitpicking, but I feel that ArrayList should be spelled in proper upper case in Javadoc.

Thanks a lot for such detailed explanation! Really helpful!

You are right about me having to abstract a lot of code. I'll do it soon. I'll try implementing the classes you mentioned.

I've just corrected the 11th, 12th... problem. Thanks for figuring it out.

Will definitely try simpleDateFormat!

You are right. Should make horizontal line private.

No worries! Nitpicking is important in coding! Will change it in JavaDocs. Thanks a lot:)

PR #304, opened 2019-08-31
[Abhinav Ramnath] Duke Increments

ambhinav:


src/main/java/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public static Command parse(String fullCommand) throws DukeException {
+        Command c = null;
+        String[] commandArr = fullCommand.split(" ");
+        String[] keywords = {};
+        int i; // skip the task type
+        int taskId;
+
+        try {
+            switch (commandArr[0].toLowerCase()) {
```
bigjunnn: I like how you used the switch statements here, makes the code look much cleaner 👍

src/main/java/DeleteCommand.java
```diff
@@ -8,13 +8,13 @@ public DeleteCommand(int taskId) {
...
-        if (taskId > tasks.size()) {
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        String str;
+        if (taskId > tasks.getSize()) {
             throw new DukeException("Please choose a task within the list");
         } else {
-            Task toDelete = tasks.remove(taskId - 1);
-            str = "Noted. I've removed this task:\n" + " " + toDelete.toString() + "\nNow you have " + tasks.size()
+            Task toDelete = tasks.deleteTask(taskId - 1);
+            str = "Noted. I've removed this task:\n" + " " + toDelete.toString() + "\nNow you have " + tasks.getSize()
```
bigjunnn: I think this might just be a careless mistake. 
Remember to indent the line breaks by 8 spaces and you should place the line break **before** an operator.

Example:
    "\nNow you have "
        + tasks.getSize() ... 

Hope this helps!

bigjunnn (commented):
I liked how you packaged your files. It's organised neatly into the different folders 😄

src/main/java/utils/Parser.java
```diff
@@ -11,6 +11,12 @@
 public class Parser {
     private static int counter;
 
+    private static final String ERROR_EMPTY_INPUT = "☹ OOPS!!! I'm sorry, but I don't know what that means :-(";
```
bigjunnn, line 22: Nice way to organise the messages!

bigjunnn (commented):
Hi Abhinav, I think your code is quite readable and very well-organised. 

Here are some things I liked about your code:
1.  You abstracted away many of the re-used codes into methods / constants, which is great!
2.  The code was well-organised into the different packages such as utils, tasks etc.
3.  Adherence to the coding standards

Some minor improvements for you:
- For longer methods, maybe you could add comments in between your code, to help others understand your code faster and better.

Overall, I think your code quality is quite high and all the best for the future enhancements 😆

src/main/java/utils/Ui.java
```diff
@@ -0,0 +1,136 @@
...
+import tasks.Task;
+
+import java.io.InputStream;
+import java.io.PrintStream;
+import java.util.List;
+import java.util.Scanner;
+
+public class Ui {
+
+    private static final String LINE_PREFIX = "|| ";
```
TanZhanMing, line 11: Good use of 'final' Strings. Makes sense now that I think about it; configuring the variables to be immutable as it will never be changed anyway. Good coding practice.

src/test/java/utils/ParserTest.java
```diff
@@ -0,0 +1,122 @@
...
+import exceptions.DukeException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
+
+
+class ParserTest {
```
TanZhanMing, line 17: Colour me impressed. 6 JUnit tests instead of the minimum 3 shows dedication to refining and perfecting your code. The tests are also modular in nature for your 'ParserTest' class; testing each of the commands 'add todo', 'add event', 'delete' etc. and also testing for various exceptions. All in all, this is a very comprehensive test set.

src/main/java/utils/Storage.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     *  Takes the stored data and populates the task list.
+     *
+     * @return list of tasks
+     * @throws DukeException if local file not found or invalid date in saved file
+     */
+    public List<Task> readFromFile() throws DukeException {
+        Scanner sc;
+        try {
+            if (file.createNewFile()) {
```
TanZhanMing: I like how you have outputs for all your steps; even for the creation or checking the existence of the 'dukeTest.txt' file as it really aids in debugging.
Just my opinion though, but I think the user does not need to know or see this.
Also, for the javadoc comment up top, it states that a DukeException is thrown if the local file is not found. I feel that it could be extrapolated on slightly as you are immediately creating a new file if it does not exist.
Do correct me if I am wrong on how it works though.

src/main/java/Duke.java
```diff
@@ -1,10 +1,55 @@
...
+    private TaskList tasks;
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Duke duke = new Duke("/home/abhinav/Desktop/2103T/duke/data/dukeTest.txt");
```
TanZhanMing: Sorry to nitpick but I have a minor gripe with the file path. I would recommend using a relative file path to create the 'dukeTest.txt' as if you send the jar executable to other users, they would not have the '/home/abhinav' directory and having the program create a brand new set of directories there would not be that good an idea.

TanZhanMing (commented):
Hello Abhinav,

First of all, I would like to apologise as I had actually commented last week but at the wrong place. I commented at https://github.com/nus-cs2103-AY1920S1/duke/pull/304 instead of here and thus I just copied everything I posted there here. Once again, I am truly sorry for my late review and feedback.

It is evident that you are a much more experienced programmer that takes pride in his work (compared to me at least). Your style of coding is clean and modular in nature, making it easy to follow.

The good points:

Really simplistic code. As the saying goes: 'It's really simple to make something something complicated, but very complicated to make something simple'. Your code is the latter and I am very impressed with it.
You have created a very extensive testing suite that is also modular in nature, being able to test individual methods within a class for multiple classes.
Points for improvement:

Maybe try to change the absolute file path to a relative one in case you need to send the executable file somewhere else.
Maybe intentionally leave a few mistakes here and there so reviewers don't have to look so hard for points for improvements? I'm just kidding.
All in all, really impressive on your part; I feel like I have learned quite alot from this.

PR #305, opened 2019-08-31
[Pin Sern] Duke Increments

fangpinsern:


src/main/java/Duke.java
```diff
@@ -1,10 +1,246 @@
...
+        // LinkedList<Task> li = new LinkedList<Task>();
+
+        //read the existing task from the file and create the Duke.tasks to put
+        //into the list. 
+
+        // try{
+        //     storage.printFileContents(li);
+        // } catch(FileNotFoundException e){
+        //     //System.out.println("File not found");
+        // }
```
stevenwjy: It would be better if you could just remove these unused part of the code instead of just commenting it.

src/main/java/Duke.java
```diff
@@ -1,10 +1,246 @@
...
+            String echo = scan.nextLine();
+            Parser split = new Parser(echo);
+            String error = "";
+
+            String firstWord = split.getType();
+
+            if (firstWord.equals("bye")) {
+                // if bye, then end the program
+                ui.printline();
+                System.out.println("\tBye. Hope to see you again soon!");
```
stevenwjy: Since you already have a `UI` class, you may want to use that class to handle all interactions with user (i.e. handle all input and output). It might be a bit confusing if sometimes you print from `UI` class and sometimes directly from `Duke`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,246 @@
...
+                int taskNum = Integer.parseInt(split.getDesc().get(0));
+                //scan.nextLine();
+                //System.out.println(taskNum);
+                
+                int taskNumb = taskNum - 1;
+                
+                if (taskNumb >= tasks.size()) {
+                    error = "taskDoNotExist";
+                } else if (tasks.getTask(taskNumb).getIsDone()) {
+                    //System.out.println("im here");
```
stevenwjy: It would be to not commit code that is used only for debugging.

src/main/java/Duke.java
```diff
@@ -1,10 +1,246 @@
+import duke.command.AException;
+import duke.command.Parser;
+import duke.command.Storage;
+import duke.command.TaskList;
+import duke.command.Ui;
```
stevenwjy: You may want to consider creating another separate package for `Ui` and moving `TaskList` class  to `duke.tasks` package.

src/main/java/duke/command/AException.java
```diff
@@ -0,0 +1,66 @@
+package duke.command;
+
+/**
+ * Represents every exception that Duke may encounter
+ * when the user uses Duke.
+ */
+public class AException {
```
stevenwjy: You may consider passing a `String` that describes the error when you construct an `AException` and create an object for every exception that you encounter. Hence, every error is described by one exception object instead of just calling a function that directly prints to stdout.

src/main/java/duke/command/Ui.java
```diff
@@ -0,0 +1,106 @@
...
+        printline();
+    }
+
+    /**
+     * Confirmation message when the task is being added into the list.
+     *
+     * @param t task that is being added into the list
+     * @param i number of task that are in the list
+     */
+    public void takeInput(Task t, int i) {
```
stevenwjy: It would be better if you can give a more descriptive name.

stevenwjy (commented):
Overall the code looks nice and it seems like you have put a lot of efforts into it, great job!!

However, here are some feedback that you may want to consider in order to improve the code quality:
- Add newlines at the end of files to make it look nicer when you push to GitHub (prevent it from showing a read symbol that indicates no newline at the end of a file).
- Avoid mixing all the logic in one place (i.e.`run` function inside `Duke` class), it might be better to create a separate class for each command that has its own execution function.
- Don't put too many responsibilities for your Duke class (i.e. handle input/output, parsing different types of command, and handle command execution), you may want to break those functionalities into multiple classes.
- It would be better to have a class that handles all input/output. This will also help you in a debugging process.
- It would be better if you ignore the .class files when you push your code to GitHub.

src/main/java/Duke.java
```diff
@@ -1,10 +1,362 @@
...
+    
+    private Ui ui;
+    private Storage storage;
+    private TaskList tasks;
+
+
+    /**
+     * Creates a Duke object that takes in a filepath to
+     * the file where the list of previous tasks is stored.
+     * @param filePath path to where the last saved task list is stored
```
KartikeKhandelwal, line 40: Leave line here

src/main/java/Duke.java
```diff
@@ -1,10 +1,362 @@
...
+    }
+
+    //write to file when there is a change in state
+
+    /**
+     * Runs the logic for Duke.
+     */
+    public String run(String input) {
+        try {
+//        tasks = new TaskList(new LinkedList<Task>());
```
KartikeKhandelwal: remove unused code

src/main/java/Duke.java
```diff
@@ -1,10 +1,362 @@
...
+        try {
+            storage.writeFile(li);
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        }
+    }
+    
+    
+//    public static void main(String[] args) {
+//        new Duke("duke.txt");
```
KartikeKhandelwal: again, remove

src/main/java/Duke.java
```diff
@@ -1,10 +1,362 @@
...
+                handleUserInput();
+            }
+        });
+
+        dialogContainer.heightProperty().addListener((observable) -> scrollPane.setVvalue(1.0));
+
+        stage.setScene(scene);
+        stage.show();
     }
 }
```
KartikeKhandelwal, line 302: For run function, need to add abstract out the parsing to your parser class, it's too long.  Too many newline comments. 
Abstract out error handling, do not add to the main.

src/main/java/Duke.java
```diff
@@ -1,10 +1,362 @@
...
+        }
+
+    }
+
+    //write to file when there is a change in state
+
+    /**
+     * Runs the logic for Duke.
+     */
+    public String run(String input) {
```
KartikeKhandelwal, line 57: run itself can be split into multiple functions within the class

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,64 @@
...
+     * This constructor keeps the state of the task whether
+     * is it done or incomplete.
+     *
+     * @param description a description given for the task
+     * @param by the date and time which the task needs to be
+     *           completed by
+     * @param doner the completion status of the task
+     *              (Completed is 1, incomplete is 2)
+     */
+    public Deadline(String description, Date by, int doner) {
```
KartikeKhandelwal, line 35: doner can be renamed to isDone
and if the status can only be 1/2, change to Boolean type.

src/main/java/duke/utils/AException.java
```diff
@@ -0,0 +1,66 @@
+package duke.utils;
```
KartikeKhandelwal, line 1: should be a subclass of exception. 
for multiple exception, should subclass DukeException class
Rename the class to more meaningful names such as DukeException

KartikeKhandelwal (commented):
Needs improvement in structure and abstraction.

PR #306, opened 2019-08-31
[Eshwar Kamalapathy] Duke Increments

ghost:
Week-2 Increments


src/main/java/DateTime.java
```diff
@@ -0,0 +1,32 @@
...
+public class DateTime {
+
+    protected String input;
+    protected Date date;
+
+    public DateTime (String input){
+        this.input = input;
+    }
+
+    public void setDateAndTime() throws DukeException {
```
sianghwee, line 20: Nice use of the build in Java library. Will definitely consider adding that to my code. However, I am curious as to whether this should be done in the constructor or have a separate function? The errors should be thrown if the input is already wrong in the first place.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Deadline(String description, DateTime by) {
+        super(description);
+        this.by = by;
+        super.changeToFileFormat('D', description, by);
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + "(by: " + by + ")";
```
sianghwee, line 17: Maybe you can consider using a String.format in this case. For example, _String.format("by: %s" ,by)_

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+
+        try {
+            readFileContents(filePath);
+        } catch (FileNotFoundException e) {
+            System.out.println("File not found!!!");
+        } catch (DukeException e) {
+            System.out.println("OOPS!!! Invalid time and date format read.");
+        }
+
+        while (!byeFlag) {
```
sianghwee: Why is there a need for this while loop when you already have another while loop in your duke() method?

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+    }
+
+    public static void updateRemove(Task task) {
+        System.out.println("Noted. I've removed this task:");
+        System.out.println("    " + task);
+
+        if (taskList.size() > 1) {
+            System.out.println("Now you have " + taskList.size() + " tasks in the list.");
+        } else {
+            System.out.println("Now you have " + taskList.size() + " task in the list.");
```
sianghwee: Instead of writing the same sentence twice, you can just do something like that to make the code neater:

```
String task = "task";
if (taskList.size() > 1) {
    task = "tasks";
}
System.out.println(String.format("Now you have %d %s in the list", taskList.size(), task);
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+
+    public static void readFileContents(String filePath) throws FileNotFoundException, DukeException {
+        File f = new File(filePath); // create a File for the given file path
+        Scanner s = new Scanner(f); // create a Scanner using the File as the source
+        while (s.hasNext()) {
+            readAsObject(s.nextLine());
+        }
+    }
+
+    public static void readAsObject(String item) throws DukeException {
```
sianghwee: If this method is only used in this class, perhaps you can considering making it a private method instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+        while (!byeFlag) {
+            try {
+                duke();
+            } catch (DukeException ex) {
+                System.err.println(ex);
+            }
+        }
+
+        try {
+            writeToFile(filePath);
```
sianghwee: Base on my understanding of your code (I did not have the chance to run it so my understanding might not be right), the writeToFile() method only occurs when one exits the program. However, base on my understanding of the project requirement, the writeToFile() should be done whenever there is a change to the list i.e. when there is an add or delete command. Perhaps, we have different understanding of the project requirement?

sianghwee (commented):
Hello, good job thus far. Just a little comment, you can add the *.class file extension to your .gitignore so that your class files will not be uploaded since they are binary files that can be compiled by the users anyway. It helps to clean up your repository by quite a bit.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,32 @@
...
+        SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
+
+        try {
+            date = formatter.parse(input);
+        } catch (ParseException e) {
+            throw new DukeException("OOPS!!! Invalid time and date format. Please enter in dd/MM/yyyy HHmm format");
```
qiujiaaa: the difference in caps "dd/MM/yyyy HHmm" might actually be a little messy. Can consider other ways to display it, or just simply use all caps/no caps at all.

src/main/java/Duke.java
qiujiaaa: Not sure about the logic behind this loop. I feel that it might better if the exceptions are caught and handled in the duke() method, instead of handling it here in this loop.

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
```
qiujiaaa: Can consider using System.err.println() instead when it comes to printing errors.

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+
+    public static void duke() throws DukeException {
+        String input = sc.nextLine();
+
+        while (!input.equals("bye")) {
+            String command = getCommand(input);
+            String description;
+            DateTime dateTime;
+
+            switch (command) {
```
qiujiaaa: There should not be any indentation for the case blocks!

qiujiaaa: Not sure if you have reached the Parser level yet, but it will definitely make your code neater if you can handle the deciphering of inputs in the Parser class! Also, having different classes for the Commands will make your code neater and more readable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+            readAsObject(s.nextLine());
+        }
+    }
+
+    public static void readAsObject(String item) throws DukeException {
+        char type = item.charAt(0);
+        String[] split = item.split(" \\| ");
+        boolean isValidTask = false;
+
+        switch (type) {
```
qiujiaaa: Indentation for switch case blocks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+    public static void writeToFile(String filePath) throws IOException {
+        FileWriter fw = new FileWriter(filePath);
+        for (Task task : taskList) {
+            String text = task.getFileFormat();
+            fw.write(text + System.lineSeparator());
+        }
+        fw.close();
+    }
+
+    public static void readFileContents(String filePath) throws FileNotFoundException, DukeException {
```
qiujiaaa: Maybe it would be better for you to catch the FileNotFoundException in this method and throw a DukeException instead? In this case you do not have to catch 2 types of exception in the method that calls this method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,264 @@
...
+            for (int i = 0; i < taskList.size(); i++) {
+                int index = i + 1;
+                System.out.println(index + "." + taskList.get(i));
+            }
+        } else {
+            System.out.println("You have no tasks in the list.");
+        }
+    }
+
+    public static void done(String position) throws DukeException {
```
qiujiaaa: The method names of these few methods, done, event, deadline etc are actually quite confusing and I had to read further downwards to understand the usage context of these methods. It might be better if the method names were more specific or if they are handled separately in other classes.

qiujiaaa (commented):
Hi, overall the bulk of the code is in the Duke class, more OOP can be shown in the code. Currently, the Duke class might be a bit messy to read, hence it might be easier if you can create more classes to separate the logic even further. Javadocs will be useful too when we are reviewing your code.

PR #307, opened 2019-08-31
[Jun Ren] Duke Increments

chanjunren:


src/main/java/Commands/AddCommand.java
```diff
@@ -0,0 +1,32 @@
+package Commands;
```
Icesiolz: Might want to follow the coding standards, package names should be in all lowercase to avoid conflict in the names of classes or interfaces.

src/main/java/Commands/AddCommand.java
```diff
@@ -0,0 +1,32 @@
...
+
+import Exceptions.DukeException;
+import Storage.Storage;
+import Tasks.Task;
+import Tasks.TaskList;
+import Ui.Ui;
+
+public class AddCommand extends Command {
```
KxxMxxx: You should have a Javadoc to describe what this class represents.

src/main/java/Commands/AddCommand.java
```diff
@@ -0,0 +1,32 @@
...
+    private Task task;
+
+    public AddCommand(Task task) {
```
KxxMxxx: You can add a Javadoc comment starting with "Constructs...".

src/main/java/Commands/Command.java
```diff
@@ -0,0 +1,19 @@
+package Commands;
+
+import Storage.Storage;
+import Tasks.TaskList;
+import Ui.Ui;
+
+public abstract class Command {
```
KxxMxxx: Considering that your abstract class Command only contains abstract methods, you can consider declaring it as an interface instead. What you have done seems to have defeated the purpose of an abstract class.

Just a suggestion, since you declared it as an abstract class, maybe you could've declared the isExit() method to return false by default and only override it in your ExitCommand class. That would remove the redundancy of overriding the method every time.

src/main/java/Parser/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+public class Parser {
+
+    /**
+     * Interprets the user's input and returns the respective command.
+     * @param command user's input string
+     * @return command according to user's input
+     * @throws InvalidDescriptionException empty description of task
+     * @throws InvalidCommandException unknown command
+     */
+    public static Command parse(String command) throws InvalidDescriptionException, InvalidCommandException {
```
KxxMxxx: Your parse method seems a bit verbose, maybe you can employ a switch case instead of using if-else statements to make it clearer.

Also, you can consider abstracting out the cases into individual methods, such as to create a Todo object, to make it less verbose.

KxxMxxx: You should consider catching exceptions as early as possible, throwing your exception outwards instead of handling it when you can will affect your program flow.

src/main/java/Tasks/TaskList.java
```diff
@@ -0,0 +1,80 @@
...
+     * @throws DukeException
+     */
+    public void addNewTask(Task task) throws DukeException {
+        tasks.add(task);
+    }
+
+    /**
+     * Removes a task from the list of tasks and returns it.
+     * @param index index of file to be deleted
+     * @return deleted task
```
KxxMxxx: The Javadoc format doesn't seem correct, you can explain what is being returned by this method.

src/main/java/Ui/Ui.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Prints a message once a task has been added.
+     * @param numOfTasks total number of tasks after a task has been added
+     * @param taskDescription description of the newly added task
+     */
+    public void showAddTaskMsg(int numOfTasks, String taskDescription) {
+        System.out.println("Got it. I've added this task:\n  " + taskDescription);
+        if (numOfTasks == 1) {
+            System.out.println("Now you have " + numOfTasks + " task in the list.");
```
KxxMxxx: Instead of using a + operator to concatenate your strings, you can consider using the String.format() method to attain the correct format.

KxxMxxx (commented):
I think your code logic is pretty good! However, you can improve the readability of your code by abstracting out into individual methods.

src/main/java/commands/DeleteCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public DeleteCommand(int taskIndex) {
+        super();
+        this.taskIndex = taskIndex;
+    }
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) {
+        Task deleted = taskList.deleteTask(taskIndex);
+        ui.showDelTaskMsg(deleted.toString(), taskList.getListSize());
```
xinmingzh: Is there a need for deleted.toString() as compared to simply passing deleted and printing out deleted in the ui?

KxxMxxx: You're right, there's no need to call toString() here as it would be implicitly called upon.

src/main/java/commands/Command.java
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Updates the task list, text document file and displays the relevant relevant messages
+     * according to the user's command
+     * @param taskList consolidated list of tasks
+     * @param ui prints the relevant messages according the user's commands
+     * @param storage updates the text document file
+     */
+    public abstract void execute(TaskList taskList, Ui ui, Storage storage);
+
+    public abstract boolean isExit();
```
xinmingzh: Why not return false for isExit() such that you only need to overwrite this method once in ByeCommand?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+     * @return command according to user's input
+     * @throws InvalidDescriptionException empty description of task
+     * @throws InvalidCommandException unknown command
+     */
+    public static Command parse(String command) throws InvalidDescriptionException, InvalidCommandException {
+        String[] splitString = command.split(" ");
+        String taskType = splitString[0];
+        if (command.equals("list")) {
+            return new ListCommand();
+        } else if (splitString[0].equals("done")) {
```
xinmingzh: Why not use taskType instead of splitstring[0] since you already assigned it?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+            int taskIndex = Integer.valueOf(splitString[1]) - 1;
+            return new DoneCommand(taskIndex);
+        } else if (command.equals("bye")) {
+            return new ByeCommand();
+        } else if (taskType.equals("todo")) {
+            String[] todoSplit = command.split("todo ");
+            if (todoSplit.length == 1 || todoSplit[1].isEmpty()) {
+                throw new InvalidDescriptionException("\u2639 OOPS!!! The description of a todo cannot be empty.");
```
xinmingzh: Why not check for correct desc of todo when initializing todo object ie. in todo constructor? Since its a exception arising from trying to create a todo object.

xinmingzh: Same for the other object creation related exceptions.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Loads the tasks stored in the file and returns them as a list of tasks.
+     * @return list of tasks stored in the text document
+     * @throws DukeException err i not v sure
+     */
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> tasks = new ArrayList<>();
+        try {
+            String content = new String(Files.readAllBytes(Paths.get(filePath)), "UTF-8");
```
xinmingzh, line 33: You can use Scanner object to read file ie. Scanner sc = new Scanner(new File(filePath)); , which gives you access to nextLine() and hasNext() methods.

PR #309, opened 2019-08-31
[Yen Peng] Duke Increments

Yen-Peng:


ryanYtan:
## Code

It may be helpful to split the main class into multiple subclasses - e.g `TaskList` being it's own class to encapsulate. It will be helpful when doing File I/O, returning String representations of the `Tasklist` in the desired format, etc.

There seems to be an extra `Task` array declaration on line 9 of `Duke.java`, is this needed in the code?

It may be better to use a `boolean` variable such as `isExit` inside the main `while` loop instead of using `while (true) {}` and using `System.exit(0)` within the `switch...case`.

The `try` block is huge, it would be better to break it down to smaller functions for better maintainability and readability, it might also be better to handle incorrect events (i.e when user inputs a wrong command) via exceptions.

Good inheritance done on the `Task` classes.

Some recomemendations I can give for the code are:

* Encapsulating printing to the terminal window, using `varargs` makes it easier to implement
* The second `try` block may be unnecessary, you might be able to handle all cases within the top-level `switch` block, and then catch multiple exceptions in the highest `try...catch` (the one that catches `IOException` from `BufferedReader`).
* Handling each `case` with its own function may be useful, but in general it may be better to abstract parsing of user input into another class (A later increment forces you to do this).

## Style

Coding style was mostly followed, with some exceptions.

* `switch...case`: `case` statements should be on the same level of indentation as `switch`
* In the function `checkErrors(String[] segments)`, `else if` should follow this style, this also applies to the `try...catch` in the main function

```
if (cond) {
    // ...
} else if (cond) {
    // ...
}
```

* Line breaks should occur before operators (such as `+`) for long lines.
* `case` statements do not require braces for each `case` block.

hooncp (commented):
**Duke.java**
- Might be good if the "________________" could be abstracted out into printLine().
- Style formatting for switch in CS2103 (See [Coding standard for CS2103](https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html)):
Same level of indentation for `switch` and `case`
```
switch(userInput) 
case "list":
...

```

**Task**
- Good use of inheritance for child classes

**Exceptions**
- Good use of streams.
- Very cute exception messages 👍 One does not simply write exception messages.

**Others**
- Might want to use more OOP for abstraction of `Duke`.

_Javadocs_
_- To be done._

PR #310, opened 2019-08-31
[Zili] Duke Increments

CowSaysBaa:


bernicechio:
- I appreciate how you used a DateTime class, different packages & handled various exceptions
- Code is clean and concise 👍🏻
- DukeTest.java could have more test methods

briyanii:
I liked how you made a DateTime class to parse dates, and your code was easy to read.

Noticed a couple things that you might have overlooked while parsing user input
1. delimiters while splitting input for separate variables 
While parsing input for adding deadline/event that you only check for "/" and the next 3 characters, so these are valid inputs:
- deadline a /at 12/12/1212 1212
- deadline a /qq 12/12/1212 1212
- deadline a /aaa12/12/1212 1212
- deadline a /   12/12/1212 1212

2. formatting of dates
31/02/1212 1212 is a valid input. 
I recommend taking a took at java's DateTime API.

3. some input can cause Duke to crash
"delete -1" IndexOutOfBoundsException is correctly caught,  but  when input is "done -1" you try to catch ArrayIndexOutOfBoundsException. When "delete a", NumberFormatException is not caught

PR #312, opened 2019-08-31
[He Songchi] Duke Increments

Akimatsu98:


src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,118 @@
...
+            list.add(cnt++, t);
+        } else if (cmd.toLowerCase().startsWith("todo")){
+            if (cmd.length() <= 5) {
+                throw (new DukeWrongTaskException("toDo"));
+            }
+            t = new ToDo(cmd.substring(5));
+            list.add(cnt++, t);
+        } else {
+            throw (new UnknownCmdException());
+        }
```
jyx11011, line 166: Maybe use a Parser class to abstract the parsing process?

src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+
+        String command = p.parseCommand(cmd);
+
+        int index = 0;
+
+        String desc = "";
+
+        switch (command) {
+            case "done":
+                if (cmd.length() <= 5 || parseInt(cmd.substring(5)) >= list.size() + 1) {
```
whneo97: Perhaps can consider catching NumberFormatException and IndexOutOfBoundsException.

eg. done 0 or done exampleText will break the program.

src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+                ArrayList<Task> lst = list.returnAllMatchingTasks(keyWord);
+                if (lst.size() != 0) {
+                    System.out.println("Here are the matching tasks in your list:");
+                    for (Task ta: lst) {
+                        System.out.printf("%d.%s\n", list.getTaskList().indexOf(ta) + 1, ta.toString());
+                    }
+                }
+                return;
+            case "delete":
+                if (cmd.length() <= 7 || parseInt(cmd.substring(7)) >= list.size() + 1) {
```
whneo97: Perhaps can consider catching NumberFormatException and IndexOutOfBoundsException.

eg. delete 0 or delete exampleText will break the program.

src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+        Task t;
+
+        String command = p.parseCommand(cmd);
+
+        int index = 0;
+
+        String desc = "";
+
+        switch (command) {
+            case "done":
```
whneo97: If you happen to be following the coding standard for lecture, "case" shouldn't be indented in a switch-case statement. (this can be configured in settings if you're using Intellij)

Please feel free to let me know if you're using other coding standards that are not from the lecture, would love to know more!

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                if (item.contains("(")) {
+                    int dateDescInd = item.indexOf("(");
+                    desc = item.substring(7, dateDescInd - 1);
+                    date = item.substring(dateDescInd + 5, item.length() - 1);
+                } else {
+                    desc = item.substring(7);
+                }
+            Task t = null;
+            switch (type) {
+                case "[D]" : t = new Deadline(desc, date); break;
```
whneo97: If you happen to be following the coding standard for lecture, "case" shouldn't be indented in a switch-case statement. (this can be configured in settings if you're using Intellij)

Please feel free to let me know if you're using other coding standards that are not from the lecture, would love to know more!

src/main/java/duke/exception/EmptyListException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exception;
+
+public class EmptyListException extends DukeException {
+    public EmptyListException() {
+        super("Your list is currently empty. Try telling me some tasks!");
```
whneo97, line 11: Love that you included a message for the user if the list is empty!

src/main/java/duke/frontend/Duke.java
```diff
@@ -0,0 +1,43 @@
+package duke.frontend;
```
whneo97, line 1: I like how your packages included frontend and backend clearly indicating to code viewers of your program the tasks that your classes are responsible for performing in your program.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                t.markAsDone();
+            }
+            list.add(t);
+        }
+        System.out.println("Loaded save data file successfully.");
+    } catch (FileNotFoundException e) {
+            //Create a new file if no file found
+            File f = new File(filePath);
+            try {
+                if (f.createNewFile())
```
whneo97: I love how you have a means to create new file if it does not exist! I think this is very good for cross-directory or even device compatibility.

src/main/java/duke/frontend/Duke.java
```diff
@@ -0,0 +1,43 @@
...
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+    }
+
+    public static void main(String[] args) throws DukeException {
+        String dir = System.getProperty("user.dir") + "/src/main/java/data/dukeData.txt";
```
whneo97: I love how you did not hardcode the location of your file but instead obtained the file path from operating directory! I think this is very good for cross-directory or even device compatibility.

src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+                    throw (new CompleteTaskException());
+                }
+                index = parseInt(cmd.substring(5));
+                list.get(index - 1).markAsDone();
+                System.out.println("Nice! I've marked this task as done:");
+                System.out.println(list.get(index - 1).toString());
+                return;
+            case "find":
+                String keyWord = p.parseDesc(cmd);
+                ArrayList<Task> lst = list.returnAllMatchingTasks(keyWord);
```
whneo97: I like how you abstracted this method and had a good descriptive method name that tells a viewer of the code what it is doing at first glance, even from the client class where this method is called. It even appears to me when I first saw this almost as if this is a concise stream operation and think this is very good.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    public String parseDesc(String s) {
+        String cmd = parseCommand(s);
+        int index = 0;
+        switch (cmd) {
+            case "deadline":
+                index = s.indexOf("/");
+                return s.substring(9, index - 1).trim();
```
whneo97: Perhaps could consider validating the input before parsing the description.

Any description that is taken in with less than 9 characters will throw StringIndexOutOfBounds exception and break the program, including but not limited to:

deadline /by sdf 
deadline /sdf /by sdf
deadline sdf /by
deadline /by

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public String parseDate(String s) {
+        int index = s.indexOf("/");
+        String rawDate = s.substring(index + 4).trim();
+        SimpleDateFormat sdf = new SimpleDateFormat("dd/MM/yyyy HHmm");
+        try {
+            Date d = sdf.parse(rawDate);
+            return d.toString();
+        } catch (ParseException e) {
+            return rawDate;
```
whneo97, line 80: Perhaps could help the user enter the correct format by validating and displaying an error in Ui instead?

This current variation allows the following variant to be added to the tasklist
event hello /s werh

Program displays:

Got it. I've added this task:
[E][✘] hello (at: erh)
Now you have 5 tasks in the list.

src/main/java/duke/frontend/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+                    try {
+                        if (list.size() == 0) {
+                            throw (new EmptyListException());
+                        }
+                        System.out.println("Here are the tasks in your list:");
+                        for(int i = 0; i < list.size(); i++) {
+                            System.out.println(((i + 1) + ".").concat(list.get(i).toString()));
+                        }
+                        break;
+                    } catch (EmptyListException e) {
```
whneo97: Perhaps could consider other general exceptions as well?

Suppose you have a corrupted data file with [P] as type.

When you initialise Storage.java, your list will end up having one element that is null. When you after which call commands such as "list", your Ui.java will go through each element and apply the toString() method to each Task. That is until it reaches the element containing null. It will then throw a NullPointerException that is not caught, throwing it all the way up to the method until the program breaks.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,82 @@
...
+                }
+            Task t = null;
+            switch (type) {
+                case "[D]" : t = new Deadline(desc, date); break;
+                case "[T]" : t = new ToDo(desc); break;
+                case "[E]" : t = new Event(desc, date); break;
+                default:
+                    System.out.println("Unknown event type encountered.");
+            }
+            if (status.equals("\u2713")) {
```
whneo97: Perhaps could consider non-unicode characters?

At least on my computer, the done attribute is not preserved when the program reloads (i.e. after marking a task as done and exiting the program, all tasks marked as done are reset as undone when the program starts up again)

Does this phenomenon occur on the device you're running the program on too?

PR #315, opened 2019-08-31
[Balam Sai Rohit] Duke Increments

ROHITREDDYBALAM:


src/main/java/Duke.java
```diff
@@ -1,10 +1,82 @@
...
+        parser = new Parser();
+        storage = new Storage(s);
+    }
+
+    /**
+     * Contains the flow of the program execution.
+     * @throws ParseException if date of event class not in specified format.
+     * @throws IOException if file not found or other input output exceptions
+     */
+    public void run() throws ParseException, IOException {
```
Russell-Loh-NUS: Will be better to abstract the commands into separate classes e.g AddCommand, ListCommand etc. Also, parsing of string can be abstracted into a separate class.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+        taskList.add(t);
+    }
+
+    /**
+     * Prints the addition of a task and number of tasks in the list after addition.
+     */
+    protected void addMessage() {
+        System.out.println(" Got it. I've added this task: ");
+        System.out.println(taskList.get(taskList.size() - 1));
+        if (taskList.size() == 1) {
```
Russell-Loh-NUS, line 38: If/else can be omitted here. Simply calling taskList.size() will return the list size from the ArrayList.

src/main/java/Duke.java
```diff
@@ -1,10 +1,82 @@
...
+                String required = scan.next();
+                this.list.find(required);
+            } else {
+                System.out.println(" OOPS!!! I'm sorry, but I don't know what that means :-(");
+                String empt = scan.nextLine();
+            }
+        }
+    }
+
+    public static void main(String[] args) throws DukeException, IOException, ParseException {
```
Russell-Loh-NUS: Will be better to use a try/catch here to print out the related exceptions.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+            System.out.println(" OOPS!!! the description of a event cannot be empty. ");
+        } else {
+            int first = b.indexOf('/');
+            String desc = b.substring(0, first - 1);
+            String on = b.substring(first + 4);
+            DateFormat df = new SimpleDateFormat("dd/MM/yyyy");
+            Date at = df.parse(on);
+            Task t1 = new Event(desc, at, false);
+            this.add(t1);
+            this.addMessage();
```
Russell-Loh-NUS: May be better to abstract all printing of output to the Ui class.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+     *Reads and processes the event given by the user.
+     * Stores the information in the list after processing.
+     *
+     * @param b The information of the event being processed.
+     * @throws ParseException If the data is not in the required format i.e. MM/dd/yyyy HH:mm.
+     */
+    protected void readEvent(String b) throws ParseException {
+
+        if (b.length() == 0) {
+            System.out.println(" OOPS!!! the description of a event cannot be empty. ");
```
Russell-Loh-NUS: Use of DukeException here will be better to print exceptions related to the application.

Russell-Loh-NUS (commented):
Methods can be abstracted further but otherwise, the methods are well documented 👍

src/main/java/TaskList.java
weiijiie: I think he is trying to account for the grammar in the case where the size of the task list is 1, so he does not print out `...have 1 tasks`.

src/main/java/Parser.java
```diff
@@ -0,0 +1,103 @@
...
+     * @throws ParseException if thee data for event type of tasks not in the specified format
+     */
+    public void readUserCommand(Ui ui, TaskList list, Storage storage) throws IOException, ParseException {
+        Scanner scan = new Scanner(System.in);
+        while (scan.hasNextLine()) {
+            String command = scan.next();
+            if (command.equals("bye")) {
+                ui.exit();
+                String writeDataToFile = "";
+                for (Task t : list.taskList) {
```
weiijiie: If the user exits through a method other than typing "bye" (ie. pressing control + c), will the user's tasks still be saved?

src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
...
+    public void run() throws ParseException, IOException {
+        Scanner scan = new Scanner(System.in);
+        this.ui.greet();
+        File f = new File("../Duke/data/Duke.txt");
+        Scanner sca = new Scanner(f);
+        while (sca.hasNext()) {
+            String dat = sca.nextLine();
+            this.parser.readTask(dat,this.list);
+        }
+        this.parser.readUserCommand(this.ui,this.list,this.storage);
```
weiijiie: No need to prefix variables with `this` here when it is unambiguous.

src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
...
+    /**
+     * Contains the flow of the program execution.
+     * @throws ParseException if date of event class not in specified format.
+     * @throws IOException if file not found or other input output exceptions
+     */
+    public void run() throws ParseException, IOException {
+        Scanner scan = new Scanner(System.in);
+        this.ui.greet();
+        File f = new File("../Duke/data/Duke.txt");
+        Scanner sca = new Scanner(f);
```
weiijiie: The variable names are a little confusing, perhaps you can be more explicit about the functionality of each by naming them `userInputScanner` and `fileScanner`.

src/main/java/Parser.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Contains the methods to read the input and process it into the form required for the algorithm.
+     *
+     * @param s The information to be read and processed
+     * @param t The list where the information is stored after being processed
+     * @throws ParseException If date for event is not in the specified format i.e. MM/dd/yyyy HH:mm
+     */
+    public void readTask(String s, TaskList t) throws ParseException {
+        DateFormat df = new SimpleDateFormat("E, MMM dd yyyy HH:mm");
```
weiijiie: It may be better to abstract the code to repopulate your `TaskList` from the file into a method in your `Storage` class. That way if the form of storage changes, it will be relatively easier to just change code in the `Storage` class.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     *Deletes the task in the list upon user's request and prints the number of tasks remaining upon deletion.
+     *
+     * @param y the position of the task in the list
+     */
+    protected void deleteTask(int y) {
+        if (taskList.size() == 0) {
+            System.out.println("The task list is empty");
```
weiijiie: Perhaps this class should not be printing stuff to the Ui, as it may lead to a difficult time refactoring when u want to change to JavaFX for example.

PR #316, opened 2019-08-31
[Khairul Azman] Duke Increments

khairulazman1997:


R-D-D-D:
Good job so far! I do not have much to comment on so I will talk more about the logic side. The main parsing of user input seems to be done in the constructor for command and not in the real parser. Maybe can consider abstract out that part to parser. For AddDeadlineCommand, looping through the array works perfectly but you may also consider just split the string using the delimeter "/by". In this way processing is easier and you can detect mistakes in user's input easier as well. Also, you can consider include more testing cases. Have equals method for command and create some command in the test class and test the parser!

ZiHawkEye:
Great job! I would like to comment on how the input is processed. For example, if I want to add a deadline, it seems like most of the processing doesn't actually occur in Parser but actually in AddDeadlineCommand. Perhaps it would be better if all the input processing was in Parser ie. a clearer separation of the different classes and their roles.

PR #320, opened 2019-08-31
[Alex] Duke Increments

alexwxh:


src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+import java.lang.Exception;
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.charset.StandardCharsets;
+import java.time.LocalDateTime;
+
+class DukeException extends Exception {
```
shawnkoh: It would be better if DukeException was declared in a different file.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+    private ArrayList<Task> todoList;
+    File saveList;
+    FileWriter editor;
+
+    public Duke() {
+        this.todoList = new ArrayList<Task>();
+    }
+
+    private void initialize(){
+        saveList=new File("src/main/data/list.txt");
```
shawnkoh: savelist=new violates the coding style.
Refer to https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
 public class Duke {
+    private ArrayList<Task> todoList;
+    File saveList;
+    FileWriter editor;
+
+    public Duke() {
+        this.todoList = new ArrayList<Task>();
+    }
+
+    private void initialize(){
```
shawnkoh: coding style

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        try {
+            isNew = saveList.createNewFile();
+            if(!isNew){
+                Scanner sc = new Scanner(saveList);
+                this.transferSavedFile(sc);
+                editor = new FileWriter(saveList, true);
+                print("Save file loaded");
+            }
+        }catch(IOException e){
+            print("Directory search error. No save file loaded. Initialized with blank list. No commands will be saved to hard disk");
```
shawnkoh: consider using line breaks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+
+        }
+    }
+
+    private void transferSavedFile(Scanner sc){
+        int index = 0;
+        while(sc.hasNext()){
+            String saveEntry = sc.nextLine();
+            String instruction[] = saveEntry.split(" \\| ");
+            if(instruction[0].equals("T")){
```
shawnkoh: consider using a switch statement.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        }
+        sc.close();
+    }
+
+    private static void print(String message) {
+        System.out.println(">>" + message);
+    }
+
+    private void list() {
+        StringBuilder outputMessageMessage = new StringBuilder("List:");
```
shawnkoh: outputMessageMessage does not communicate the intent of the StringBuilder object.
consider using a clearer name like outputMessageBuilder.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        int i = todoList.size();
```
shawnkoh: i is not descriptive enough.
consider using todoListCount instead.

it is alright to use scratch variables like i, j, k in the context of extremely small scopes like a for loop increment variable, but not really appropriate here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+            } catch (IndexOutOfBoundsException e) { //same as event time
+                Duke.print("Error: no deadline provided");
+            } catch (IOException e){
+                Duke.print("File write error");
+            }
+        }
+    }
+
+    private void raiseSingleWordInput(String badInput) throws DukeException {
+        if (badInput.equals("todo") || badInput.equals("event") || badInput.equals("deadline")) {
```
shawnkoh: consider using a switch statement.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        if (spacePosition == -1) {  //raise exception: only 1 word given
+            try {
+                this.raiseSingleWordInput(userInput);
+            } catch (DukeException e) {
+                Duke.print(e.getMessage());
+            }
+        } else {
+            String command = userInput.substring(0, spacePosition);
+            String taskContent = userInput.substring(spacePosition + 1);
+            if (command.equals("done")) {
```
shawnkoh: consider using a switch statement

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        public String toString() {
+            if (isDone) {
+                return "[" + "\u2713" + "]" + this.description;
+            } else {
+                return "[" + "\u2718" + "]" + this.description;
+            }
+        }
+    }
+
+    public class Todo extends Task {
```
shawnkoh: class Todo should be declared in Todo.java

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        if (badInput.equals("todo") || badInput.equals("event") || badInput.equals("deadline")) {
+            throw new DukeException("Error: no description for task.");
+        } else if (badInput.equals("done")) {
+            throw new DukeException("Error: done task index missing");
+        } else {
+            throw new DukeException("Error: no such command");
+        }
+    }
+
+    public class Task {
```
shawnkoh: class Task should be declared in Task.java

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+            Duke.print(message);
+        }
+
+        public void silentSetDone(){
+            this.isDone = true;
+        }
+
+        @Override
+        public String toString() {
+            if (isDone) {
```
shawnkoh: Consider using a ternary operator.

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+    public class Task {
+        protected String description;
+        protected boolean isDone;
+
+        public Task(String description) {
+            this.description = description;
+            this.isDone = false;
+        }
+
+        public void setDone() {
```
shawnkoh: Consider decoupling the printing of the message. The task should focus on being a task in order to be extensible.

shawnkoh: Also, what if I want to unmark the task as complete?

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+        public String toString() {
+            if (isDone) {
+                return "[T]" + "[" + "\u2713" + "]" + this.description;
+            } else {
+                return "[T]" + "[" + "\u2718" + "]" + this.description;
+            }
+        }
+    }
+
+    public class Event extends Task {
```
shawnkoh: Declare in Event.java

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+            startMinute = Integer.parseInt(timeInput[0])%100;
+            endHour = Integer.parseInt(timeInput[1])/100;
+            endMinute = Integer.parseInt(timeInput[1])%100;
+            this.startDateTime = LocalDateTime.of(year, month, day, startHour, startMinute);
+            this.endDateTime = LocalDateTime.of(year, month, day, endHour, endMinute);
+        }
+        @Override
+        public String toString() {
+            if (isDone) {
+                return "[E]" + "[" + "\u2713" + "]" + this.description + " (at: " + this.startDateTime.toLocalDate()
```
shawnkoh: you can use super.toString() instead of rewriting ```"[" + "\u2713" + "]" + this.description```

src/main/java/Duke.java
```diff
@@ -1,10 +1,387 @@
...
+                        + " " + this.startDateTime.toLocalTime() + "-" + this.endDateTime.toLocalTime() + ")";
+            } else {
+                return "[E]" + "[" + "\u2718" + "]" + this.description + " (at: " + this.startDateTime.toLocalDate()
+                        + " " + this.startDateTime.toLocalTime() + "-" + this.endDateTime.toLocalTime() + ")";
+            }
+        }
+    }
+
+    public class Deadline extends Task {
```
shawnkoh: declare in Deadline.java

shawnkoh (commented):
Your codebase would be a lot better if you adhered to the coding style here
https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html

Also, it would be even better if you declared your classes in separate files. It cleans up your codebase and allows you to track changes made in every file.

Your code is also very coupled. You can check out how I tried to decouple my objects as much as possible here: https://github.com/shawnkoh/duke

src/main/java/Duke.java
```diff
@@ -1,10 +1,41 @@
 public class Duke {
+    private UserInterface ui;
+    private TaskManager tasks;
+    private FileManager fileManager;
+
+    private Duke(String pathname) {
+        this.ui = new UserInterface();
+        ui.printWelcome();
```
pohlinwei: This is an interesting way of printing the greeting! However, would you consider this as a task Duke's constructor should be doing? I feel that the printing is not essential to the creation of a Duke object. Perhaps consider calling this function after Duke is created?

src/main/java/Duke.java
```diff
@@ -1,10 +1,41 @@
...
+        this.tasks = new TaskManager();
+        try{
+            this.fileManager = new FileManager();
+            if(this.fileManager.initialize(pathname, tasks)){
+                ui.printNewFile();
```
pohlinwei: I like how you inform your user of what is going on :D

src/main/java/Duke.java
```diff
@@ -1,10 +1,41 @@
...
+    }
+
+    private void run(){
+        boolean isExit = false;
+        while(!isExit){
+            try {
+                String userCommand = ui.readLine();
+                Command c = Parser.parse(userCommand);
+                c.execute(tasks, ui, fileManager);
+                isExit = c.isExit;
```
pohlinwei: This might not be absolutely relevant in this context. However, what if someone has access to c? Then, s/he would be able to change `c`'s `isExit`value easily. I would suggest wrapping it as a function instead and making `isExit` `private`.

src/main/java/Parser.java
```diff
@@ -0,0 +1,35 @@
+public class Parser {
+    public static Command parse(String fullCommand) throws DukeException{
+        fullCommand = fullCommand.trim();
+        int spacePosition = fullCommand.indexOf(" ");
+        if(spacePosition == -1 && !(fullCommand.equals("list") || fullCommand.equals("bye"))){
+            raiseSingleBadInput(fullCommand);
+        }
+        String[] taskInfo = fullCommand.split(" ", 2);
+        switch(taskInfo[0]){
+            case "todo":
```
pohlinwei: Interesting way of handling the different cases! I wouldn't have thought of doing this. However, according to the style guide for CS2103, you should comment that you are allowing a fall-through case. I feel that this makes it clearer to a less-experienced programmer as well. (:

src/main/java/Command.java
```diff
@@ -0,0 +1,40 @@
+public class Command {
+    String[] info;
+    boolean add;
+    boolean delete;
+    boolean done;
+    boolean isExit = false;
+
+    public Command(String[] info, boolean add, boolean delete, boolean done){
+        this.info = info;
+        this.add = add;
```
pohlinwei: I think that it would be great if you followed how you name your boolean values in `Duke.java`. In particular, using `'isAdd`, `isDelete`, etc. would make the purpose of theses variables clearer for me.

src/main/java/FileManager.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        try {
+            editor.write(task.toFileString());
+            editor.flush();
+        } catch (IOException e){
+            throw new DukeException("File write error");
+        }
+    }
+
+    protected void fileSetDone(int doneIndex) throws DukeException{
```
pohlinwei: Great effort! However, I feel that your code could look neater if you could group the code which handles the state of the task: have the code which creates `<TaskType> | 0 | <TaskDetails...>` and  `<TaskType> | 1 | <TaskDetails...>` placed in the same class.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+    protected Task addNewTask(String[] taskInfo) throws DukeException{
+        if(taskInfo[0].equals("todo")){
+            return this.addTodo(taskInfo[1]);
+        }else if(taskInfo[0].equals("event")){
+            return this.addEvent(taskInfo[1]);
+        }else if (taskInfo[0].equals("deadline")){
+            return this.addDeadline(taskInfo[1]);
+        }
+        throw new DukeException("Something wrong lol");
```
pohlinwei: Haha. I like the humour (:

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private Todo addTodo(String task) {
+        task = task.trim();
+        Todo todo = new Todo(task);
+        this.todoList.add(todo);
+        return todo;
+    }
+
+    private Event addEvent(String task) throws DukeException {
+        int split = task.indexOf(" /at");
```
pohlinwei: This is quite a bit of work. Do you think that it would help if you did the parsing in the `Parser` instead? (This comment applies to `addDeadline` as well) I believe that the `TaskManager` should be concerned managing the task rather than parsing commands.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,111 @@
...
+            } catch (IndexOutOfBoundsException e) {         // happens when input is "event xx /at" with no time given
+                throw new DukeException("No event time provided");
+            }
+        }
+    }
+
+    private Deadline addDeadline(String task) throws DukeException {
+        int split = task.indexOf(" /by");
+        if (split == -1) {
+            throw new DukeException("Deadline not given. Specify deadline using \"/by\"");
```
pohlinwei: It is great that you ensure that the date is entered by the user 👍

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,111 @@
...
+    protected String generateList() {
+        StringBuilder outputMessage = new StringBuilder();
+        int i = todoList.size();
+        for (int x = 1; x <= i; x++) {
+            outputMessage.append("\n  " + x + "." + todoList.get(x - 1));
+        }
+        return outputMessage.toString();
+    }
+
+    protected int getSize(){
```
pohlinwei: I like the abstraction here! (:

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,111 @@
...
+                throw new DukeException("Error: no deadline provided");
+            }
+        }
+    }
+
+    protected void load(Task task){
+        this.todoList.add(task);
+    }
+
+    protected void savedDone(int index){
```
pohlinwei: It might just be me but I feel that this method is not necessary because if you look at `done` method below, you didn't use `saveDone`. Instead, you called `setDone` directly. Nonetheless, it is a good attempt at abstraction 😄

src/main/java/UserInterface.java
```diff
@@ -0,0 +1,69 @@
...
+                "| . (_) `. |'   ( \\.-.|| (_,_)   __ |  (_,_)___| \n" +
+                "|(_    ._) '' (`. _` /||  |\\ \\  |  |'  \\   .---. \n" +
+                "|  (_.\\.' / | (_ (_) _)|  | \\ `'   / \\  `-'    / \n" +
+                "|       .'   \\ /  . \\ /|  |  \\    /   \\       /  \n" +
+                "'-----'`      ``-'`-'' `--'   `'-'     `'-..-'   \n" +
+                "(◕ᴗ◕✿)\n";
+        System.out.println("Hewwo from" + logo);
+    }
+
+    protected void printAdd(Task task, int listSize){
```
pohlinwei: Out of curiosity, is there a reason for using `protected`?

src/main/java/Event.java
```diff
@@ -0,0 +1,75 @@
...
+public class Event extends Task {
+    private LocalDateTime startDateTime;   // DD/MM/YYYY
+    private LocalDateTime endDateTime;
+
+    public Event(String description, String timePeriod) throws DukeException{
+        super(description);
+        this.setDateTime(timePeriod);
+    }
+
+    private void setDateTime(String timePeriod) throws DukeException{
```
pohlinwei: Perhaps it would be a good idea to do the parsing before this (in `Parser` for instance). This way you wouldn't have duplicate code and it saves you some work :D

pohlinwei (commented):
Great job! :D I would say that the code is generally quite organised. However, I feel that some dependencies between classes can be avoided by grouping related code together. In addition, although it is note major, there is some styling issue which can be amended easily. Nonetheless, great effort!

src/main/java/Duke.java
alexwxh: Hi Shawn, thank you for taking the time to review my code i really appreciate it. And also sorry for replying so late I'm really struggling with this iP.
I'll consider your suggestions. I am really behind schedule so I'm sorry you had to review such a messy code.

src/main/java/Duke.java
alexwxh: Hi pohlinwei, thank you for taking the time to review my code i really appreciate it. And also sorry for replying so late. I enjoyed reading your feedback a lot. 
I'll take your suggestions into consideration.

src/main/java/Duke.java
alexwxh: i guess this does make sense. I'll change it next increment thank you.

src/main/java/Command.java
alexwxh: I've changed it in my recent increments! thanks for the suggestion

src/main/java/UserInterface.java
alexwxh: uh. I thought this was the best access modifier in this case. since it would be accessed by other classes but in the same package/folder. should it be something else?

src/main/java/Event.java
alexwxh: ah. I'll look into it thanks!

src/main/java/TaskManager.java
alexwxh: hi! this was added in frustration because this, by right, should never be thrown if the Parser class works as intended 😒

src/main/java/TaskManager.java
alexwxh: i've not thought of that. thanks for pointing it out

PR #321, opened 2019-08-31
[Austin Santoso] Duke Increments

austinsantoso:
Complete tasks up to week 3


data/dukeData.txt
```diff
@@ -0,0 +1,2 @@
+todo | false | sleep
```
lrchema: I don't think this file needs to be committed, since it is dependent on the user input and does not add anything to the actual working of the program.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,20 @@
...
+     * @param ui The Ui responsible for displaying things in the current Duke app.
+     * @param storage The Storage unit used for the current Duke app.
+     * @param newTask The new task that will be added.
+     */
+    public void addTask(TaskList tasks, Ui ui, Storage storage, Task newTask) {
+        tasks.add(newTask);
+        ui.showMessage("Got it. I've added this task:");
+        ui.showMessage(6, newTask.toString());
+        ui.showMessage("Now you have " + tasks.size() + " tasks in the list.");
+        storage.writeList(tasks);
```
lrchema, line 21: Is there a need to rewrite the file with every change to the task list, or is it sufficient to read in once at the start and write the final list after 'bye' is entered?

yuhongtay (commented):
Code follows proper styling format and is well documented. Good idea to use DukeException child classes in the spirit of OOP and organisation. Perhaps DukeException could be an abstract class?
Good use of tests, as well as running Text UI Testing to access overall performance of code.

src/main/java/Storage.java
```diff
@@ -0,0 +1,59 @@
+import java.io.File;
+import java.io.IOException;
+
+import java.util.ArrayList;
+
+/**
+ * Represents a storage class which is responsible for saving and loading to and from the save file.
+ */
+public class Storage {
```
lrchema, line 9: any reason for separating storage, file reader and file writer?

text-ui-test/ACTUAL.TXT
```diff
@@ -0,0 +1,53 @@
+Hello from
```
lrchema, line 1: consider adding this to gitignore, since it is dependent on program output.

lrchema:
I like your class naming conventions, with all classes that inherit from a parent containing the parent class name. Makes for easy reading

PR #322, opened 2019-08-31
[Gabriel Seow] Duke Increments

gabrielseow:


src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+public class DukeException extends Exception{
```
LeeYiyuan: There should be a space between the opening brace here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        printGreetingMessage();
+
+        while(sc.hasNextLine()) {
```
LeeYiyuan: There should be a space between `while` and `(sc.hasNextLine())`.

LeeYiyuan (changes requested):
- Missing a few commits for submissions up to and including those from Week 3.
- Very minor errors in formatting, but considering how gradle has not been set up yet for checkstyle, the formatting is generally good.
- For the current commits, code is very well written, concise and clear. Well done!

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+
+    private String description;
+    private boolean doneStatus;
```
waynefong0401: The boolean variable should be named to sound like booleans.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,12 @@
+public class ToDo extends Task {
+    private String by;
```
waynefong0401: What does this variable mean? It seems redundant to me.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+import java.util.ArrayList;
+
 public class Duke {
-    public static void main(String[] args) {
+
+    private static ArrayList<Task> tasks = new ArrayList<Task>();
+    private static final String indent = "    ";
+
+    public static void main(String[] args) throws DukeException{
+        Scanner sc = new Scanner(System.in);
```
waynefong0401: The code can be more OOP by creating a new ui class to handle the input and output.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
         System.out.println("Hello from\n" + logo);
+
+        printGreetingMessage();
+
+        while(sc.hasNextLine()) {
+            String newTaskLine = sc.nextLine();
+            String[] newTaskSplit = newTaskLine.split(" ");
+            String taskType = newTaskSplit[0];
+
+            switch(taskType) {
```
waynefong0401: The code can be more OOP by creating multiple command classes (addCommand, listCommand etc)

src/main/java/DateTime.java
```diff
@@ -0,0 +1,107 @@
+public class DateTime {
```
ArunBeCoding: you might want to use Java API for this, there is already an existing DateTime feature in Java

src/main/java/Duke.java
```diff
@@ -1,10 +1,337 @@
+import java.util.Scanner;
```
ArunBeCoding: Duke.java is super long; try to split it up into more classes

src/main/java/duke/datetime/DateTime.java
```diff
@@ -0,0 +1,112 @@
+package duke.datetime;
```
waynefong0401, line 1: You can use the java library DateTimeFormatter to replace this class.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,240 @@
+package duke.parser;
```
waynefong0401, line 1: This class is way too long, it would be better if you can divide it into classes. Also, the parser should only parse the input to command rather than creating task object(todo, deadline etc). Such classes can be created in AddCommand. It would be more OOP in this way.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            createNewSaveFile();
+        } catch (DukeSaveFileCorruptedError e) {
+            createNewSaveFile();
+        }
+        return new ArrayList<Task>();
+    }
+
+    public void saveTasks(ArrayList<Task> tasks) {
+        try {
+            new File(saveFilePath).delete();
```
waynefong0401, line 40: You can rewrite the file instead of deleting and creating the file every time. The File class can check if file exists. If it fails to read or the file is missing, then you can create the file.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,37 @@
...
+public abstract class Task {
+
+    protected String description;
+    protected boolean taskIsDone;
+    protected int taskIsDoneState;
+
+    public Task(String description) {
+        this.description = description;
+        this.taskIsDone = false;
+        this.taskIsDoneState = 0;
```
waynefong0401, line 19: You can create a public method String getTaskState() { return taskIsDone ? "0" : "1";} To avoid redundant variable. And the this. at line 11 and line 12 is violating the coding standard I believe. The this. should be removed.

PR #323, opened 2019-08-31
[Neo Si Hao] Duke Increments

neoshh:


calaquende:
# Feedback

### Comments
+ Good and functional code.
+ Code logic works.

### Suggestions
There can be more abstractions and classes to utilize the benefits of OOP more. I recommend using a `TaskList` class to manage the list of tasks in the program. This may be easier to organize code especially in later levels.

The `doTask(String, ArrayList<Task>)` logic flow is abit too long and can be abstracted into more logical functions/methods so that tweaking behavior in the future can be more easier and more concise.

Raikonen:
# Feedback
- Good use of inheritance for Task and Exception classes
- Adheres to the coding standard given
- Good naming for variables throughout the project

# Suggestion
- Use relative path instead of absolute path for the filepath in line 113 of Action.java file would help the code run on other machines better
- As the above reviewer mentioned, it would be good to break the doTask into smaller parts, so as make it easier to debug and neater 
- Might want to consider using switch cases instead of if/else as it makes the code slightly more readable and concise

PR #324, opened 2019-08-31
[Ong Bing Jue] Duke Increments

bjhoohaha:


src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * The <code>AddCommand</code> is the parent class of {@link AddTodoCommand}, {@link AddDeadlineCommand},
+ * {@link AddEventCommand}.
+ */
+public abstract class AddCommand implements Command {
+
+    /**
+     * This is the task to be added by the add command.
+     */
+    protected Task task;
```
niqiukun, line 20: Hi, I like the fact that you have included JavaDoc for protected class variables :) It will really help reviewers to understand your code.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,53 @@
...
+import duke.storage.Storage;
+import duke.task.Task;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * The <code>AddCommand</code> is the parent class of {@link AddTodoCommand}, {@link AddDeadlineCommand},
+ * {@link AddEventCommand}.
+ */
+public abstract class AddCommand implements Command {
```
niqiukun: The `AddCommand` Class is declared abstract but there is no abstract methods to be implemented. Maybe declare `execute` as abstract so that its child classes has to implement it?

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    /**
+     * Returns <code>true</code> if the command is an exit command and <code>false</code> otherwise.
+     * @return <code>false</code>
+     */
+    public boolean isExit() {
+        return false;
+    }
```
niqiukun: Since all child classes of `AddCommand` are not exit command and `isExit` is implemented in the parent class, probably there is no need for the child classes to override this method.

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+import duke.exception.InvalidParameterException;
+import duke.task.Deadline;
+
+/**
+ * The <code>AddDeadlineCommand</code> is created when the user enters <code>"deadline"</code>. The todo command add a \
+ * deadline task into the list of tasks in {@link duke.task.TaskList}. The user interface will display the new deadline
+ * task that is added.
+ */
+public class AddDeadlineCommand extends AddCommand {
```
niqiukun, line 13: An alternative way of implementing this is to have only one class of `AddCommand` and store the type of event as a class variable. But implementing them as child classes is also fine to me :)

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,26 @@
...
+     * @param ui the user interface
+     * @param storage the storage for the tasks
+     */
+    void execute(TaskList tasks, Ui ui, Storage storage);
+
+    /**
+     * Returns <code>true</code> if the command is an exit command and <code>false</code> otherwise.
+     * @return <code>true</code> if the command is an exit command and <code>false</code> otherwise.
+     */
+    boolean isExit();
```
niqiukun: Consider an alternative implementation: you can declare `Command` as an abstract class, and implement `isExit` as follows: 
```Java
public boolean isExit() {
    return this instanceof ExitCommand;
}
```
As a result, you don't need to reimplement this method in all its child classes.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,57 @@
...
+     * This is the index of task to be deleted in the list of tasks.
+     */
+    int index;
+
+    /**
+     * Constructs a new delete command with the specified index of the task to be deleted in the list of tasks.
+     * @param index the index of the task to be deleted in the list of tasks
+     * @throws InvalidParameterException if the index of the task specified is not a number
+     */
+    public DeleteCommand(String index) {
```
niqiukun: An alternative implementation is let the `Parser` handle type conversion and have the integer as a paramater here.

src/main/java/duke/exception/DukeException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exception;
+
+/**
+ * This is the parent class of all <code>RuntimeExcpetion</code> that may occur in the Duke program.
+ */
+public abstract class DukeException extends RuntimeException {
```
niqiukun: If you are expecting to catch and resolve the `DukeException` in your code, consider extending `Exception` instead of `RuntimeException`. Extending from `Exception` will require you to indicate in each method that you are throwing this exception, which is a good code practice. See another explanation [here](https://stackoverflow.com/questions/19857008/extending-exception-runtimeexception-in-java).

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface Command {
+
+    /**
+     * Executes the command.
+     * @param tasks the list of tasks
+     * @param ui the user interface
+     * @param storage the storage for the tasks
+     */
+    void execute(TaskList tasks, Ui ui, Storage storage);
```
niqiukun: If your `DukeException` extends from `Exception`, you will have to indicate `throws ...` here. Please see my comments in the `DukeException` class.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+     * @param task the task to be added into the list
+     * @param size the number of tasks in the list
+     */
+    public void showAddInformation(String task, int size) {
+        System.out.println("     ------------------------------------------------------------");
+        System.out.println("     Got it. I've added this task:");
+        System.out.println("     ➕  " + task);
+        System.out.println("     Now you have " + size + " tasks in the list.");
+        System.out.println("     ------------------------------------------------------------");
+    }
```
niqiukun, line 164: It is good to see that you have different methods for message display for different purposes :)

src/main/java/duke/exception/InvalidCommandException.java
```diff
@@ -0,0 +1,37 @@
+package duke.exception;
+
+/**
+ * This is a runtime exception that occur when the user enters an invalid command.
+ */
+public class InvalidCommandException extends DukeException {
```
niqiukun, line 6: It is good to see that you have multiple child classes for `DukeException` that handle different types of exceptions :)

src/main/java/duke/exception/InvalidCommandException.java
```diff
@@ -0,0 +1,37 @@
...
+        super();
+        this.command = command;
+    }
+
+    /**
+     * Gets the invalid command that is entered by the user.
+     * @return the invalid command
+     */
+    public String getInvalidCommand() {
+        return command;
```
niqiukun, line 34: Note that if you construct the class with `InvalidCommandException()`, `command` may not be initialised when you call this method. My suggestion is to initialise `command` as `""` when the parameter is `null`.

data/duke.txt
```diff
@@ -0,0 +1,3 @@
+todo,false,read books
+deadline,false,eat,22/08/2019 1800
+event,true,cook,25/08/2019 1200
```
niqiukun: You may consider adding this file into your `.gitignore` to prevent it from being tracked by git.

src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,81 @@
...
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                Command c = CommandParser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
+            } catch (InvalidCommandException ice) {
+                ui.showError("Invalid command: " + ice.getInvalidCommand());
```
niqiukun: Consider placing the type of exception in the `Ui` class? You can display different message for different types of exceptions by method overloading.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+        return reader.lines();
+    }
+
+    /**
+     * Saves the stream of lines inside the file.
+     * @param stream the stream of lines to be saved into the file
+     * @throws IOException if an I/O error occurs
+     */
+    public void save(Stream<String> stream) throws IOException {
+        BufferedWriter writer = new BufferedWriter(new FileWriter(filePath));
```
niqiukun: You can use the `FileWriter` to write file directly without the need for `BufferedWriter`. See the following example:
```Java
FileWriter writer = new FileWriter(filePath);
writer.write("This is one line\n");
writer.flush();
writer.close();
```

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Constructs a new deadline task with description and the date where the deadline task is due by that has not been
+     * done.
+     * @param description the description of the deadline task
+     * @param by the date the task is due by
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = DateParser.parse(by);
```
niqiukun: An alternative way of implementation is to keep the type conversion inside the `Parser` class. Hence, the constructor can take `Date` type as parameter directly. The benefit is that all type conversion related exceptions can be caught in the `Parser` class instead of being thrown by constructors of different classes.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,61 @@
...
+    public Deadline(String description, Date by, boolean isDone) {
+        super(description, isDone);
+        this.by = by;
+    }
+
+    /**
+     * Encodes the deadline task to be stored in the storage.
+     * @return a string representation of the encoded deadline task
+     */
+    public String encode() {
```
niqiukun, line 39: Since you used a separate class for decoding the format from files (`FileLineParser`), consider using a seperate class for encoding them as well.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+    public void add(Task task) {
+        taskList.add(task);
+    }
+
+    /**
+     * Removes the task at the specified index from the list of tasks.
+     * @param index the index of the task to be removed
+     * @return a string representation of the task removed to be printed on a user interface.
+     */
+    public String delete(int index) {
```
niqiukun: I would suggest to catch the `IndexOutOfBoundsException` here and throws a new `DukeException` to be caught by upper level class. Because the upper level class, i.e. `DeleteCommand`, should not be implementation dependent and hence should not know that it has to catch `IndexOutOfBoundsException`.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param tasks the list of tasks
+     * @param ui the user interface
+     * @param storage the storage for the tasks
+     * @throws duke.exception.InvalidParameterException if the index is out of range
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        try {
+            String task = tasks.delete(index);
+            ui.showDeletedMessage(task, tasks.size());
+        } catch (IndexOutOfBoundsException aioube) {
```
niqiukun, line 52: I would suggest to catch `IndexOutOfBoundsException` in `TaskList` instead of here because the actual implementation (and hence the need to catch `IndexOutOfBoundsException`) is not revealed to this class. Please see my comments in `TaskList` class.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+        taskList.remove(index - 1);
+        return task.toString();
+    }
+
+    /**
+     * Marks the task at the specified index from the list of tasks as done.
+     * @param index the index of the task to be done.
+     * @return a string representation of the done task to be be printed on a user interface.
+     */
+    public String done(int index) {
```
niqiukun: Same comment as `delete(int index)` method.

src/test/java/duke/parser/DateParserTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+class DateParserTest {
```
niqiukun, line 13: An alternative way to test the parser is to compare the parsed result with an actual instance of `Date` object. See the example below:
```Java
Calendar calendar = Calendar.getInstance();
calendar.set(Calendar.DAY_OF_MONTH, 28);
calendar.set(Calendar.MONTH, Calendar.AUGUST);
calendar.set(Calendar.YEAR, 2019);
calendar.set(Calendar.HOUR_OF_DAY, 23);
calendar.set(Calendar.MINUTE, 59);
calendar.set(Calendar.SECOND, 0);
calendar.set(Calendar.MILLISECOND, 0);
Date date = calendar.getTime();
assertEquals(date, DateUtil.parseDate("28/8/2019 2359"));
```

niqiukun (commented):
Hi Bing Jue, your code is quite well written :) Keep up the good work! I have left several comments in the code with alternative implementations so that you may consider improving it further.

docs/duke/ui/Ui.html
```diff
@@ -0,0 +1,525 @@
+<!DOCTYPE HTML>
```
febee99: You can consider adding these files to .gitignore :)

src/main/java/duke/command/AddDeadlineCommand.java
febee99: Separating each task type into child classes that inherit from AddCommand works well and could be clearer and more comprehensive to the reader, although there might be lots of code repetition. As @niqiukun suggested, you could store the type of event as a class variable. Alternatively, you could store a private Task variable. `private Task t;`and then assign `t = new Deadline/Todo/Event` in your code body afterwards.

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Constructs a new add deadline command that will be executed in the <code>run</code> method of
+     * {@link duke.main.Duke} with the specified line as a parameter.
```
febee99, line 17: Perhaps try to leave an empty line after the description for every JavaDoc comment in adherance to the style guide.

src/main/java/duke/command/AddEventCommand.java
```diff
@@ -0,0 +1,37 @@
...
+     * @param line the line contents of the command passed as a parameter
+     * @throws InvalidParameterException if the line is blank
+     */
+    public AddEventCommand(String line) {
+        super(line);
+        try {
+            String[] arr = super.line.split(" /at ");
+            super.task = new Event(arr[0], arr[1]);
+        } catch (ArrayIndexOutOfBoundsException aioobe) {
+            throw new InvalidParameterException(line.isBlank() ? null : line);
```
febee99: I like how you have individual classes that extend from DukeException! 👍 It makes things much clearer as it specifies exactly what type of exception is being thrown.

src/main/java/duke/command/DoneCommand.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ *  The <code>DoneCommand</code> is created when the user enters <code>"done"</code>. The done command will mark a
+ *  specified task entered by the user as done. The user interface will display the updated information if it is
+ *  successful.
+ */
+public class DoneCommand implements Command {
+
+    /**
+     * The index of the task to be mark as done in the list of tasks.
+     */
```
febee99, line 20: I like how you wrote JavaDoc comments for even private variables! 💯

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Executes the command. This will save the list of tasks in {@link TaskList} into the storage. The storage is the
+     * file path specified by {@link duke.main.Duke} and {@link Storage}.
+     * @param tasks the list of tasks
+     * @param ui the user interface
+     * @param storage the storage for the tasks
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        try {
+            storage.save(tasks.save());
```
febee99: Good idea to only save the tasks after all commands have been ran, rather than saving and updating time after time.

src/main/java/duke/parser/CommandParser.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Parses the string representation of a command into a <code>Command</code> class object.
+     * @param fullCommand the full command to be parse
+     * @return a command
+     * @throws InvalidCommandException if a invalid command is entered by the user
+     */
+    public static Command parse(String fullCommand) {
+        String[] arr = fullCommand.split(" ");
+        String parameter = Arrays.stream(arr).skip(1).collect(Collectors.joining(" ")).trim();
```
febee99: Sleek method to parse the command. 💯

src/main/java/duke/parser/TaskParser.java
```diff
@@ -0,0 +1,20 @@
...
+
+import duke.task.Task;
+
+/**
+ * This is a parser used to parse individual tasks into a string representation that will be saved into the file path
+ * specified in {@link duke.main.Duke} and {@link duke.storage.Storage}.
+ */
+public class TaskParser {
+
+    /**
```
febee99: As this class only contains one method which is parse(), it could perhaps be done away with. You could also consider shifting the parse method into the Task class, where it can be used like `Task.parse()` without having to specify the task in the parameters. :)

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Constructs a new list of tasks from a stream of line. This stream of lines will be parsed by
+     * {@link FileLineParser} to each individual tasks that will be added individually to form a list of tasks.
+     * @param lines the stream of lines to be parsed.
+     */
+    public TaskList(Stream<String> lines) {
+        this.taskList = new ArrayList<>();
+        lines.forEach(line -> {
```
febee99, line 39: Interesting use of Streams here

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+     */
+    public String readCommand() {
+        return sc.nextLine();
+    }
+
+    /**
+     * Prints the current tasks in the list.
+     * @param list the list of tasks to be printed
+     */
+    public void showTable(String list) {
```
febee99: Effort for your formatting 💯

src/test/java/duke/parser/DateParserTest.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    @Test
+    public void parse_fail_with_ExceptionThrow() {
+        Assertions.assertThrows(InvalidDateTimeException.class, () -> DateParser.parse("233555"));
+        Assertions.assertThrows(InvalidDateTimeException.class, () -> DateParser.parse(""));
+        Assertions.assertThrows(InvalidDateTimeException.class, () -> DateParser.parse("11/02/2931"));
+        Assertions.assertThrows(InvalidDateTimeException.class, () -> DateParser.parse("11/02/2931 test"));
+        Assertions.assertThrows(NullPointerException.class, () -> DateParser.parse(null));
```
febee99, line 41: Perhaps you could standardise the exceptions thrown such that if the date is null, InvalidDateTimeException will also be thrown instead of NullPointerException

febee99 (commented):
Hi there, I really enjoyed reading and evaluating your work. Keep it up! I have added several comments in the files committed and I hope that they can come in useful in one way or another.

src/main/java/duke/command/Command.java
bjhoohaha: Thank you. I think this implementation is sweet and has implemented it in my duke!

PR #327, opened 2019-09-01
[Abdul Haliq] Duke Increments

AHaliq:
Completed `Level-1` to `Level-6`, `A-TextUiTesting` and `A-Enums`.


src/main/java/com/core/State.java
```diff
@@ -0,0 +1,31 @@
+package com.core;
+
+import java.util.ArrayList;
+import com.tasks.DoableTask;
+
+/**
+ * An instance of the State class is a state of the running program.
```
cambrian-dk, line 7: Perhaps a different word other than "State" can be used to make your program more readable(maybe Duke, or Bot). For first-time readers, it may not be immediately clear that a State should contain an ArrayList of tasks in the English meaning of the word.

src/main/java/com/util/Printer.java
```diff
@@ -0,0 +1,127 @@
+package com.util;
+
+public class Printer {
+
```
cambrian-dk, line 4: good use of final variables to abstract out magic numbers

src/main/java/com/core/Response.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.stream.IntStream;
+
+import com.util.Printer;
+import com.tasks.Deadline;
+import com.tasks.DoableTask;
+import com.tasks.Event;
+import com.tasks.Todo;
+import java.util.stream.Stream;
+
+public enum Response {
```
cambrian-dk, line 15: Maybe it may be better to split up the responses into command and error classes, so that in the future if more functionalities get added, it reduces the code from getting too bloated.

src/main/java/com/tasks/DoableTask.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * get string representation of task.
+     *
+     * @return string
+     */
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + super.toString();
+    }
```
tohcejasmine, line 46: Nice idea of overriding the toString() method to help with the printing of tasks 👍

src/main/java/com/util/json/JsonArray.java
```diff
@@ -0,0 +1,73 @@
+package com.util.json;
+
+import com.util.Printer;
+import java.util.ArrayList;
+
+public class JsonArray extends ArrayList<JsonValue> {
```
tohcejasmine, line 6: Maybe a JavaDoc comment on the role of JsonArray class?

src/main/java/com/core/Response.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * Insert task into list and prints message string.
+     *
+     * @param t task to be added
+     * @param s state object
+     */
+    private static void addTask(DoableTask t, State s) {
+        s.list.add(t);
```
tohcejasmine, line 279: Another way would be to abstract this step (same for deleting, marking a task done etc) out and let the State object handle modifications to the list of tasks; this would help Response focus on the console outputs.

tohcejasmine (commented):
Really nice use of enumerations and handling json files, your code was logical and easy to follow :-)

PR #328, opened 2019-09-01
[Wong Chuan Kai] Duke Increments

wongchuankai:
Duke


src/main/java/ByeCommand.java
```diff
@@ -0,0 +1,36 @@
+/**
+ * Exceute the bye command in Duke.
+ */
+
+public class ByeCommand extends Command {
+
+    /**
+     *  Bye Constructor with command string input.
```
choongyx, line 8: Could delete a space before "Bye" to ensure that the comments are aligned as seen in example.
image

src/main/java/ByeCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param command user string input
+     */
+    public ByeCommand(String command) {
+        super(command);
+    }
+
+    /**
+     * Execute the bye command
```
choongyx: Header could start with "executes".
image

src/main/java/Command.java
```diff
@@ -0,0 +1,30 @@
+/**
+ * Execute the command in Duke.
+ */
+public class Command {
```
choongyx: The command class could be an abstract class. Methods such as execute do not need to be fully described and left for the child classes to define since each of them exhibit different behaviour.

src/main/java/time.java
```diff
@@ -0,0 +1,104 @@
+/**
```
choongyx, line 1: An interesting way to format the Date and Time. Could consider using the `DateFormat` class to make the formatting simpler with `DateFormat dateFormat = new SimpleDateFormat("dd/MM/yyyy HHmm")`

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,62 @@
...
+     * Execute the Done command
+     * @param task Tasklist data
+     * @param ui Ui interfaces and strings
+     * @param storage stored file
+     * @throws DukeException if any of raw values are invalid
+     */
+    @Override
+    public void execute(TaskList task, Ui ui, Storage storage) throws DukeException {
+        try {
+            String[] donearr = command.split(" ");
```
choongyx: Variables should be in camelCase.

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,50 @@
...
+        ArrayList<Task> tasklist = task.getTaskList();
+
+        String[] strarr = command.split(" ");
+        String search = strarr[1];
+        System.out.println("     Here are the matching tasks in your list:");
+
+        boolean none = true;
+
+        for(int i = 0; i <tasklist.size();i++) {
+            Task t = tasklist.get(i);
```
choongyx, line 40: t is not an iterator, a more descriptive name could be given.

src/main/java/Parser.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Parses user input into command for execution.
+     * @param command user string input
+     * @return the command based on string input
+     */
+    public Command parse(String command) {
+        if (command.startsWith("todo ")) {
+            ToDoCommand c = new ToDoCommand(command);
+            return c;
```
choongyx, line 18: Could place `return c` at the end of the if-else block instead of repeating them after each case.

choongyx (commented):
The Javadocs was generally detailed and well done! Perhaps, could try checking the code by running the checkstyle with gradle for some coding style violations such as variable naming. In general, the code was easy to read. :)

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param ui Ui interfaces and strings
+     * @param storage stored file
+     * @throws DukeException if any of raw values are invalid
+     */
+    @Override
+    public void execute(TaskList task,Ui ui,Storage storage) throws DukeException {
+        ArrayList<Task> tasklist = task.getTaskList();
+
+        String[] strarr = command.split(" ");
+        String search = strarr[1];
```
shiweing, line 35: You might want to check for IndexOutOfboundsException in case user does not provide a keyword.

PR #330, opened 2019-09-01
[Tan Wei Ming, Bertrand] Duke Increments

Berttwm:
A0183875M Tan Wei Ming, Bertrand
Level-9


src/main/java/Duke.java
```diff
@@ -1,10 +1,57 @@
...
+    public static void main(String[] args) throws Exception {
+        new Duke("data/duke.txt").run();
+    }
+
+    /**
+     * Constructs a Duke object which will initialise 1)storage, 2)tasks 3)Ui class..
+     *
+     * @param filePath is the file directory for where duke.txt is saved
+     */
+    public Duke(String filePath) throws Exception, DukeException {
```
podocarp: Exception is bigger than DukeException, can remove.

src/main/java/Util/Storage.java
```diff
@@ -0,0 +1,183 @@
...
+        bw = new BufferedWriter(fw);
+        pw = new PrintWriter(bw);
+    }
+
+    /**
+     * Reads all data from the file filePath
+     *
+     * @return ArrayList of tasks to be stored under TaskList
+     */
+    public ArrayList<Task> readFile() {
```
podocarp, line 61: You can avoid all these by making Task(s) Serializable, then you can use the API functions to dump (and read) the arraylist to disk. This has the benefit that if you add additional members to any Task(s), it will adapt automatically, although it will deprecate the old saves.

src/main/java/Util/Storage.java
```diff
@@ -0,0 +1,183 @@
...
+        fw.close();
+        fw = new FileWriter(new File(this.filePath), true);
+    }
+
+    /**
+     * Write a Todo task into the file from filePath
+     *
+     * @param task the Todo task which will be analysed
+     */
+    public void writeTodo(Task task) {
```
podocarp, line 144: You can actually combine all these into 1 function, or move them into the tasks themselves and extend them. That will reduce code duplication.

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private Parser parser;
+    private TaskList tasksList;
+    private Storage storage;
+
+    /**
+     * To store the several suffixed that will be used in a month starting from
+     * 0 ending at 31. (i.e 2nd, 11th, 23rd)
+     */
+    static String[] suffixes =
```
podocarp, line 27: This is hardcore.

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+                        throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
+                    }
+                    showDeadline(input);
+                    continue;
+                }
+                if(input.substring(0,5).equals("event")) {
+                    try {
+                        input.substring(6,7);
+                    } catch (Exception e){
+                        throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.");
```
podocarp: This is an event, not deadline.

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+                continue;
+            }
+            if((input.length() > 5) && input.substring(0, 5).equals("find ")) {
+                showFind(input);
+                continue;
+            }
+            else {
+                // just to catch wrong input
+                try {
+                    input.substring(0,4).equals("todo");
```
podocarp: Should just add it to the bottom of the switch, anything you don't recognize is bad input and you don't need to check for it explicitly.

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+            String input = scanner.nextLine();
+            if(input.equals("bye")) {
+                showExit();
+                break;
+            }
+            if(input.equals("list")) {
+                tasksList.showAllTasks();
+                continue;
+            }
+            if((input.length() > 5) && input.substring(0, 5).equals("done ")) {
```
podocarp, line 53: instead of counting the strings, you can do a split(" ") instead and match the first (trimmed) word.

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+        String incorrectTime = dateArr[3].substring(0, 5).replace(":", "").trim();
+        return day + " of " + month + " " + year + ", " + incorrectTime + "hours";
+    }
+
+    public void showFind(String input) throws Exception {
+        String toMatch = input.substring(5).trim();
+        TaskList tempTaskList = new TaskList(new ArrayList<Task>());
+        for(int i = 0; i < tasksList.getTaskList().size(); i++) {
+            String desc = tasksList.getTaskList().get(i).getDescription();
+            if((toMatch.length() < desc.length()) && desc.contains(toMatch)) {
```
podocarp: checking < is redundant

src/main/java/Util/Ui.java
```diff
@@ -0,0 +1,244 @@
...
+        tasksList.addTask(todo);
+        storage.writeTodo(todo);
+        System.out.println("  " + todo);
+    }
+
+    /**
+     * shows deadline message when called and performs read on storage
+     * and appends a deadline task in TaskList
+     */
+    public void showDeadline(String input) throws Exception{
```
podocarp: showing and adding should be two separate things. Also you can just add it in first, then call a toString on the task, and then all these functions can be collapsed into one.

src/test/java/Tasks/DeadlineTest.java
```diff
@@ -0,0 +1,19 @@
+package Tasks;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+
+    @Test
+    public void todoTest() {
```
podocarp, line 8: great!

src/main/java/Util/Parser.java
```diff
@@ -0,0 +1,10 @@
+package Util;
+
+import Tasks.Task;
+
+public class Parser {
```
podocarp, line 5: maybe can move the huge conditional that does the testing of input here

podocarp (commented):
LGTM!

src/main/java/Util/Ui.java
Berttwm: Ah I like this suggestion actually!

src/main/java/Util/Parser.java
Berttwm: So UI takes in parser as a variable and inputs queries to parser variable? All my show methods will be in Parser or UI then?

src/main/java/Duke.java
```diff
@@ -44,6 +37,9 @@ public Duke(String filePath) throws Exception, DukeException {
             throw new DukeException("File is empty");
         }
     }
+    public Duke(){
```
hua-lun: Maybe this part can be removed?

src/main/java/Duke.java
```diff
@@ -34,35 +34,30 @@
     private Image user = new Image(this.getClass().getResourceAsStream("/images/DaUser.png"));
     private Image duke = new Image(this.getClass().getResourceAsStream("/images/DaDuke.png"));
 
-//    public static void main(String[] args) throws Exception {
-//        new Duke("data/duke.txt");
-//    }
+    public static void main(String[] args) throws Exception {
+
```
Sam-limyr, line 38: I think main( ) should not be empty, and also shouldn't be throwing an Exception if it's empty.

src/main/java/Duke.java
```diff
@@ -34,35 +34,30 @@
...
-     * @param filePath is the file directory for where duke.txt is saved
      */
-    public Duke(String filePath) throws Exception, DukeException {
-        storage = new Storage(filePath);
+    public Duke() throws Exception{
+        storage = new Storage("data/duke.txt");
         try {
             tasks = new TaskList(storage.readFile());
             ui = new Ui(tasks, storage);
         } catch (Exception e) {
```
Sam-limyr, line 50: DukeException should be extending Exception; also if it's IOException that you're trying to catch, I think you should catch IOException e instead of Exception e?

src/main/java/Duke.java
```diff
@@ -151,14 +130,12 @@ public void start(Stage stage) {
...
-//
-//        userInput.setOnAction((event) -> {
-//            handleUserInput();
-//        });
+
 
         dialogContainer.heightProperty().addListener((observable) -> scrollPane.setVvalue(1.0));
     }
+
+    public String showWelcome() {
```
Sam-limyr: I feel like this method should be in the Ui class instead?

src/main/java/Tasks/TaskList.java
```diff
@@ -32,37 +32,41 @@ public void addTask(Task task) {
      *
      * @param itemNum is the index of the task to be deleted
      */
-    public void deleteTask(int itemNum) {
+    public String deleteTask(int itemNum) {
         assert(itemNum>=0 && itemNum<taskList.size()):"itemNum does not fall within range";
-        System.out.println("Noted. I've removed this task:");
-        System.out.println("  " + taskList.get(itemNum));
+        String toReturn = "";
```
Sam-limyr, line 37: I feel like the string building should occur in the Ui class instead of the TaskList class?

src/main/java/Util/Ui.java
Sam-limyr: I think this section should be placed in the Parser class instead?

Sam-limyr (commented):
Hi; good job on the code - I'm here from tutorial W13 just cos we're supposed to do it during tutorial HAHA

PR #332, opened 2019-09-01
[Sim Khiang Leon] Duke Increments

khiangleon:


minpyaemoe:
According to my observation, the flow of your code is quite logical. Every line of codes is properly indented and formatted such that it looks clean.
Also, I noticed that you have implemented multiple exception classes to handle possible invalid user inputs. That’s great, however, will it be better if there is only one exception class with message string attributes which can be used to handle different invalid user inputs by calling constructor which takes in a string as parameter?

In your codes, comments are also provided for respective methods. That is very useful for readers to understand the idea of those methods. My suggestion is that those comments will be more effective if they follow the java doc comments format which include the documentation of parameters and return value. You may also use java doc comments for java classes as well.

Lastly, you have implemented inheritance and abstraction for your classes which is great!

ryoarmanda:
I see you have developed your codebase with a more object-oriented kind of manner. A pretty good effort as you have split the work to several parts. However, most of the logic is still in the Duke class. It would be better to split the logic into more classes and methods as well.

You noticed and caught some error cases where the input is incomplete or not valid and went on to throw the appropriate exceptions for it. Good eye! We as developers have to always keep in mind that user input is not always going to be correct. My advice in the long-run is to keep discovering these kinds of cases so that you can cover all possible situations from the user side.

Overall, your code is good so far! Lets push on to more increments :)

src/main/java/DateTime.java
```diff
@@ -0,0 +1,128 @@
...
+        case 10:
+            month = "October";
+            break;
+        case 11:
+            month = "November";
+            break;
+        case 12:
+            month = "December";
+            break;
+        }
```
Auxinnn: It is better to have a default method to throw exceptions.

src/main/java/DateTime.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public void reformatYear(String split) {
+        year = split;
+    }
+
+    public void reformatTime(String split) {
```
Auxinnn: This method may seem a bit too long. And there are a few Java classes you may consider [SimpleDateFormat](https://docs.oracle.com/javase/9/docs/api/java/text/SimpleDateFormat.html) / [LocalDateTime](https://docs.oracle.com/javase/9/docs/api/java/time/LocalDateTime.html)

src/main/java/Parser.java
```diff
@@ -0,0 +1,82 @@
...
+
+    String fullCommand;
+
+    private static Command command;
+
+    public Parser(String fullCommand) {
+        this.fullCommand = fullCommand;
+    }
+
+    public static Command parse(String fullCommand) throws DukeException {
```
Auxinnn: This method looks rather long; perhaps it can be shortened by extracting the methods?

src/main/java/Storage.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+            count ++;
+        } else if (item.startsWith("E")) {
+            // split command into 4 parts
+            String[] splitStr = item.split(" \\| ", 4);
+            tasks.add(new Event(splitStr[2], splitStr[3]));
+            if (splitStr[1].equals("1")) {
+                tasks.get(count).setDone();
+            }
+            count ++;
```
Auxinnn: There are duplicated count++, you may want to put it at the end outside of the if-else block.

src/main/java/DateTime.java
khiangleon: I have added it in. Thanks for your suggestion!

src/main/java/DateTime.java
khiangleon: That is really useful, I will look further into it! Thanks for your suggestion!

src/main/java/Parser.java
khiangleon: I agree. I will try it out!

src/main/java/Storage.java
khiangleon: Oh I see. Ok! Thanks!

PR #333, opened 2019-09-01
[Ziheng Xiu] Duke Increments

Cary-Xx:


src/main/java/Duke.java
```diff
@@ -6,5 +11,299 @@ public static void main(String[] args) {
...
+            }
+        }
+    }
+
+    private void level6() {
+        List<Task> taskList = new ArrayList<>();
+        Scanner sc = new Scanner(System.in);
+        while (true) {
+            String s = sc.nextLine();
+            if (s == null) {
```
rongrongrr: It might be better if you use a switch case for this, rather than else-if, especially for future expansions of your code.

src/main/java/Duke.java
```diff
@@ -6,5 +11,299 @@ public static void main(String[] args) {
...
+            } else if (s.contains("delete")) {
+                if (s.substring(6).isEmpty()) {
+                    System.out.println("☹ OOPS!!! You cannot delete an empty entry");
+                    return;
+                }
+                Task task = taskList.get(Integer.parseInt(s.substring(7)) - 1);
+                taskList.remove(task);
+                System.out.println("Noted. I've removed this task: ");
+                System.out.println(task.toString());
+                System.out.println("Now you have " + taskList.size() + ((taskList.size() == 1) ? " task in the list" : " tasks in the list."));
```
rongrongrr: Maybe you can shorten this line so that it is easier for readers to understand? For example, use String.format to format the String message first before printing it.

src/main/java/Duke.java
```diff
@@ -6,5 +11,299 @@ public static void main(String[] args) {
...
+                taskIndex++;
+                System.out.println("Now you have " + taskIndex + ((taskIndex == 1) ? " task in the list" : " tasks in the list."));
+            } else {
+                System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                return;
+            }
+        }
+    }
+
+    private void level6() {
```
rongrongrr: You can try to use different methods for each command rather than combining them into one whole void method for the level. You might want to refer to the Single Responsibility Principle from the S.O.L.I.D principles for more information.

PR #334, opened 2019-09-01
[Wang Xinman (Mandy)] Duke Increments

tirameshu:


duke/src/main/java/Command.java
```diff
@@ -0,0 +1,18 @@
+import java.text.ParseException;
+
+abstract class Command {
+	protected String command;
+	protected String remainingCommand;
```
timothyleong97: It is unclear what the purpose of the remainingCommand variable is.

timothyleong97: It is unclear what the purpose of this remainingCommand variable is. Perhaps a short javadoc comment would help?

duke/src/main/java/AddCommand.java
```diff
@@ -0,0 +1,149 @@
...
+				throw new DukeException("☹ OOPS!!! The description of a " + type + " cannot be empty.");
+			}
+			this.task = loadTodo(isDone, description);
+			break;
+
+		case "deadline":
+			if (description.isEmpty()) {
+				throw new DukeException("☹ OOPS!!! The description of a " + type + " cannot be empty.");
+			}
+			this.task = loadDeadline(isDone, description);
```
timothyleong97: Why did you choose to do additional parsing inside AddCommand instead of handling it in your Parser class?

duke/src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
+public class Deadline extends Task {
+
+    protected String by;
```
timothyleong97: why does this variable need to be protected instead of package-private when you are not subclassing it?

duke/src/main/java/DukeException.java
```diff
@@ -0,0 +1,7 @@
+public class DukeException extends Exception{
+    protected String type;
```
timothyleong97: why is this variable protected instead of private? You are not subclassing it. Same for other protected variables that don't need to be accessed by subclasses.

duke/src/main/java/AddCommand.java
```diff
@@ -0,0 +1,149 @@
+import java.text.SimpleDateFormat;
+import java.text.ParseException;
+
+import java.util.Date;
+
+class AddCommand extends Command {
+	private static SimpleDateFormat formatterIn = new SimpleDateFormat("d/M/yyyy HHmm");
```
timothyleong97: it looks as if your spacing is greater than 4 spaces

duke/src/main/java/AddCommand.java
```diff
@@ -0,0 +1,149 @@
+import java.text.SimpleDateFormat;
```
timothyleong97: It might be good to package your classes into relevant packages such as a Command package, a Components package for your Ui, Parser etc.

duke/src/main/java/AddCommand.java
```diff
@@ -0,0 +1,149 @@
...
+	 * Parses command to create task by type with its details, reused by loading
+	 * @param type of task
+	 * @param isDone
+	 * @param description of task, deadline, details, etc
+	 * @return Task created by command
+	 * @throws DukeException  could be thrown by loading invalid tasks
+	 * @throws ParseException could be thrown by unparseable datetime in loading functions
+	 */
+	
+	public Task createTask(String type, boolean isDone, String description) throws DukeException, ParseException {
```
limhawjia: It could be useful to abstract this out! It would reduce class bloat and could potentially be used elsewhere

duke/src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,22 @@
...
+		ArrayList<Task> list = tasks.getList();
+		if (list.isEmpty()) {
+			throw new DukeException("You have no task to delete!");
+		} else if (remainingCommand.isEmpty()) {
+			throw new DukeException("Which task do you want to delete?");
+		}
+		
+		int index = Integer.valueOf(remainingCommand); // assuming only has number left
+		Task task = list.get(index - 1);
+		list.remove(task);
```
limhawjia: I feel like the deletion should be done by TaskList itself! In case you need to change its implementation, you would not have to refactor this class too

duke/src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,21 @@
...
+	public void execute(TaskList tasks, Ui ui) throws DukeException {
+		ArrayList<Task> list = tasks.getList();
+		if (list.isEmpty()) {
+			throw new DukeException("You have no task to do!");
+		} else if (remainingCommand.isEmpty()) {
+			throw new DukeException("Do include the task number that you have done!");
+		}
+		
+		int index = Integer.valueOf(remainingCommand); // assuming only has number left
+		Task task = list.get(index - 1);
```
limhawjia: I feel like this should be done by TaskList itself too

duke/src/main/java/Duke.java
```diff
@@ -0,0 +1,200 @@
...
+			ui.showError("Something went wrong with the file: " + e.getMessage());
+		}
+	}
+
+    public static void main(String[] args) {
+		new Duke(filename).run();
+    }
+
+	@Override
+	public void start(Stage stage) {
```
limhawjia: Seems like there's something wrong with the indentation

limhawjia (commented):
Gradle build files and tutorial files could be gitignored. Code was neat and easy to understand! There was good separation of concerns but I feel that the roles of some classes could be more well-defined!

PR #336, opened 2019-09-01
[Wang Runding] Duke Increments

R-D-D-D:


khairulazman1997:
Great work overall! I think it's neat the way you used a single AddCommand to add all the Tasks regardless of which Task type was to be added into the TaskList. I couldn't find much but one room for improvements is to possibly change your Task class into an interface or abstract class since it would never be used on its own to prevent it from ever being initialised on its own. All round great work and all the best for the rest of the module!

ZiHawkEye:
Great job! I like the fact that you accounted for a lot of the edge cases in the user input and had detailed error messages to accompany them. I also think that using the Formatter to standardise user input is also an interesting idea.

PR #337, opened 2019-09-01
[Ahmad Hatziq] Duke Increments

AhmadHatziq:


src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+
+import javafx.application.Application;
+import javafx.scene.Scene;
+import javafx.scene.control.Label;
+import javafx.stage.Stage;
+
+/**
+ * Project Duke is a Personal Assistant Chatbot that helps a person to keep track of various things.
+ * storage attribute is a Storage object, which helps read and write data to the text file
+ *
```
lumos309: Comment is not complete- you should either explain all three attributes (storage + tasks + ui), or none at all.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+        ui.showWelcome();
+
+        try {
+
+            // Loads the data from txt file to the TaskList object, tasks.
+            tasks = new TaskList(this.storage.load());
+
+        } catch (FileNotFoundException e) {
+
+            System.out.println(e.getMessage());
```
lumos309: If the data file cannot be found, we could instead inform the user, then create a new and empty one for them.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+
+        }
+
+        // Creates scanner object to handle input.
+        Scanner in = new Scanner(System.in);
```
lumos309: This and all other input could be handled by the `Ui` class, since it makes sense that both `System.out` and `System.in` commands are dealt with by `Ui`.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+        String fullCommand = in.nextLine().trim();
+
+        while (true) {
+
+            taskType = Parser.parseCommand(fullCommand);
+
+            try {
+
+                if (taskType.equals(PossibleTasks.LIST.toString().toLowerCase())) {
```
lumos309: Using `switch` instead of multiple `if-else`s may make for neater code.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+     */
+    public void run() {
+        // Variable initialization. description will hold the description of all tasks.
+        // extraDescription will hold either the
+        // dateTime attribute of Deadline class or location attribute of Event class.
+        String output = "";
+        String taskType = "";
+        String description = "";
+        String extraDescription = "";
+        int taskNum = -1;
```
lumos309: Number of tasks seems like it should be encapsulated within `TaskList` instead of the main class.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+
+                    tasks.addTask(newDeadline);
+
+                    ui.printDeadlineSequence(tasks, newDeadline);
+
+                } else if (taskType.equals(PossibleTasks.EVENT.toString().toLowerCase())) {
+                    // EVENT command.
+
+                    if ((fullCommand.length() < 6)) {
+                        // Input is only "event".
```
lumos309: Examining the length of an input string to determine its type seems to be more the job of the parser. You could make your `Duke` class a lot cleaner by moving these bits into that class.

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param args Main entry point.
+     */
+    public static void main(String[] args) {
+        new Duke("C:\\Users\\hatzi\\Documents\\Sourcetree\\duke\\data\\tasks.txt").run();
+    }
+
+    /**
+     * possibleTasks  is an enumeration of the constant, pre-defined, recognizable commands.
+     */
+    enum PossibleTasks {
```
lumos309: Possible "Tasks" is a slightly confusing name in my opinion, since it can be mistaken for `Task` and its subclasses (`Todo`  etc.). "PossibleCommands" might be a more illustrative name.

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @param filepath Absolute filepath of the text file.
+     * @throws IOException An IOException may occur when trying to write the file.
+     */
+    public void createFile(String filepath) throws IOException {
+        File f = new File(filepath);
+        String data = "";
+
+        // If the file does not exist, create sa new text file.
+        if (f.exists() == false) {
```
lumos309: Alternatively, `if (!f.exists())`.

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+            inputsFromFile.add(scanner.nextLine());
+        }
+
+        for (String input : inputsFromFile) {
+            // A Possible input string is: "D | 0 | CS2103 Ip  | Wed 2359".
+
+            String[] words = input.split("\\|");
+            Boolean isDone = false;
+
+            if (words[0].length() < 3) {
```
lumos309: Since you know that the first line and only the first line of the file will be a header, it might make more sense to read the header, discard it, then enter the main loop, since there is no reason to check for the header every time.

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+                // If condition to avoid reading in the header.
+
+
+                if (words[0].contains("T")) {
+                    // Create a Todo class.
+
+                    if (words[1].contains("1")) {
+                        isDone = true;
+                    } else if (words[1].contains("0")) {
```
lumos309: The `contains("0")` condition is redundant, unless you suspect the status may not be 0 or 1, in which case you should be checking for it elsewhere anyway.

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+                    description = words[2].trim();
+
+                    Todo newTodo = new Todo(description, isDone);
+                    tasks.add(newTodo);
+
+                } else if (words[0].contains("E")) {
+
+                    // Create an Event class.
+
+                    if (words[1].contains("1")) {
```
lumos309: Here you have multiple checks for the same thing (0 or 1). You could condense your code by checking for it just once, outside of the block where you check for the Task type.

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+                        isDone = true;
+                    } else if (words[1].contains("0")) {
+                        isDone = false;
+                    }
+                    description = words[2].trim();
```
lumos309: Same comment for setting the description, since the description is common across all task types.

src/main/java/seedu/duke/task/Deadline.java
```diff
@@ -0,0 +1,148 @@
...
+    public Deadline(String description, String dateTime) {
+        // Sets isDone to the default value, false.
+        // Possible String for dateTime = 2/12/2019 1800.
+        // Sample fullCommand = "deadline cs /by 21/12/2019 0800".
+        super(description);
+        this.dateTime = dateTime;
+
+        if (!dateTime.contains("of")) {
+            // Checks if the String has already been formatted.
+            // If "of" is present, this indicated that dateTime has already been formatted.
```
lumos309, line 31: What if the user enters a dateTime like "5th of December"?

src/main/java/seedu/duke/task/Deadline.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Returns a parsed dateTime String.
+     * String "by" must be of the format DD/MM/YYYY HHHH Eg. "31/08/2019 1215".
+     * Will be converted to "31st of August 2019, 12.15pm".
+     *
+     * @param by Unparsed dateTime String.
+     * @return Parsed dateTime String.
+     */
+    public String parseBy(String by) {
```
lumos309: Java has built-in methods for doing this: see http://tutorials.jenkov.com/java-internationalization/simpledateformat.html

src/main/java/seedu/duke/task/Task.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Returns a Task object after initializing with the String description.
+     * By default, isDone is set to false.
+     *
+     * @param description Description String of the task.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        this.taskType = PossibleTaskTypes.DEFAULT;
```
lumos309, line 38: Will the user ever create a default task? If not, this may not be needed.

src/main/java/seedu/duke/task/Task.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Returns a Task object after initializing with the String description.
+     * By default, isDone is set to false.
+     *
+     * @param description Description String of the task.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
lumos309, line 37: The `this` quantifier is not needed here.

lzw12345:
Hi , I like the coding style overall as it is neat with good comments and little arrow heads. It looks very clean!

src/main/java/seedu/duke/storage/Storage.java
```diff
@@ -0,0 +1,234 @@
...
+            String[] words = input.split("\\|");
+            Boolean isDone = false;
+
+            if (words[0].length() < 3) {
+                // If condition to avoid reading in the header.
+                String taskType = words[0];
+
+                switch (taskType){
+                case "T":
+                    // Create a Todo class.
```
joannasara: This method is quite long so you might want to extract the handling of each case.

src/main/java/seedu/duke/task/Task.java
```diff
@@ -0,0 +1,175 @@
...
+        // return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
+    }
+
+    /**
+     * Setter function for setting a Task isDone from false to true.
+     */
+    public void setDone() {
+        if (!this.isDone) {
+            this.isDone = true;
+            updateLastModified();
```
joannasara, line 93: Wouldn't it be better to put the "this.lastModified = LocalDate.now();" here instead of calling a new function since its only one line and you don't use it anywhere else? It also looks nicer and comply with the SLAP principle

PR #338, opened 2019-09-01
[Mei an] Duke Increments

meiannn:


ckb055:
Hello, good job on the main class. Perhaps you could extract classes to handle the code in a clearer manner. Such as the Ui class to take in user-input and output, Storage class to store your list of tasks.. etc. Press on :)

calvincxz:
Hi, very interesting approach for this project! I think the logic of the code is generally correct, but your Parser class does not seem to help as much as it still returns a list of strings. Maybe you can transfer the code in your Duke class to the Parser class such that the parse method simply prints the output. This way, you can improve the readability of the Duke class. If you want more OOP, the parse method can return a command object which then prints the output :)

PR #339, opened 2019-09-01
[Philip Alexander] Duke Increments

philipalexanderb:


src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+     * Parse a one line input into a Task or Command.
+     * Print message is produced every time an input is passed in.
+     *
+     * @param input  Command to be executed or Task to be created.
+     * @throws DukeException  if Task has incomplete description.
+     */
+    public void parseLineInput(String input) throws DukeException, IOException, ParseException {
+        String firstWord = input.split(" ")[0];
+        //ArrayList<Task> list = taskList.getList();
+        if (firstWord.equals("bye")) {
```
lye-jw: Your logic is smooth and easy to understand, well done! However, do you think it would be tidier and more readable to use switch statements instead of a large number of switch statements?

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Parse a one line input into a Task or Command.
+     * Print message is produced every time an input is passed in.
+     *
+     * @param input  Command to be executed or Task to be created.
+     * @throws DukeException  if Task has incomplete description.
+     */
+    public void parseLineInput(String input) throws DukeException, IOException, ParseException {
```
lye-jw: You handled Exceptions caused by various possible faulty inputs quite comprehensively, this should ensure that your program runs as it is supposed to even if users make mistakes 👍 👍.

src/main/java/duke/command/Parser.java
lye-jw, line 60: Your logic is smooth and easy to understand, well done! However, do you think it would be tidier and more readable to use switch statements instead of a large number of if-else statements?

src/main/java/duke/command/Storage.java
```diff
@@ -0,0 +1,143 @@
...
+        this.list = taskList.getList();
+    }
+
+    /**
+     * Parse the Text from the File to the corresponding Task objects.
+     * @param taskText the Line from the text to be parsed into Task object.
+     * @param list The list to which the resulting parsed text wish to be added.
+     * @throws ParseException if format of the text does not fit the expected format.
+     */
+    public void parseTextToTask(String taskText, ArrayList<Task> list) throws ParseException {
```
lye-jw, line 72: Would Parser be a more appropriate class to handle the processing and adding of Tasks to the list in the program, while Storage only handles the operations on the external save file?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,63 @@
+package duke.task;
+
+/**
+ * Represents a Task that contains the fields description that describes the
+ * task, isDone that indicates if a task has been done and typeOfTask that
+ * differentiates the type of task which can be an Event task, Todo task, or
+ * Deadline task.
+ */
+public class Task {
```
lye-jw: Good job on the logical use of inheritance.

lye-jw (commented):
Well-structured code overall

ghost (commented):
Dear Philip,
it's great to see that you did a decent chunk of refactoring.
But, would it be possible to do consider refactoring run() in duke.java?
for each conditional block, i think the lines of code can be written as a function to enhance readability.

The above comment applies to parseTextToTask().

Thanks!

PR #340, opened 2019-09-01
[Yoon Jia Jun Ken] Duke Increments

Xelyion:


KendrickAng:
Great work! Duke works completely as specified. I can see the effort you have put into defining a better UX than required. The listing of tasks when Duke is opened is a great touch!

Some minor corrections: The javadocs are comprehensive, but the comment syntax seems to diverge from standard. Perhaps you could add in gradle's checkstyle capabilities, and run that as a gradle task to spot errors - there's a tutorial for that in the tutorials folder!

Your code seems to clumped together into a few files - you could achieve clearer abstraction by following what is suggest in the A-MoreOOP goal:

```
public void run() {
    ui.showWelcome();
    boolean isExit = false;
    while (!isExit) {
        try {
            isExit = c.isExit();
        } catch (DukeException e) {
            ui.showError(e.getMessage());
        } finally {
            ui.showLine();
        }
    }
}
```

- The parser class will solely be in charge of interpreting the string command, read by the Ui class. 
- Based on what it has parsed, the parser class then returns the corresponding Command (an abstract class with abstract method execute(), from which we make the various ListCommand, DeleteCommand etc). Or it could throw an error, if it doesn't understand the command. 
- The command class should have a boolean isExit field, which we can get to see if we should quit Duke after executing the command.
- The Ui class deals with printing to the terminal and reading input from what the user has typed.

john0227 (commented):
Hi Xelyion! I'll also leave my comments here.
Is the field date in the Event class necessary?
I think ArrayList<Task> should be in another class than the Task class.
When you have time, maybe you can also work on packaging to fit OOP design!

PR #342, opened 2019-09-01
[Quek Wei Ping] Duke Increments

qweiping31415:


MauriceTXS:
Code looks good and neat!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public String getTime() {
+        return this.time;
+    }
+
+    public String toString() {
+        return String.format("[D][%s] %s(by: %s)", getStatusIcon(),
+                getDescription(), getTime());
+    }
+}
```
Yehezkiel01: I would suggest adding a new line at the end of each file. Although, it is not in the coding standard, but usually git complains when there is no new line.

src/main/java/Duke.java
```diff
@@ -1,10 +1,175 @@
...
+    final private static String horizontalLine = "    ____________________________________________________________";
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        List<String> start = new ArrayList<>();
```
Yehezkiel01: Duke.java seems to be too complicated since it contains too much logic inside. Therefore, I recommend you to do some refactoring in the code so that the code be more mantainable. Try to refer to Week3's Project on A-MoreOOP increments, it really helps me in improving my code design.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,23 @@
+import java.lang.StringBuilder;
+
+public class ToDo extends Task{
+    
+    public ToDo(String description) {
+        super(description);
+    }
+
+    public static ToDo createToDo(String [] tokens) {
+        StringBuilder builder = new StringBuilder();
```
Yehezkiel01: You can take a look at String.join method to make the code simpler.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#join(java.lang.CharSequence,java.lang.CharSequence...)

Yehezkiel01 (commented):
I would recommend to git ignore the *.class file since those file will most likely be changed everytime you change your java code. Hence, it will be more convenient to just ignore them. Moreover, the bytecode will no longer be used when you have automated your build with gradle.

Overall, you have done a great job on the project. In fact, you have prepared a lot of test cases. Therefore, you should be able to refactor your code easily without causing regeression.

PR #344, opened 2019-09-02
[Natalie Tan] Duke Increments

nattanyz:


src/main/java/Duke.java
```diff
@@ -1,10 +1,10 @@
+import java.util.Scanner;
```
gary-lgy: Unused import :)

src/main/java/DukeBot.java
```diff
@@ -0,0 +1,53 @@
+import java.util.InputMismatchException;
+import java.util.NoSuchElementException;
+import java.util.Scanner;
+
+public class DukeBot {
+    private Scanner in;
+    private TaskList taskList = new TaskList();
+    private static final String welcomeMessage = "What can I do for you?";
```
gary-lgy: From the coding standard:
> Constant names must be all uppercase using underscore to separate words.

src/main/java/DukeBot.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Starts DukeBot and prints welcome message.
+     */
+
+    public void initialise() {
```
gary-lgy: These code should be in the constructor and `run` can be a public method that can be called separately IMO.

src/main/java/DukeBot.java
```diff
@@ -0,0 +1,53 @@
...
+        String logo = " ____        _        \n"
```
gary-lgy: Maybe this should be a constant too.

src/main/java/DukeBot.java
```diff
@@ -0,0 +1,53 @@
...
+        String input; // stores input from user
+
+        this.taskList = new TaskList();
+
+        // continuously take in user input until terminated
+        while (in.hasNextLine()) {
+            input = in.nextLine();
+            if (input.equals("bye")) {
+                // if "bye", terminate
+                // todo: is there a way to combine this into processInput()?
```
gary-lgy: Maybe you can make `processInput` return a boolean to indicate whether the command is "bye"

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+        return type + status + this.description;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); // return tick or X symbols
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+        System.out.println("Nice! I've marked this task as done:");
```
gary-lgy: User interaction should be handled at the call site. Otherwise, it would be difficulty to integrate with GUI.

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
```
gary-lgy: Maybe this class should be static :)

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+    private String description;
+    private Boolean isDone;
+    protected String type;
+
+    public Task(String description) {
```
gary-lgy: IMO this constructor should take in `type` as a parameter as well because `type` is declared in this class.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+    public Task getTask(int taskIndex) throws IndexOutOfBoundsException {
+        return tasks.get(taskIndex - 1);
+    }
+
+    /**
+     * Processes the given input string.
+     *
+     * @param input input string given by user.
+     */
+    public void processInput(String input) throws InputMismatchException {
```
gary-lgy: Instead of having almost all the app logic here, maybe it would help to define a `Command` class and a few subclasses like `AddCommand`, `DoneCommand`, etc. It would make the code more maintainable.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+        // add newTask to taskList
+        tasks.add(newTask);
+        System.out.println("Okay! I've added: " + description
+                + ". Use list to see all your tasks!");
+        // todo: description has trailing space due to using /at or /by as delimiter. how?
+    }
+
+    public void printList() {
+        int i = 1;
+        for (Iterator<Task> iterator = this.tasks.iterator(); iterator.hasNext(); i++) {
```
gary-lgy: I don't think the iterator here is necessary :)

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+            // this should not happen; invalid task type should be caught in caller method already
+            System.out.println("Something went wrong!");
+            break;
+        }
+
+        // add newTask to taskList
+        tasks.add(newTask);
+        System.out.println("Okay! I've added: " + description
+                + ". Use list to see all your tasks!");
+        // todo: description has trailing space due to using /at or /by as delimiter. how?
```
gary-lgy: Use `strip()`, `stripLeading()`, `stripTrailing()`, whichever that suits your need :)

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private ArrayList<Task> tasks = new ArrayList<>();
+
+    public enum TaskType {
+        DEADLINE,
+        EVENT,
+        TODO
+    }
+
+    public void addToList(TaskType taskType, String description, String deadline)
```
gary-lgy: Maybe make this method take in a `Task` and construct the `Task` at the call site. This reduces duplicate code.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+                description = inputReader.nextLine();
+                deadline = "no deadline";
+                this.addToList(TaskList.TaskType.TODO, description, deadline);
+            } catch (NoSuchElementException e) {
+                // user imput after task type is blank
+                System.out.println("Oops! You did not enter a description!");
+            }
+            break;
+        case "event":
+            inputReader.useDelimiter("/at");
```
gary-lgy: This is pretty neat :)

gary-lgy (commented):
Hi Natalie!
Basic functionalities are implemented. Comments are helpful for understanding the code.
Please refer to the individual comments for the specifics.
Don't forget to finish the leftover increments in time ;)

src/main/java/Duke.java
SamYJ2606: Not needed

PR #345, opened 2019-09-02
[Sng Jia Ming Fadi Faris] Duke Increments

fadisng:


src/main/java/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+            c = new ListCommand();
+            break;
+        case "done":
+            c = new DoneCommand(Integer.parseInt(arr[1]) - 1);
+            break;
+        case "delete":
+            c = new DeleteCommand(Integer.parseInt(arr[1]) - 1);
+            break;
+        default:
+            c = new AddCommand(arr);
```
justarock111: It may be more easier to specify the deadline, description, whether it is done and other attributes inside the constructor of AddCommand, since the String input is already split.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage){
+        String description = "";
+        Task t;
+        String next = arr[0];
+        try {
+            if (next.equals("todo") || next.equals("deadline") || next.equals("event")) {
+                if (arr.length == 1) {
+                    throw new EmptyDescriptionException("☹ OOPS!!! The description of a " + next + " cannot be empty.");
```
justarock111: It may be a little easier to have the message already in toString() method of EmptyDescriptionException, so you only have take in the type as a parameter. E.g throw new EmptyDescriptionException("todo"). Since you will be throwing this exception frequently, you won't need to type the message everytime. :)

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
...
+                } else {
+                    for (int i = 1; i < arr.length; i++) {
+                        description += " " + arr[i];
+                    }
+                }
+            } else {
+                throw new UnknownTaskException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
justarock111: It may be a little easier to have the message already in toString() method of UnknownTaskException :)

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+            System.out.println("Got it. I've added this task:");
+            if (next.equals("todo")) {
+                t = new Todos(description.trim());
+            } else if (next.equals("deadline")) {
+                int index = description.indexOf("/");
+                String byWhen = description.substring(index + 4);
```
justarock111: Good idea! Another idea to split your string by "...".split("/by")  instead  so you would not need to care about the index:)

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
...
+                String desc = description.substring(1, index - 1);
+                t = new Deadline(desc, byWhen);
+            } else {
+                int index = description.indexOf("/");
```
justarock111: This is a smart idea! You can also split the string by "...".split("/at") so you would not need to count the index. :)

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
...
+                } else {
+                    for (int i = 1; i < arr.length; i++) {
+                        description += " " + arr[i];
+                    }
+                }
+            } else {
+                throw new UnknownTaskException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+            System.out.println("Got it. I've added this task:");
+            if (next.equals("todo")) {
```
justarock111: Not sure if this is better, but it used switch case so that .equals(...) would not be used that often :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,37 @@
...
+    }
+
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            String fullCommand = ui.readCommand();
+            Command c = Parser.parse(fullCommand);
+            c.execute(list, ui, storage);
+            isExit = c instanceof ExitCommand;
```
justarock111: Good idea!

src/main/java/Command.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * Abstract class to represent the possible commands given to Duke.
+ */
+
+public abstract class Command {
+
+    /**
+     * Executes the different commands respectively.
+     * @param tasks Arraylist of tasks.
```
justarock111, line 9: it may be a good idea to add attributes like isDone or methods like markAsDone() to encapsulate whether as task is done(needed when we load tasks from the file). Since all tasks need this. :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+import java.text.SimpleDateFormat;
+
+public class Deadline extends Task {
+
+    protected String by;
+    protected Date date;
+
+    public Deadline(String description, String by) throws ParseException {
+        super(description);
+        this.by = by;
```
justarock111: Just wondering, why do you need by and date attributes? As I was thinking that since they are similar representations of the same thing, and date can also be used in the toString() method, it is alright to keep date only. :)

src/main/java/Event.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public Event(String description, boolean isDone, String at) throws ParseException {
+        super(description, isDone);
+        this.at = at;
+        this.date = new SimpleDateFormat("dd/MM/yyyy HHmm").parse(at);
+    }
+
+    @Override
+    public String toString() {
+        SimpleDateFormat formatter = new SimpleDateFormat("d MMMM yyyy, hmma");
```
justarock111, line 24: You could also format the date attribute to this format in the constructor, and use it for all your methods, like toString() and more.

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,24 @@
...
+public class FindCommand extends Command {
+    private String description;
+
+    public FindCommand(String description) {
+        this.description = description;
+    }
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        System.out.println("Here are the matching tasks in your list:");
+        ArrayList<Task> matchingTasks = new ArrayList<>();
```
justarock111, line 16: I thought about this idea too! I chose to use streams in the end as the code is much simpler, this is a good idea!

src/main/java/ListCommand.java
```diff
@@ -0,0 +1,16 @@
+/**
+ * Represents a command to print out the tasks in the arraylist.
+ */
+
+public class ListCommand extends Command {
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        System.out.println("Here are the tasks in your list:");
+        int listCount = 1;
```
justarock111, line 9: I would recommend streams to simplify your code.

src/main/java/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class Parser {
+
+    /**
+     * Parses the input string and determines the type of command given.
+     * @param str The input string supplied to Duke.
+     * @return Returns the corresponding command depending on the first word of the input string.
+     */
+    public static Command parse(String str) {
+        String[] arr = str.split(" ");
```
justarock111, line 14: Splitting the string is a good idea! I used contains(...) method instead.

src/main/java/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+     * Is called upon when Duke initialises.
+     * @return Returns an arraylist of tasks written in the file given the file path.
+     */
+    public ArrayList<Task> load()  {
+        ArrayList<Task> list = new ArrayList<>();
+        try {
+            File f = new File(file);
+            Scanner sc = new Scanner(f);
+            Task t;
+            while (sc.hasNext()) {
```
justarock111, line 31: Using Scanner is a good idea!

src/main/java/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            FileWriter fw = new FileWriter(file);
+            String textToAdd = "";
+            for (Task task : list) {
+                textToAdd += task.print() + "\n";
+            }
+            fw.write(textToAdd);
+            fw.close();
+        } catch (IOException e) {
+            System.out.println("File not found.");
```
justarock111, line 68: I used some helper methods like updateText(int taskNum), deleteText(int taskNum) to help me execute 'done' , 'delete' and more commands. I would recommend that.

src/main/java/Ui.java
```diff
@@ -0,0 +1,36 @@
...
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke \nWhat can I do for you?");
+    }
+
+    public String readCommand() {
+        Scanner sc = new Scanner(System.in);
```
justarock111: It may not be necessary to create a new Scanner object everytime readCommand() is called.

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,14 @@
+/**
+ * Test class to test parse() method in Parser.
+ */
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ParserTest {
+    @Test
+    public void parseTest() {
```
justarock111, line 10: I would recommend using this naming standard to name your test methods: featureUnderTest_testScenario_expectedBehavior()

src/test/java/ParserTest.java
```diff
@@ -0,0 +1,14 @@
...
+        assertEquals(new DoneCommand(2), Parser.parse("done 3"));
```
justarock111, line 11: Could I ask why is the command DoneCommand(2) instead of DoneCommand(3) ? As the attribute name in constructor DoneCommand is called taskNum. Thank you!

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,54 @@
+/**
+ * Represents the command to add a task into the arraylist.
+ */
+
+import java.text.ParseException;
+
```
Q-gabe, line 5: I think it's custom to have the class's javadoc comments to be directly above the declaration of the class, after any relevant import statements.

src/main/java/AddCommand.java
Q-gabe: Just to add on, if you use the indexOf("/"), this might not have the intended effect when you have a deadline task initialised with the following command as an example:
`deadline todo/or not to do that is the question /by 12/12/12 12:12`
You can split as @Joanna-YJA suggested, and check if there is more than 2 items in the resulting String array. From there you may proceed to throw the appropriate DukeExceptions.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public String print() {
+        if (this.isDone) {
+            return "D @ 1 @ " + this.description + " @ " + this.by;
+        } else {
+            return "D @ 0 @ " + this.description + " @ " + this.by;
+        }
+    }
+}
```
Q-gabe, line 44: Consider adding newlines at the end of your files, as it is a general convention. It doesn't really cause any issues beside making GitHub and checkstyle whine, but there is a reason behind doing so, https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline. Hope this gives some insight!

src/main/java/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+            System.out.println("Date in wrong format");
+        } finally {
+            return list;
+        }
+    }
+
+    /**
+     * Rewrites the file with the new arraylist of tasks.
+     * @param list arraylist of tasks that are updated after every command given to Duke.
+     */
```
Q-gabe, line 57: Perhaps it may be a more computationally-efficient approach to have writeFile run the moment an exit command runs as the purpose of storage is only for data-permanence through starting and exiting Duke. We need not actually track every manipulation of the task list with every command. 😉

src/main/java/Ui.java
```diff
@@ -0,0 +1,36 @@
...
+        return sc.nextLine();
+    }
+
+    public void printTask(int count, Task t) {
+        if (count == 1) {
+            System.out.println("  " + t + "\nNow you have " + count + " task in the list.");
+        } else {
+            System.out.println("  " + t + "\nNow you have " + count + " tasks in the list.");
+        }
```
Q-gabe, line 18: Nice attention to detail here. 👍

Q-gabe (commented):
I think you did a good job! Aside from some possible bugs that may be missed in terms of parsing the input and other more graceful methods of implementation, it looks pretty solid. LGTM!

src/test/java/TodosTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    public void testPrint() {
+        assertEquals("T @ 1 @ borrow book", new Todos("borrow book", true).print());
+    }
+
+    @Test
+    public void testToString() {
+        assertEquals("[T][\u2713] play basketball", new Todos("play basketball", true).toString());
+    }
+}
```
ShaunNgTX, line 23: Very neat and well written code

PR #350, opened 2019-09-02
[Caesar Pang] Duke Increments

caesarpjz:
Duke Increments


src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
...
+     */
+    public void run() {
+        ui.greeting();
+        ui.nextCommand();
+    }
+
+    /**
+     * Main method.
+     */
+    public static void main(String[] args) throws IOException{
```
LeowWB: I think this is a bad practice. You should probably catch the IOException somewhere within your code, rather than just `throw`ing it from your `main` method.

src/main/java/Event.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Method to give the string that is to be
+     * added to the list of tasks.
+     *
+     * @return Returns the string to be loaded into
+     * the file and printed out.
+     */
+    @Override
+    public String toString() {
+        String date = formatDate(when);
```
LeowWB: You may want to consider storing this formatted date as a variable, so you don't have to keep converting it all the time.

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
+import java.io.*;
+import java.io.BufferedReader;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    protected static ArrayList<Task> taskList = new ArrayList<Task>();
+    protected static String file = "todo.txt";
+
+    public Storage(String file) {
```
LeowWB: You should consider removing this parameter, since it doesn't seem to do anything.

LeowWB: You should consider removing this parameter, since it doesn't seem to do anything.

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+     * @param filepath File that the task is added to.
+     * @param textToAdd Tasks that needs to be added.
+     * @throws IOException If the named file exists but
+     * is a directory rather than a regular file,
+     * does not exist but cannot be created, or
+     * cannot be opened for any other reason.
+     */
+    public void addToFile(String filepath, String textToAdd) throws IOException {
+            FileWriter typer = new FileWriter(filepath, true);
+            typer.write(textToAdd + System.lineSeparator());
```
LeowWB: Have to say that I didn't know about this `System.lineSeparator()` function. Looks like it might be useful for a cross-platform application. Good job 👍👍

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+     * @param filename File that the tasks are in.
+     * @return Returns the number of tasks.
+     * @throws IOException If the named file exists but
+     * is a directory rather than a regular file,
+     * does not exist but cannot be created, or
+     * cannot be opened for any other reason.
+     */
+    public static int countLines(String filename) throws IOException {
+        try (InputStream inputs = new BufferedInputStream(new FileInputStream(filename))) {
+            byte[] characters = new byte[1024];
```
LeowWB: Just curious, is there a reason you chose to use `byte` rather than `char`?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Deletes the specified tasks off the task list.
+     *
+     * @param text Delete command with the which tasks to delete based of their numbering.
+     * @throws IOException If the named file exists but is a directory rather than a regular file,
+     * does not exist but cannot be created, or cannot be opened for any other reason.
+     */
+    public void deleteCommand(String text) throws IOException {
+        int num = text.indexOf(" ");
```
LeowWB: Maybe this variable can be named better?

LeowWB:
Overall, I think you did a great job with the documentation. I didn't notice any `public` functions missing JavaDoc comments. Good work 👍

src/main/java/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        ArrayList<Task> tempList;
+        if (countLines(file) == 0) {
+            throw new DukeException("Woahsies wavy! There is nothing in this file!");
+        } else {
+            while (sc.hasNext()) {
+                String task = sc.nextLine();
+                int index = task.indexOf("[");
+                String taskType = task.substring(index + 1, index + 2);
+                int spaceIndex = task.indexOf(" ");
+                switch (taskType) {
```
jon-chua: There's a specific convention to follow for switch statements! https://nus-cs2103-ay1920s1.github.io/website/coding-standards/java/intermediate.html

src/main/java/TaskList.java
```diff
@@ -0,0 +1,167 @@
+import java.io.IOException;
+import java.sql.SQLOutput;
```
jon-chua: This seems like an unused import.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,167 @@
...
+     * Event command with event and date at which it is and adds
+     * it into the arraylist and file.
+     *
+     * @param text Takes in a event command with the task and date.
+     * @throws DukeException If the user inputs more than the normal
+     * number of days/months/time, this error will be thrown out instead.
+     * @throws IOException If the named file exists but is a directory rather than a regular file,
+     * does not exist but cannot be created, or cannot be opened for any other reason.
+     */
+    public void eventCommand(String text) throws DukeException, IOException {
```
jon-chua: I think the parsing of the String should be moved to the Parser class, the TaskList class should be concerned with the operations on the list itself.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,167 @@
...
+     * Deadline command with deadline task and date at which it is and adds
+     * it into the arraylist and file.
+     *
+     * @param text Takes in a deadline command with the task and date.
+     * @throws DukeException If the user inputs more than the normal
+     * number of days/months/time, this error will be thrown out instead.
+     * @throws IOException If the named file exists but is a directory rather than a regular file,
+     * does not exist but cannot be created, or cannot be opened for any other reason.
+     */
+    public void deadlineCommand(String text) throws DukeException, IOException {
```
jon-chua: Similar to ```eventCommand```, the parsing can be abstracted out to the Parser class.

src/main/java/Ui.java
```diff
@@ -0,0 +1,253 @@
...
+        System.out.println("Ill be adding in more features soon! Please be patient! :)");
+        printLine();
+    }
+
+    /**
+     * Scans the input from the user and does the respective
+     * command based on the input. Prints out error messages as well.
+     */
+    public void nextCommand() {
+        while (scan.hasNext()) {
```
jon-chua: I think we should try to use the Parser class you have created to parse the input first! This way we can abstract out the parsing functions and so the UI class can just be concerned with the UI tasks.

jon-chua (commented):
Your code is very well-documented 👍 
As mentioned, I think it would be a good idea to abstract out parsing-related functionality to the Parser class.
On another note, your printed messages were interesting and useful 😄 
Implementing more tests would prevent regressions from being introduced in the future too!

src/main/java/Task.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Method to format the date into the appropriate format.
+     * For example, 10/02/2012 1800 to 10th of February 2012, 6:00 pm.
+     *
+     * @param date Takes in a valid date to format it.
+     * @return Returns the correctly formatted date with the
+     *     appropriate strings.
+     */
+    public String formatDate(String date) {
```
joshuaseetss: Good method to get the date from a string!

src/main/java/Task.java
```diff
@@ -0,0 +1,132 @@
...
+    public boolean getIsCorrectFormat() {
+        return isCorrectFormat;
+    }
+
+    /**
+     * Returns tick or cross symbol.
+     *
+     * @return A tick or cross to symbolize whether the task has been done.
+     */
+    public String getStatusIcon() {
```
joshuaseetss: Can add [] to the icons here, this would reduce complication and repetition of code

src/main/java/Task.java
```diff
@@ -0,0 +1,132 @@
...
+                + Ui.getNumOfTasks() + " tasks in the list.");
+        Ui.printLine();
+    }
+
+    /**
+     * Method to get the String for whether
+     * the task is done or not.
+     *
+     * @return Returns the tick or cross in brackets.
+     */
```
joshuaseetss: i think its better if there is an @Override over here

src/main/java/TaskList.java
```diff
@@ -0,0 +1,167 @@
...
+                    + "And remember, no negative dates allowed as well!!! :)");
+        }
+    }
+
+
+    /**
+     * Deadline command with deadline task and date at which it is and adds
+     * it into the arraylist and file.
+     *
+     * @param text Takes in a deadline command with the task and date.
```
joshuaseetss: Great documentation here!

src/main/java/Ui.java
```diff
@@ -0,0 +1,276 @@
...
+                printLine();
+            }
+        }
+    }
+
+    /**
+     * Prints indentation.
+     * Helps to order the output, making it much neater.
+     */
+    public static void printIndent() {
```
joshuaseetss: This actually makes code very neat!

joshuaseetss (commented):
Documentation is very thorough and the OOP is generally good! Although some methods can be shorter :)

src/main/java/Command.java
```diff
@@ -0,0 +1,15 @@
+public class Command {
```
groot0x12: I feel that this is a plausible solution but you might want to consider abstract classes as a more efficient way of inheritance for all your other command sub-classes.

src/main/java/Command.java
```diff
@@ -0,0 +1,15 @@
...
+    private String task;
+
+    public Command(action, task) {
+        this.action = action;
+        this.task = task;
+    }
+
+    public String nextCommand(String input) {
+        String text = input;
+        switch ()
```
groot0x12: Missing documentation and code looks incomplete. Missing a return statement as well. I'm guessing what you're trying to do here is abstract the Command class to have some form of inheritance, instead of using methods to facilitate your commands. You're on the right track, just add more to this switch statement.

src/main/java/Parser.java
```diff
@@ -0,0 +1,26 @@
...
+public class Parser {
+
+    Scanner sc = new Scanner(System.in);
+
+    public static Command parse(String line) {
+
+    }
+
+
+    public static String dateFormatter(String text) {
```
groot0x12: This is definitely a plausible solution to obtaining substrings from the command. However, you might want to consider using the java String.split() method to make obtaining your data more efficient!

src/main/java/Duke.java
```diff
@@ -1,10 +1,245 @@
...
+import javafx.scene.control.TextField;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.Region;
+import javafx.scene.layout.VBox;
+import javafx.scene.control.Label;
+import javafx.stage.Stage;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+
+public class Duke extends Application {
```
groot0x12: You might want to consider using multiple classes to facilitate your JavaFX GUI (MainWindow.java / DialogBox.java etc.)
Having all the code in your Duke.java makes it very messy and unreadable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,245 @@
...
+                            ui.printIndent();
+                            throw new DukeException("☹ OOPSY DAISY!!! Please follow the correct format! :<");
+                        }
+                    }
+                } else {
+                    ui.printLine();
+                    ui.printIndent();
+                    switch (text) {
+                    case "todo":
+                        throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty. "
```
groot0x12: Your error format can be stored in a string variable so that it's easy to reuse.

src/main/java/Task.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param date Takes in a valid date to format it.
+     * @return Returns the correctly formatted date with the
+     *     appropriate strings.
+     */
+    public String formatDate(String date) {
+        String formatted = date;
+        if (!date.contains(")")) {
+            try {
+                Date d = new SimpleDateFormat("dd/MM/yyyy hhmm").parse(date);
```
groot0x12: Great effort in handling date and time!

groot0x12 (commented):
Hi! You might want to clean up your repository by avoiding pushing your class files or just do a repository refactor. Overall, great effort but you might want to do more javadocs and documentation because it was a bit hard to understand the intention behind your code :)

PR #352, opened 2019-09-02
[Kerwin Lim] Duke Increments

kerwin97:


yewon0303:
I see that you have added JavaDoc for most of the essential methods which require additional description, that's great! However, the JavaDoc format doesn't seem to be consistent for some of the JavaDoc, for example:
```
/**
     * Constructor for Deadline task
     *
     * @param description description of the deadline.
     * @param by date and time of the deadline in String format
     *
     */
    public Deadline (String description, String by) {
```
There is a full stop at the end of the first param sentence but none at the end of the second param. Also, comparing this with the JavaDoc from another class:
```
/**
	 * Constructor for Duke
	 * Creates new Storage instance and Parser
	 */
```
There is a new line before the end of JavaDoc for Deadline class while for Duke class, JavaDoc comes to an end immediately after the end of the explanation. Maybe you can consider editing the JavaDoc to make it more consistent throughout the different classes? 👍 

The explanation is mostly very clear and explains what the method is expected to do. I like it! The rest looks clean and nice!

src/main/java/Parser.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    /**
+     * Reads the user input and converts it to instructions.
+     *
+     * @throws IOException  if the user input is wrong
+     */
+
+    public void parserRead() throws IOException {
```
Wingedevil, line 46: This function is _very_ long, perhaps it is better if it is better abstracted.

src/main/java/Parser.java
```diff
@@ -0,0 +1,208 @@
...
+        while (sc.hasNext()) {
+            String command = sc.nextLine();
+            if (command.equals("bye")) {
+                UI.bye();
```
Wingedevil, line 50: Is calling static functions directly from the UI class advisable? What if we wanted a ui change?

src/main/java/Parser.java
```diff
@@ -0,0 +1,208 @@
...
+import java.io.IOException;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalTime;
+
+public class Parser {
+
+    static Scanner sc = new Scanner(System.in);
+    private static UI ui;
```
Wingedevil, line 13: Is static necessary here?

src/main/java/UI.java
```diff
@@ -0,0 +1,54 @@
+public class UI {
```
Wingedevil, line 1: Perhaps a better name for this class can be `TextUi`

PR #353, opened 2019-09-02
[Arun Kumarr] Duke Increments

ArunBeCoding:


yaaanch:
It's really good that you've written javadocs comments! But in the style guide, they suggest that the javadoc comments should be capitalised and start with a singular verb. Also, @param doesn't need the "is" word. So for example,

```
constructor for child class
defines the midcommand, and calculates the formattedDate
@param command is the user input string
@throws DukeException in case date is not entered in the correct format
```
could be something like
```
Constructs a Deadline with a description of the deadline and the date of the deadline.
@param command The user input string.
@throws DukeException If date is not entered in the correct format.
```

It's good that you've split up the tasks into more OOP, like with the parser, ui, and storage. I think another way of splitting it up would be to move the interpretation of the commands into the parser, and just have the ui for the printing.

It's also possible to go into even more OOP, for example having different commands. Then your parser could pass command objects back to Duke, then Duke executes the command with the Ui, Storage, and TasksList.

For error handling, some error names could be more descriptive, like DukeException for date formatting.

Good job overall!

yyuanxin:
Hi Arun! I like that you have included Javadoc comments for most of your methods and have separated your test cases into different classes for jUnit testing which makes it more organised! (something I should consider doing as well).

Some suggestions:
- introduce command classes to handle the user input - making it more OOP
- organise your code with packages.

Overall, I think you have done a good job!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * constructor for child class
+     * defines the midcommand, and calculates the formattedDate
+     * @param command is the user input string
+     * @throws DukeException in case date is not entered in the correct format
+     */
+    public Deadline(String command) throws IllegalArgumentException {
+        super(command);
+        this.done = false;
+        String[]splitUpDate = command.split("/",2);
```
WilliamRyank: Need a space between String[] splitUpDate

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        midCommand = splitUpDate[0];
+    }
+
+    /**
+     * to print for "list" command
+     * @return string in the format required
+     */
+    public String printer(){
```
WilliamRyank: Need a space before the curly bracket.

Yen-Peng: add a space between brackets and {

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        String[]splitWords = command.split(" ");
+
+        try {
+            int val = Integer.parseInt(splitWords[1]);
+            ui.deleteMessage(val-1, tasks);
+            //System.out.println("Noted. I've removed this task:"+ "\n" + tasks.taskPrint(val-1) +
```
SimNee: Can delete the comments if it is not needed anymore :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+                String result = "[D][✗] " + midCommand + "(by: " + formattedDate + ")";
+                return result;
+            }
+    }
+
+    /**
+     * to print for text file
+     * @return string in the format required
+     */
+    public String printToOutput(){
```
Yen-Peng: add a space between () and {

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * read user input as convert it into a Task
+     * @param s is the user input string
+     * @return a Task (Deadline) object
+     * @throws DukeException in case segments[3] is not in the proper date format
+     */
+    public static Task outputAsDeadline(String s) throws DukeException {
+        String[]segments = s.split("\\|");
```
Yen-Peng: add a space between [] and segments

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+        return newTask;
+    }
+
+    /**
+     * Creates a new deadline task
+     * @param command is the user string input to be processed
+     * @throws Exception in case user inputs in an incorrect format
+     */
+    public static void createDeadline(String command, TaskList tasks, Storage storage) throws DukeException {
+        String[]splitWords = command.trim().split("\\s",2);
```
Yen-Peng: add space between [] and splitWords

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+     * @throws DukeException in case segments[3] is not in the proper date format
+     */
+    public static Task outputAsDeadline(String s) throws DukeException {
+        String[]segments = s.split("\\|");
+        String taskCommand = segments[2] + "/by: " + segments[3];
+        Deadline newTask = new Deadline(taskCommand);
+
+        if (segments[1].equals(" 1 ")) {
+            newTask.taskDone();
+        } else {}
```
SimNee: Should not have an empty else block :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+    }
+
+    /**
+     * read user input as convert it into a Task
+     * @param s is the user input string
+     * @return a Task (Deadline) object
+     * @throws DukeException in case segments[3] is not in the proper date format
+     */
+    public static Task outputAsDeadline(String s) throws DukeException {
```
WilliamRyank: Maybe can use better argument name instead of `s` as it is not intuitive

src/main/java/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        } catch(Exception e){
+            throw new DukeException("");
+        }
+    }
+
+    /**
+     * marks when task is done
+     */
+    public void taskDone(){
```
Yen-Peng: line 97, 104 & 112: space between () & {
line 97 to 107: You may consider either using try-catch block and throw your exception inside the try block, or just if-else and throw your exception in else block.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,12 @@
+/**
+ * Exception class for incorrect date format
+ */
+public class DukeException extends Exception {
+    /**
+     * exception for incorrect date format
+     * @param message is dummy parameter
+     */
+    public DukeException(String message){
```
WilliamRyank: Space before curly bracket

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        String[]splitWords = command.split(" ",2);
+        String wordToFind = splitWords[1];
+        TaskList findResults = new TaskList();
+
+        for(int i = 0; i < tasks.size(); i++){
+            String taskCommand = tasks.get(i).getCommand();
+            if (taskCommand.contains(wordToFind) ){
+                findResults.add(tasks.get(i));
+            }else{}
```
WilliamRyank: Consider removing empty else.

src/main/java/Event.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param command is the user input string
+     * @throws DukeException in case date is not entered in the correct format
+     */
+    public Event(String command) throws IllegalArgumentException {
+        super(command);
+        this.done = false;
+        String[]splitUpDate = command.split("/",2);
+
+        try {
+            SimpleDateFormat ft = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
SimNee: Could use more informative naming :) eg. dateformat

PR #355, opened 2019-09-02
[Ryan Chew] Duke Increments

iltep64:
(Note: I couldn't get Github to initiate a pull request from an older commit in master, so I have to use a new branch.)


src/main/java/org/duke/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+        this.dispatcher.bindCommand("deadline", this::makeDeadlineTask);
+        this.dispatcher.bindCommand("event", this::makeEventTask);
+        this.dispatcher.bindCommand("todo", this::makeToDoTask);
+        this.dispatcher.bindCommand("delete", this::deleteTask);
+        this.dispatcher.bindCommand("find", this::findTasks);
+        this.dispatcher.setUnknownCommandHandler(cmd -> {
+            throw new DukeException("I'm sorry, but I don't know what that means. :-(");
+        });
+        this.io.setCommandDispatcher(this.dispatcher);
+
```
garylyp, line 44: I like how you implemented all your commands with a central dispatcher. It is very neat and shows clearly what text is bound to each command. Also like how you had a default fallback command handler to handle unknown commands from user input.

src/main/java/org/duke/cmd/Command.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Represents parsed user input, divided into a command type and arguments.
+ */
+public class Command {
+    private static final Pattern switchBoundary = Pattern.compile("\\s+/");
+    private static final Pattern cmdBoundary = Pattern.compile("\\s+");
+    public final String type;
+    public final String arguments;
+    public final Map<String, String> namedArguments;
+    public Command(String type, String arguments, Map<String, String> namedArguments) {
```
garylyp: I felt that the access modifiers for the above fields could be set to private to prevent its values from being modified freely by other classes.

garylyp (commented):
Hi Ryan, I decided to review yours since one of my assignees hasn't submitted his pull request yet. 

Pretty impressive work, with plenty of additional features being implemented to improve the flow of the application and cover all corner cases. The json package was extremely substantial, though I am not too sure if the amount of data stored would actually warrant such intricate designs.

Nevertheless, it was refreshing and eye-opening to see varied applications of new Java features and functional paradigms put in play.

src/main/java/org/duke/cmd/Command.java
iltep64: The `Command` class is intended to be an immutable value type, akin to a C struct. The fields cannot be modified, as they are final (bar the `Map` which has interior mutability). You still have a point in that it breaks encapsulation, so future updates may break consumers of the API.

At the very least, I'll fix it up for the Map case by wrapping it with `Collections.unmodifiableMap()`. Thanks for the catch.

src/main/java/CounterDecorator.java
```diff
@@ -0,0 +1,13 @@
+import java.util.function.Function;
+
+class CounterDecorator implements Function<String, String> {
```
kelvinnharris: I like that you create a class to format the string representation of a `Task`.

src/main/java/Command.java
```diff
@@ -0,0 +1,52 @@
...
+
+		String type = mainArgs[0];
+		//If no arguments, leave as empty string
+		String arguments = mainArgs.length > 1 ? mainArgs[1] : "";
+		Map<String, String> namedArguments = new HashMap<>();
+
+		if(splits.length > 1) {
+			//Each split value is "switch-name args..."
+			//with no preceding slash
+			for(int i=1;i<splits.length;i++) {
```
kelvinnharris: I think it would be more readable to include spaces to become `for (int i = 1; i < splits.length; i++)`

kelvinnharris: I think it would be more readable to include spaces to become `for (int i = 1; i < splits.length; i++)`

src/main/java/Command.java
```diff
@@ -0,0 +1,52 @@
...
+				if(splits[i].isEmpty()) {
+					continue;
+				}
+				//Split at first whitespace
+				String[] switchArgs = cmdBoundary.split(splits[i], 2);
+				String switchName = switchArgs[0];
+				//If no arguments, leave as empty string
+				String switchArguments = switchArgs.length > 1 ? switchArgs[1] : "";
+				namedArguments.put(switchName, switchArguments);
```
kelvinnharris: I find it a bit confusing to use `HashMap` and pass them around as there would be a need to remember the keys. May I know why you decided to proceed with it?

src/main/java/org/duke/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private void addTask(Task t) {
+        this.taskList.add(t);
+        this.io.say(
+                "Got it. I've added this task:",
+                "  " + t,
+                String.format("Now you have %d task%s in the list.",
+                        this.taskList.size(),
+                        this.taskList.size() == 1 ? "" : "s")
```
kelvinnharris: Neat formatting!

src/main/java/org/duke/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+        );
+    }
+
+    private boolean handleBye(Command input) {
+        this.io.say("Bye. Hope to see you again soon!");
+        return true;
+    }
+
+    private boolean displayList(Command input) {
```
kelvinnharris: Is there any other purpose of returning a `boolean` instead of `void` besides passing it as a predicate to the dispatcher? Sorry, I'm a bit lost :')

kelvinnharris (commented):
An eye-opening and inspiring piece with many additional features! I kinda skimmed through to understand what's going on and there's definitely a lot to learn for me.

src/main/java/org/duke/Duke.java
```diff
@@ -133,6 +133,18 @@ private boolean deleteTask(Command input) {
 		return false;
 	}
 
+	private boolean findTasks(Command input) {
+		this.io.say("Here are the matching tasks in your list:");
+		String target = input.arguments.toLowerCase();
+		this.io.say(this.taskList.stream()
```
nishcafe: Nice use of streams here!

src/main/java/org/duke/json/JsonParser.java
```diff
@@ -0,0 +1,248 @@
...
+        while (!reachedEnd) {
+            this.eatWhitespace();
+            char first = this.next();
+            switch (first) {
+            case '"':
+                if (commaState == CommaState.ReadElement) {
+                    throw new JsonException("Missing comma between fields");
+                }
+                String fieldName = this.readString();
+                this.eatWhitespace();
```
nishcafe, line 74: Good use of the SLAP principle here, everything is really easy to read

src/main/java/org/duke/json/JsonWriter.java
```diff
@@ -0,0 +1,270 @@
...
+                    T value = entry.getValue();
+                    ctx.writeField(entry.getKey().toString(),
+                            vctx -> coder.accept(vctx, value));
+                }
+            });
+        }
+    }
+
+    private abstract class BlockContext implements AutoCloseable {
+        private static final String separator = ",";
```
nishcafe, line 202: Good use of access modifiers

PR #356, opened 2019-09-02
[Karan Dev Sapra] Duke Increments

aang114:


src/main/java/DateClass.java
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public static Date stringToDate (String string) {
+
+        SimpleDateFormat format = new SimpleDateFormat("d/MM/yyyy HHmm");
+
+        Date date = null;
+
+        try {
+            date = format.parse(string);
+        } catch (ParseException e) {
```
honhaochen, line 25: Would it be better to handle the exception instead of doing nothing?
Let's say you can throw ParseException to Duke class and let it handle the output/ message.

src/main/java/Duke.java
```diff
@@ -1,10 +1,69 @@
...
+     * @param filePath File Path of duke.txt
+     * @param folderPath Folder Path in which duke.txt is saved
+     */
+    public Duke(String filePath, String folderPath) {
+        ui = new Ui();
+        storage = new Storage(filePath, folderPath);
+        try {
+            tasks = new TaskList(storage.retrieve());
+        } catch (FileNotFoundException e) {
+
```
honhaochen, line 55: Would it be better to inform the user about the exception?
For me particularly I would inform that "file isn't found, a new file will be created instead"

src/main/java/TaskList.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param index The index of the task
+     * @return The deleted task
+     */
+    public Task delete (int index) {
+
+        Task task = list.remove(index);
+        return task;
+    }
+
+    public Task[] find(String text) {
```
honhaochen, line 64: Would it be easier to return ArrayList<Task>?

src/main/java/ListCommand.java
```diff
@@ -0,0 +1,11 @@
+import java.io.IOException;
+
+public class ListCommand extends Command {
+
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws IOException {
+        ui.list();
+        taskList.list();
```
tcgj: Perhaps you might want to make the two methods here somewhat clearer in terms of naming? Both `ui.list()` and `taskList.list()` seem to mean the same thing even though they most likely do very different things.

src/main/java/Parser.java
```diff
@@ -1,33 +1,123 @@
...
+            return new DeleteCommand(index);
         }
 
         switch (command) {
         case "list":
-            return Bot.TaskType.LIST;
+            return new ListCommand();
         case "bye":
-            return Bot.TaskType.BYE;
+            return new ExitCommand();
```
tcgj, line 48: Would it be better if you kept these two commands together with the rest of the switch statement so that it remains as a block? If you have no spaces, split actually returns you an array containing just the command itself, so `word[0]` would still correspond to "list" and "bye" in those cases.

src/main/java/Parser.java
```diff
@@ -1,33 +1,123 @@
...
+
+        if (words[1].isBlank()) throw DukeException.emptyDescription();
+
+        int index = Integer.valueOf(words[1]) - 1;
+
+        return index;
+
+    }
+
+    public static Task parsesAdd (String command, TaskType taskType) throws DukeException {
```
tcgj, line 120: Perhaps it may be better to rename your function to an imperative, maybe `parseAdd`? This is so that the function adheres to coding standards, and states "what it will do" rather than "what it does".

src/main/java/Parser.java
```diff
@@ -1,33 +1,123 @@
...
+
+        String[] words = command.split(" ", 2);
+
+        if (words.length == 1) throw DukeException.emptyDescription();
+
+        Task task = null;
+
+        String description;
+
+        switch (taskType) {
```
tcgj, line 132: Hmm, I wonder if it's very efficient if you have another switch statement here considering you already individually parsed the type of task to add inside of your `parse` function. Not sure if it would be better to have three different functions here instead. Might be debatable!

src/main/java/Storage.java
```diff
@@ -2,27 +2,39 @@
...
-    public static void writeToFile(String filePath, String textToAdd) throws IOException {
+    public void writeToFile(String filePath, String textToAdd) throws IOException {
         FileWriter fw = new FileWriter(filePath);
         fw.write(textToAdd);
         fw.close();
     }
 
-    public void retrieve () throws FileNotFoundException {
-        File f = new File(Storage.DUKE_TXT_PATH); // create a File for the given file path
+    public ArrayList<Task> retrieve () throws FileNotFoundException {
```
tcgj, line 54: Would it be better if you checked if the file existed, then returned an empty list, rather than throw a `FileNotFoundException` if it didn't already exist?

tcgj (commented):
Looks good!

honhaochen (commented):
Command Class:
- Under execute method, i think it's very expensive to store the entire list again and again for every command. Wouldn't it already served the purpose if you just save it in during your ExitCommand?

Overall:
- Some files are missing JavaDocs (FindCommand)
- Great coding conventions
- Great implementations! However, some recommendations can be found in the code specifically.

src/main/java/Parser.java
honhaochen: Yes, I do agree with Terence

PR #357, opened 2019-09-03
[Elizabeth Koh] Duke Increments

elsakoh:


src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
         System.out.println("Hello from\n" + logo);
     }
+    static void greeting(){
+        String output = "    ____________________________________________________________\n" +
+        "     Hello! I'm Duke\n" + 
+        "     What can I do for you?\n" +
+        "    ____________________________________________________________\n";
+        System.out.println(output);
+    }
+    static String addDoubleLine(String str){
```
ArunBeCoding: Seems like quite a tedious process, especially when you increase the OOP later on, you can try to just print out a line before and after parsing the string

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+        String line = "    ____________________________________________________________";
+        return line + "\n" + str + "\n" + line;
+    }
+    static void handleList(ArrayList<Task> ls){
```
ArunBeCoding: Take note of java coding standards for parameter naming

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+        }
+        System.out.println(addDoubleLine(output));
+    }
+    static void printMsg(Task t, int size){
+        System.out.println(addDoubleLine("     Got it. I've added this task: \n" + "      " + t.toString() + "\n     Now you have " + size + " tasks in the list."));
+    }
+    static void handleInput(){
+        ArrayList <Task> list = new ArrayList<Task>();
+        Scanner sc = new Scanner(System.in);
+        while (sc.hasNextLine()){
```
ArunBeCoding: Take note of Java coding standard for spacings between brackets in loops

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                String [] splitDate = line.replace("deadline", "").split("/by");
+                if(splitDate.length < 2){
+                    System.out.println(addDoubleLine("    ☹ OOPS!!! Deadlines require both a description and a date by \n    (e.g. deadline homework3 /by tomorrow)"));
+                }else{
+                    list.add(new Deadline(splitDate[0].trim(), splitDate[1].trim()));
+                    Task current = list.get(list.size()-1);
+                    printMsg(current, list.size());
+                }
+            //error handling
+            }else{
```
ArunBeCoding: you might want to use try catch for this part, because to my understanding, if the user inputs "done hello" it will still throw error and not catch it

ArunBeCoding (commented):
just small changes, for now, most of the changes will come in at later stages though

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+    void logo(){
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
     }
+    static void greeting(){
+        String output = "    ____________________________________________________________\n" +
```
fangpinsern: What i think you can do is to put those formatting lines as a function by itself instead. So everything you have a new message, you can just call the function instead of copy and pasting. 

Also in the event you want to change the format, it will make things a little easier

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+        Storage storage = new Storage();
+        ArrayList <Task> list = storage.recoverTasks();
+        Scanner sc = new Scanner(System.in);
+        while (sc.hasNextLine()){
+            try{
+                String line = sc.nextLine().toLowerCase();
+                String[] splited = line.split(" ");
+                String check = splited[0];
+                //bye
+                if (check.equals("bye")) {
```
fangpinsern: possible to change this to a switch statement. I did it this way too. But would have made it a switch statement on hind sight

src/main/java/Storage.java
```diff
@@ -0,0 +1,81 @@
+import java.io.FileNotFoundException;
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.text.ParseException;
+
+public class Storage {
+    private String fileLocation =  "data/duke.txt";
```
fangpinsern: Try not to hard code this to be the text file. maybe it could be the main's choice instead so different instances can have different file paths

fangpinsern (commented):
Overall pretty well done, just some minor things to take note. Also, you can ignore the .class files when pushing. will look more elegant i guess :) Great job!

src/main/java/DateTime.java
```diff
@@ -0,0 +1,22 @@
...
+        this.date = formatter.parse(input);
+    }
+
+    @Override
+    public String toString() {
+        String dateFormat = "dd-MM-yyyy HH:mm";
+        SimpleDateFormat formatter = new SimpleDateFormat(dateFormat);
+        return formatter.format(this.date);
+    }
+}
```
stevenwjy: It might look better in GitHub if you can add newline at the end of every file, because otherwise there will be a red symbol that appears indicating that there is no newline.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,15 @@
...
+        super(taskName);
+        this.by = by;
+    }
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
+    }
+    @Override
+    public String getStorageString(){
+        return "D | " + super.isDone + " | " + taskName + " | " + by.toString();
```
stevenwjy: You can just use `isDone` directly instead of `super.isDone`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+        "     Hello! I'm Duke\n" + 
+        "     What can I do for you?\n" +
+        "    ____________________________________________________________\n";
+        System.out.println(output);
+    }
+    static String addDoubleLine(String str){
+        String line = "    ____________________________________________________________";
+        return line + "\n" + str + "\n" + line;
+    }
+    static void handleList(ArrayList<Task> ls){
```
stevenwjy: It would be better if you can give a more descriptive name for function parameter(s). You may refer to Java coding standards about the proper way of doing it.

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+                //bye
+                if (check.equals("bye")) {
+                    System.out.println(addDoubleLine("     Bye. Hope to see you again soon!"));
+                    storage.saveFile(list);
+                    System.exit(0);
+                //list
+                }else if(check.equals("list")){
+                    handleList(list);
+                //done
+                }else if(check.equals("done")){
```
stevenwjy: To improve readability, it would be better if you put spaces between brackets. You may refer to Java coding standards about it.

stevenwjy (commented):
Overall the code looks fine and the logic seems to work correctly, nice!! 

However, here are some feedback that you may want to consider in order to improve the code quality:
- Run a checkstyle scan and fix the errors indicated. You may also want to use InteliJ auto-formatter to assist you in fixing the issues.
- Separate the code into packages (e.g. ui, commands, and tasks).
- Avoid mixing all the logic in one place (i.e. `handleInput` function), it might be better to create a separate class for each command that has its own execution function.
- Don't put too many responsibilities for your `Duke` class (i.e. handle input/output, logic, saving to file, etc.), you may want to break those functionalities into multiple classes.
- It would be better if you ignore the `.class` files when you push your code to GitHub.

PR #363, opened 2019-09-03
[Alvin Ng] Duke Increments

Alvinnyk:


YiJiee:
Packages are well organised into 3 components - Storage, Ui and Model.
Naming conventions for class file is a little bit inconsistent (some classes start with smaller case and others start with capital case).
Code is written clearly and easy to understand.
Specific exceptions were not implemented and exceptions were caught at the highest level (Exception).
Other than that, the app compiles and works great until Level-9.
Reviewed on 4 Sep 2019

R-D-D-D:
The TaskList extends ArrayList which saves so much trouble I should learn from you.
I think generally the run is too bulky, you might consider abstracting out the command and maybe a parser to parse user inputs. 
Instead of representing a deadline task as a whole string, maybe can have a Date object to represent the date.
The rest looks good!

src/main/java/Logic/EventCommand.java
```diff
@@ -0,0 +1,40 @@
...
+    public EventCommand(String arguments){
+        this.arguments = arguments;
+    }
+
+    @Override
+    public void execute(Tasklist tasks, UI ui, Storage storage) {
+        if(arguments == null){
+            ui.printData("OOPS!!! The description of a event cannot be empty.");
+        } else {
+            String[] sp = arguments.split(" /at ", 2);
```
sreesubbash: Maybe tokens or stringArray is a better variable name than "sp".

src/main/java/Logic/DoneCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public void execute(Tasklist tasks, UI ui, Storage storage) {
+        int index = Integer.parseInt(arguments);
+
+        tasks.get(index - 1).markAsDone();
+
+        String content = "Nice! I've marked this task as done:\n" +
+                "[" + tasks.get(index - 1).getIsDoneSymbol() + "] " + tasks.get(index - 1).getDescription() +"\n";
```
sreesubbash: "+" comes on this line, not 20. Maybe running checkstyle would help?

src/main/java/Logic/CommandParser.java
```diff
@@ -0,0 +1,55 @@
...
+        } else if (commandName.equals(EVENT)){
+            return new EventCommand(arguments);
+        } else if (commandName.equals(DEADLINE)){
+            return new DeadlineCommand(arguments);
+        } else if (commandName.equals(FIND)){
+            return new FindCommand(arguments);
+        } else if (commandName.equals(HELP)) {
+            return new HelpCommand();
+        } else {
+            return new WrongCommand();
```
sreesubbash: Since this isn't expected behavior, could throw an exception?

src/main/java/Logic/DeadlineCommand.java
```diff
@@ -0,0 +1,38 @@
...
+        if(arguments == null) {
+            ui.printData("OOPS! The description of a deadline cannot be empty.\n");
+        } else {
+            String[] sp = arguments.split(" /by ", 2);
+
+            tasks.add(new deadline(sp[0], sp[1]));
+
+            String content = "";
+
+            content = content.concat("Got it. I've added this task:\n");
```
sreesubbash: You could use "hello " + "world" to join strings.

src/main/java/Logic/DeadlineCommand.java
```diff
@@ -0,0 +1,38 @@
...
+            content = content.concat("[D][✗] " + sp[0] + " (by: " + sp[1] + ")\n");
+            content = content.concat("Now you have " + tasks.size() + " tasks in this list\n");
+
+            ui.printData(content);
```
sreesubbash: Maybe you could abstract the formatting of the output string to toString() or Ui

src/main/java/Logic/ListCommand.java
```diff
@@ -0,0 +1,41 @@
...
+
+    }
+
+    @Override
+    public void execute(Tasklist tasks, UI ui, Storage storage){
+        int i;
+
+        String content = "";
+
+        for(i = 0; i < tasks.size(); i++){
```
sreesubbash: You can make the scope of `i` more narrow: to be within the for loop

src/main/java/Model/deadline.java
```diff
@@ -0,0 +1,28 @@
+package Model;
+
+public class deadline extends Task{
```
sreesubbash: `Deadline`?

src/main/java/Storage/Storage.java
```diff
@@ -0,0 +1,176 @@
...
+    public Tasklist parseFromFile(Path filepath){
+        Tasklist tasks = new Tasklist();
+        try{
+            Reader reader = Files.newBufferedReader(filepath);
+            Scanner sc = new Scanner(reader);
+            while(sc.hasNextLine()){
+                tasks.add(parseLine(sc.nextLine()));
+            }
+
+        } catch (Exception E){
```
sreesubbash: Perhaps you could catch the specific exception.

sreesubbash (commented):
Reviewed by 2103T-T11-2

iskandarzulkarnaien:
Code looks good! Some parts have heavily nested if-else blocks which might be worth refactoring

PR #364, opened 2019-09-04
[Yoon Jia Jun Ken] Duke Increments

Xelyion:


KendrickAng:
Hello Xelyion! I understand that the pull request should be done with your project's master branch, but as the bulk of your business logic is in your previous pull request (branch-Level-9), I have made my comments there instead.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,62 @@
...
+	public AddCommand(String type, String command) {
+        super(type, command);
+    }
+
+    /**
+     * Executes the AddCommand.
+     * Adds Deadlines: "deadline".
+     * Adds Events: "event".
+     * Adds ToDos: "todo".
+     *
```
ZhangHuafan: Very detailed Javadoc but may consider writing more clear explanations.

src/main/java/Command.java
```diff
@@ -0,0 +1,21 @@
...
+    protected static boolean isExit = false;
+
+    public Command(String type, String fullCommand) {
+        this.type = type;
+        this.command = fullCommand;
+    }
+
+    public String execute(Ui ui, TaskList taskList, Storage storage) throws DukeException {
+        return "";
+    }
```
ZhangHuafan: May consider making this method as abstract so that you do not need to return a default string.

src/main/java/Command.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * An abstract class command, created from the Parser.
+ * Contains Strings to store the command and type of command.
+ */
+
```
ZhangHuafan: I think there is no line break between them =)

src/main/java/Command.java
```diff
@@ -0,0 +1,21 @@
...
+        this.type = type;
+        this.command = fullCommand;
+    }
+
+    public String execute(Ui ui, TaskList taskList, Storage storage) throws DukeException {
+        return "";
+    }
+
+
+}
```
ZhangHuafan: Personally, I think the child classes of Command can self-explain their "type". Maybe you could consider making full use of this.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+                long min = Long.parseLong(splitString[1].substring(2, 4));
+                date.setTime(date.getTime() + ((hour * 60 + min) * 60000));
+                this.time = date.toString();
+            } else {
+                throw new DukeException("    Time is not in the correct format");
+            }
+        } else {
+            throw new DukeException("    Date is not in the format: (DD/MM/YYYY HHMM)");
+        }
+    }
```
ZhangHuafan: For parsing date string to a date object. You may find LocalDateTime.parse() method useful. You may want to refer to this link: https://www.baeldung.com/java-string-to-date. Besides, would it be better if this method belongs to Parse class?

1.

src/main/java/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+            command = new ExitCommand(type, fullCommand);
+        } else if (type.equals("save")) {
+            command = new StorageCommand(type, fullCommand);
+        } else {
+            command = new TaskListCommand(type, fullCommand);
+        }
+        return command;
+    }
+
+}
```
ZhangHuafan: It is very nice that you added "save" command! However, you may consider using switch and case to make the code structure clearer. Besides, you could consider using the String.split().

ZhangHuafan (commented):
Hi, Jia Jun! I think your JavaDoc is very detailed and most of them are following the correct formats. The logic of indvidual parts is mostly clear as well. However, you may want to refactor some parts so that the whole structure could be more reasonable.

PR #365, opened 2019-09-04
[Oliver Cheok] Duke Increments

olivercheok20:


src/main/java/Duke.java
```diff
@@ -1,10 +1,160 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        // Create a scanner to take in user input
+        Scanner sc = new Scanner(System.in);
+        Parser parser = new Parser();
```
Abhiman2211, line 300: Your process is very systematic, but a little long and defeats the purpose of OOP design. I would suggest refactoring and breaking down every big chunk of code performing a single function into different functions and then simply calling this function in the if-else statements. This would improve the readability of your code.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,66 @@
+package duke.parser;
```
Abhiman2211, line 1: Good that you included a parser class to improve abstraction. However, big functions from the Duke class can also be put into the parser to parse the user commands. This would increase the use of this class and the readability of your code.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,18 @@
+package duke.ui;
+
+public class Ui {
+    /**
+     * Method to format messages between lines and indented.
+     */
+
+    public static String formatMessage(String message) {
```
Abhiman2211, line 8: Nice! This is clever abstraction that saves space and improves code readability.

PR #367, opened 2019-09-04
[Yew Kai Zhe] Duke Increments

ZiHawkEye:


khairulazman1997:
Great work overall! I couldn't find much but one room for improvements is to possibly change your Task class into an interface or abstract class since it would never be used on its own to prevent it from ever being initialised on its own. All round great work and all the best for the rest of the module!

PR #368, opened 2019-09-04
[Jason] Duke Increments

ReignOfComputer:


src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+            } finally {
+                ui.showLine();
+            }
+        }
+    }
+
+    public static void main(String[] args) throws DukeException, IOException {
+        new Duke("list.txt").run();
+    }
+
```
ShirleyWangxt: Maybe it is advisable to remove the commented code if it is not used to avoid confusion.

src/main/java/Ui.java
```diff
@@ -0,0 +1,9 @@
+public class Ui {
```
ShirleyWangxt: It would be better if there can be more functionality to represent interaction with users, including the different response messages.

ShirleyWangxt (commented):
Good job on implementing the features. But I think it would be better if the project can be more OOP. I am not really sure about certain classes implemented as some do not have a lot of content inside. Maybe there can be more classes to represent different commands for example.

src/main/java/Duke.java
ReignOfComputer: You caught me mid-refactor, laptop ran out of battery so I had to make a Checkpoint commit.

PR #384, opened 2019-09-07
Add assertions to code

lye-jw:


lye-jw:
Attempting to merge to wrong master

PR #443, opened 2019-09-10
Branch a assertions

jeunhoe:


damithc:
This should be in your own fork.

PR #463, opened 2019-09-11
A assertions

waynefong0401:


damithc:
This PR should be in your own fork

src/main/java/duke/memo/message/MessageGenerator.java
```diff
@@ -0,0 +1,78 @@
+package duke.memo.message;
+
+import duke.memo.data.TaskList;
+import duke.memo.task.Task;
+
+import java.util.ArrayList;
+
+public class MessageGenerator {
+
+    public MessageGenerator() {
```
gabrielseow, line 10: Could consider having the methods as static, since the methods do not depend on any instance variables

src/main/java/duke/memo/data/TaskList.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Convert raw log string to a task object.
+     *
+     * @param log  Raw log.
+     * @return Task return the parsed log
+     * @throws DukeException Throw if there is a problem in the log.
+     */
+    public Task readLog(String log) throws DukeException {
```
gabrielseow, line 39: Not much issues with the code, but perhaps the readLog method would be more appropriate to be placed in the Parser class, since you are parsing strings

PR #482, opened 2019-09-13
[Michelle] Duke Increments

michelleykw:


src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+        Scanner sc = new Scanner(System.in);
+        Parser parser = new Parser(sc);
+
+        while (sc.hasNext()) {
+            String text = parser.nextCommand();
+            String[] commandArr = parser.breakDownCommand(text);
+            String command = parser.getCommand(commandArr);
+            if (command.equals("todo")) {
+                try {
+                    if (text.length() <= 4) {
```
tysg: Good guard clause!

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+    public void run() throws IOException, ParseException {
+        ui.printHello();
+        Scanner sc = new Scanner(System.in);
+        Parser parser = new Parser(sc);
+
+        while (sc.hasNext()) {
+            String text = parser.nextCommand();
+            String[] commandArr = parser.breakDownCommand(text);
+            String command = parser.getCommand(commandArr);
+            if (command.equals("todo")) {
```
tysg: Maybe you can try to abstract out the handler methods, and make use of `switch`?

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+                    taskList.addTask(todo);
+                    ui.printTaskAdded(todo, taskList.getSize());
+                    storage.appendTaskToFile(todo);
+                } catch (DukeException e) {
+                    ui.printTodoError();
+                }
+            } else if (command.equals("deadline")) {
+                try {
+                    if (text.length() <= 8) {
+                        throw new DukeException();
```
tysg: Shall we consider specifying the error occurred here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+                    Todo todo = parser.getTodo(text);
+                    taskList.addTask(todo);
+                    ui.printTaskAdded(todo, taskList.getSize());
+                    storage.appendTaskToFile(todo);
+                } catch (DukeException e) {
+                    ui.printTodoError();
+                }
+            } else if (command.equals("deadline")) {
+                try {
+                    if (text.length() <= 8) {
```
tysg: Maybe avoid using magic number here. Instead, you could use `command.split(" ")[0].equals("deadline")`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+                    }
+                    String[] descArr = parser.breakDownDescription(text);
+                    Date date = parser.getDate(descArr, storage);
+                    Deadline deadline =
+                            new Deadline(parser.getDeadlineDesc(descArr), date);
+                    taskList.addTask(deadline);
+                    ui.printTaskAdded(deadline, taskList.getSize());
+                    storage.appendTaskToFile(deadline);
+                } catch (DukeException e) {
+                    ui.printDeadlineError();
```
tysg: Try adding in the error message inside the `DukeException` class, and just call `e.printMessage`. Same for the rest of the code.

src/main/java/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Gets the todo based on the command line.
+     *
+     * @param line The command list.
+     * @return The todo.
+     */
+    public Todo getTodo(String line) {
+        String task = line.substring(5);
```
tysg: Since you have already split the command by `breakDownCommand`, maybe try making use of that abstraction?

src/main/java/Storage.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Converts the task to string format to be added to the file.
+     *
+     * @param task The task to be converted in string format.
+     * @return The task in a string format.
+     */
+    private String convertTaskToFileFormat(Task task) {
+        StringBuffer textToAdd = new StringBuffer();
```
tysg, line 176: Good use of `StringBuffer` to avoid extra overhead.

src/main/java/Storage.java
```diff
@@ -0,0 +1,164 @@
...
+        fw.close();
+    }
+
+    /**
+     * Updates the current list of tasks to the file.
+     *
+     * @param list The list of tasks to be updated to the file.
+     * @throws IOException If an input or output exception occurred.
+     */
+    public void updateTaskInFile(ArrayList<Task> list) throws IOException {
```
tysg, line 228: Try using `List` instead of `ArrayList`, as programming to interface is a good practice.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,102 @@
+import exception.DukeException;
+import tasks.Task;
+
+import java.util.ArrayList;
+
+/**
+ * Contains the task list and updates tasks in the list.
+ */
+public class TaskList {
+    ArrayList<Task> list = new ArrayList<Task>();
```
tysg, line 9: This should be done in the class constructor.

PR #536, opened 2019-09-19
[Rushagrid] Duke Increments

Krit-K:


src/main/java/Storage.java
```diff
@@ -26,56 +26,64 @@
...
-                    case "D":
-                        String deadlineDescription = task[2].trim();
-                        String by = task[3].trim();
-                        SimpleDateFormat sdf2 = new SimpleDateFormat("dd/MM/yyyy HHmm");
-                        Deadline newDeadline = new Deadline(deadlineDescription, sdf2.parse(by));
-                        if (isTaskDone) newDeadline.markAsDone();
-                        taskList.add(newDeadline);
-                        break;
-                    default:
+                updateData(taskList, task);
```
elsakoh, line 35: it was a great idea to abstract out this long chunk :)

